PR #60, opened 2022-02-19
Enable MyPy in strict mode

kaxil:
Enable MyPy in a strict mode and run via docker instead of pre-commit so the stubs are installed.


kaxil:
> I can see a couple of inconsistencies on which we need to conclude
> 

I agree. For now, let's park it and we can discuss this next week

kaxil:
🎉 Ok, I have finally got this in !! We also have it running in the CI

Issue #108, opened 2022-03-10
S3KeysUnchangedTrigger run method on infinite loop


sunank200:
**Describe the bug**
Currently, run method calls `is_keys_unchanged()` and the  self.previous_objects , self.last_activity_time and self.inactivity_seconds is being reset every time in the while loop but it should have been the value returned from `is_keys_unchanged()`. In the [code](https://github.com/astronomer/astronomer-providers/blob/main/astronomer/providers/amazon/aws/hooks/s3.py#L255) for `is_key_unchanged`, it should return 
`return {"status": "pending", "previous_objects":previous_objects, "last_activity_time":last_activity_time,"inactivity_seconds":inactivity_seconds }
`
So that it can be set on trigger

PR #109, opened 2022-03-10
Fix bug in `S3KeysUnchangedTrigger` and `is_key_unchanged` in hook

sunank200:
- Fix the bug on `S3KeysUnchangedTrigger` and `is_key_unchanged`

Closes : #108

### DIFF
--- a/astronomer/providers/amazon/aws/hooks/s3.py
+++ b/astronomer/providers/amazon/aws/hooks/s3.py
@@ -252,7 +252,12 @@ class S3HookAsync(AwsBaseHookAsync):
             last_activity_time = datetime.now()
             inactivity_seconds = 0
             previous_objects = current_objects
-            return {"status": "pending"}
+            return {
+                "status": "pending",
+                "previous_objects": previous_objects,
+                "last_activity_time": last_activity_time,
+                "inactivity_seconds": inactivity_seconds,
+            }
 
         if len(previous_objects) - len(current_objects):
             # During the last poke interval objects were deleted.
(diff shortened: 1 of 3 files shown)


astronomer/providers/amazon/aws/hooks/s3.py
```diff
@@ -215,7 +215,7 @@ async def is_keys_unchanged(
         inactivity_seconds: int = 0,
         allow_delete: bool = True,
         last_activity_time: Optional[datetime] = None,
-    ) -> Dict[str, str]:
+    ) -> Dict[str, Any]:
```
phanikumv, line 218: why changed str to Any?

astronomer/providers/amazon/aws/triggers/s3.py
```diff
@@ -184,9 +184,13 @@ async def run(self) -> AsyncIterator["TriggerEvent"]:  # type: ignore[override]
                         self.allow_delete,
                         self.last_activity_time,
                     )
-                    if result.get("status") == "success" or result.get("error") == "error":
+                    if result.get("status") == "success" or result.get("status") == "error":
                         yield TriggerEvent(result)
                         return
+                    elif result.get("status") == "pending":
```
phanikumv, line 190: unit test should cover the "pending" scenario

astronomer/providers/amazon/aws/hooks/s3.py
sunank200: it returns  str objects and DateTime object too now in the dict

phanikumv: Try Dict[str, Union[str, datetime, int]]

sunank200: Values for the dict is being parsed on Trigger and the value for the key can be anything rather than just one type.

phanikumv (approved):

PR #121, opened 2022-03-14
Fix file paths in `CHANGELOG.rst`

kaxil:
We had old import paths : `astronomer_operators` -> `from astronomer.providers` for some of the operators


phanikumv (approved):
Looks good to me.

PR #123, opened 2022-03-14
Dev: Add make command to rebuild Docker image

kaxil:
This will allow rebuilding the docker-image whenever needed, example after we added new deps


phanikumv (approved):
Looks great!

PR #102, opened 2022-03-08
Parameterize example DAGs for integration tests

pankajastro:
issue: #63 
Parameterize example dags to run automation test

PR #137, opened 2022-03-17
Bug fix for gcs operator example

pankajastro:
Some of the GCS operators/sensors accept only `google_cloud_conn_id` as connection param. 
because this https://github.com/astronomer/astronomer-providers/pull/102 merge example was failing.


sunank200 (approved):
LGTM

kaxil (approved):

PR #142, opened 2022-03-22
Add `bandit` pre-commit for Security scanning

kaxil:
https://github.com/PyCQA/bandit - Bandit is a tool designed to find common security issues in Python code.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -89,6 +89,12 @@ repos:
           .*example_dags/.*|
         additional_dependencies: ['toml']
 
+  - repo: https://github.com/PyCQA/bandit
+    rev: 1.7.4
+    hooks:
+      - id: bandit
+        exclude: ^tests/.*\.py$|.*example_dags/.*
+
   - repo: local
     hooks:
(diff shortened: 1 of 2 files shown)


pankajastro:
Shall we also add a badge for bandit check in the README.rst ?

.pre-commit-config.yaml
```diff
@@ -89,6 +89,12 @@ repos:
           .*example_dags/.*|
         additional_dependencies: ['toml']
 
+  - repo: https://github.com/PyCQA/bandit
+    rev: 1.7.4
+    hooks:
+      - id: bandit
+        exclude: ^tests/.*\.py$|.*example_dags/.*
```
sunank200, line 96: Why did we exclude the `example_dags`?

sunank200 (approved):

rajaths010494 (approved):

.pre-commit-config.yaml
kaxil: Temporarily added, as some of the example DAGs don't satisfy bandit but that is OK for now. 

I will create a separate GitHub issue to fix it

Issue #143, opened 2022-03-23
Databricks triggerer fails to check databricks job state


wolfier:
**Describe the bug**
Databricks triggerer fails to check databricks job state.

**To Reproduce**
1. Open a python interpreter.
2. Run the following snippet provided a valid databricks connections and run id.
```
async def test():
    from asgiref.sync import sync_to_async
    h = DatabricksHookAsync(databricks_conn_id='databricks')
    endpoint = ('GET', 'api/2.1/jobs/runs/get')
    json = {"run_id": 142730}
    response = await h._do_api_call_async(endpoint, json)
    print(response)

asyncio.run(test())
```

**Expected behavior**
I expect the a 200 response when provided a valid credential with run id.

wolfier:
The issue is because the authentication header is malformed.

There should not be a colon in the header.

https://github.com/astronomer/astronomer-providers/blob/1.0.0/astronomer/providers/databricks/hooks/databricks.py#L68

wolfier:
I was able to fix the issue inline.

```
async def test():
    from airflow.providers.databricks.hooks.databricks import USER_AGENT_HEADER, DatabricksHookAsync
    h = DatabricksHookAsync(databricks_conn_id='databricks')
    headers = USER_AGENT_HEADER
    c = await sync_to_async(h.get_connection)(h.databricks_conn_id)
    auth_str = f"{c.login}:{c.password}"
    encoded_bytes = auth_str.encode("utf-8")
    auth = base64.b64encode(encoded_bytes).decode("utf-8")
    headers["Authorization"] = f"Basic {auth}"
    url = "***/api/2.1/jobs/runs/get"
    method, endpoint = ('GET', 'api/2.1/jobs/runs/get')
    json = {"run_id": ***}
    print(auth_str)
    print(auth)
    print(headers)
    async with aiohttp.ClientSession() as session:
        response = await session.get(
            url,
            json=json if method in ('POST', 'PATCH') else None,
            params=json if method == 'GET' else None,
            # auth=auth,
            headers=headers,
            timeout=10,
        )
        print(response)


asyncio.run(test())
```

The output is a 200.

```
<ClientResponse(https://***/api/2.1/jobs/runs/get?run_id=***) [200 OK]>
[...]

PR #147, opened 2022-03-23
Fix Databricks triggerer fails to check databricks job state

sunank200:
- Remove the colon in the auth header in the hooks. The issue was that the authentication header is malformed.

`Basic: <credential>`

There should not be a colon in the header.
`Basic <credential>`

- Propagate the error message from triggerer to work in case life_cycle_state: `TERMINATED`, `SKIPPED` or `INTERNAL_ERROR`.
- Update the tests

Closes: #143


phanikumv (approved):
Good job Ankit for turning this up so quickly !!

PR #151, opened 2022-03-23
Bug fix on  example_bigquery_sensors.py dag

rajaths010494:
This PR fixes a bug on example_bigquery_sensors.py where schedule interval was set to string none rather than type None.


phanikumv (approved):
Looks good to me.

Issue #157, opened 2022-03-24
Add limitations or differences between sync & async versions in Docs


kaxil:
We should add differences of Sync version vs Async version of an Operator/Sensor in the docstrings.

For example,

- `HttpSensorAsync` will fall back to the "sync" version when `response_check` is passed
- The logs for KubernetesOperatorAsync won't be available in webserver until the task is actually completed. (which will be fixed in https://github.com/astronomer/astronomer-providers/pull/139 :) )

PR #158, opened 2022-03-24
Document limitations of some async operators

kaxil:
closes https://github.com/astronomer/astronomer-providers/issues/157

**With this change**:

image

image


phanikumv (approved):
Looks good

Issue #119, opened 2022-03-14
Implement Async for native Spark Operators


phanikumv:

kaxil:
If we have too many issues with it, we can park it for now, or use LivyOperator as an alternative to submit Spark jobs

phanikumv:
We've got the local connection working for the SparkSubmitOperator yesterday post further analysis. I think this is story is unblocked for now @sunank200 , @bharanidharan14 - please comment.

sunank200:
Yes, `spark-submit` works for the deploy-mode `local`. So `SparkSubmitOperator` example dag works with local spark setup. For `SparkSqlOperator` i am setting up the hive along spark to create a table to test the example DAG.

bharanidharan14:
on 7-03-2022 and 21-03-2022 Successfully Installed spark in the airflow worker to run the spark-submit job and Created the spark cluster container tried running the sample spark job via airflow worker, I was facing some issues with container resource allocation issue.
22-03-2022 Tried creating separate docker image, created spark master, worker cluster container and tried submitting the job via airflow worker it was able to pick up the job but there is python version mismatch

bharanidharan14:
Working on implementing Spark submit operator async

sunank200:
`SparkSqlOperator` requires the `spark-sql` to work when it's installed along in the docker. Right now, for deployment mode as `local`, the `SparkSqlOperator` fails as `spark-sql`  fails with the following error:

> `22/03/24 05:12:55 ERROR Utils: Uncaught exception in thread shutdown-hook-0
> java.lang.InterruptedException
>         at java.util.concurrent.locks.AbstractQueuedSynchronizer.doAcquireSharedNanos(AbstractQueuedSynchronizer.java:1039)
>         at java.util.concurrent.locks.AbstractQueuedSynchronizer.tryAcquireSharedNanos(AbstractQueuedSynchronizer.java:1328)
>         at java.util.concurrent.CountDownLatch.await(CountDownLatch.java:277)
>         at org.sparkproject.jetty.server.AbstractConnector.doStop(AbstractConnector.java:371)
>         at org.sparkproject.jetty.server.AbstractNetworkConnector.doStop(AbstractNetworkConnector.java:88)
>         at org.sparkproject.jetty.server.ServerConnector.doStop(ServerConnector.java:246)
>         at org.sparkproject.jetty.util.component.AbstractLifeCycle.stop(AbstractLifeCycle.java:94)
[...]

PR #162, opened 2022-03-24
Implemented Async Spark submit operator

bharanidharan14:
story id : https://github.com/astronomer/astronomer-providers/issues/119

Operator: SparkSubmitOperatorAsync

- Implemented Spark submit operator in Master mode  local
- Added Async Operator, Trigger, hooks
- Added test case for operator, trigger and hooks
- Covered mypy fix, doc strings

### DIFF
--- a/astronomer/providers/apache/spark/example_dags/example_spark_dag.py
+++ b/astronomer/providers/apache/spark/example_dags/example_spark_dag.py
@@ -2,6 +2,7 @@
 Example Airflow DAG to submit Apache Spark applications using
 `SparkSubmitOperator`.
 """
+import os
 from datetime import datetime
 
 from airflow import DAG
@@ -10,6 +11,8 @@ from airflow.operators.dummy import DummyOperator
 from astronomer.providers.apache.spark.operators.spark_sumbit import (
     SparkSubmitOperatorAsync,
 )
+
+SPARK_CONN_ID = os.environ.get("SPARK_CONN_ID", "spark_default")
 
 with DAG(
     dag_id="example_spark_operator",
@@ -27,6 +30,7 @@ with DAG(
     submit_job_sync = SparkSubmitOperatorAsync(
         application="/opt/spark-3.1.3-bin-hadoop3.2/examples/src/main/python/pi.py",
         task_id="submit_job_async",
+        conn_id=SPARK_CONN_ID,
     )
     # [END howto_operator_spark_submit_async]
 
--- a/astronomer/providers/apache/spark/example_dags/example_spark_dag.py
+++ b/astronomer/providers/apache/spark/example_dags/example_spark_dag.py
@@ -1,0 +1,37 @@
+"""
+Example Airflow DAG to submit Apache Spark applications using
+`SparkSubmitOperator`.
+"""
+from datetime import datetime
+
+from airflow import DAG
+from airflow.operators.dummy import DummyOperator
+
+from astronomer.providers.apache.spark.operators.spark_sumbit import (
+    SparkSubmitOperatorAsync,
+)
+
+with DAG(
+    dag_id="example_spark_operator",
+    schedule_interval=None,
+    start_date=datetime(2021, 1, 1),
+    catchup=False,
+    tags=["example", "async", "spark"],
+) as dag:
+
+    # [START howto_Dummy_operator_start]
+    start = DummyOperator(task_id="start")
+    # [END howto_Dummy_operator_start]
+
+    # [START howto_operator_spark_submit_async]
+    submit_job_sync = SparkSubmitOperatorAsync(
+        application="/opt/spark-3.1.3-bin-hadoop3.2/examples/src/main/python/pi.py",
+        task_id="submit_job_async",
+    )
+    # [END howto_operator_spark_submit_async]
+
+    # [START howto_Dummy_operator_end]
+    end = DummyOperator(task_id="end")
+    # [END howto_Dummy_operator_end]
+
+    start >> submit_job_sync >> end
--- a/astronomer/providers/apache/spark/hooks/spark_submit.py
+++ b/astronomer/providers/apache/spark/hooks/spark_submit.py
@@ -1,5 +1,4 @@
 import asyncio
-import os
 from typing import Any, Dict, Optional
 
 from airflow import AirflowException
@@ -13,7 +12,7 @@ class SparkSubmitHookAsync(SparkSubmitHook):
     supplied.
 
     :param conf: Arbitrary Spark configuration properties
-    :param spark_conn_id: refer `spark connection id <howto/connection:spark>` as configured
+    :param conn_id: refer `spark connection id <howto/connection:spark>` as configured
         in Airflow administration. When an invalid connection_id is supplied, it will default
         to yarn.
     :param files: Upload additional files to the executor running the job, separated by a
@@ -68,11 +67,6 @@ class SparkSubmitHookAsync(SparkSubmitHook):
         # to build the spark submit command
 
         spark_cmd = " ".join(spark_submit_cmd)
-
-        if self._env:
-            env = os.environ.copy()
-            env.update(self._env)
-            kwargs["env"] = env
 
         self._submit_sp = await asyncio.create_subprocess_shell(
             spark_cmd,
--- a/astronomer/providers/apache/spark/hooks/spark_submit.py
+++ b/astronomer/providers/apache/spark/hooks/spark_submit.py
@@ -1,0 +1,163 @@
+import asyncio
+import os
+from typing import Any, Dict, Optional
+
+from airflow import AirflowException
+from airflow.providers.apache.spark.hooks.spark_submit import SparkSubmitHook
+
+
+class SparkSubmitHookAsync(SparkSubmitHook):
+    """
+    This Async hook is a wrapper around the spark-submit binary to kick off a spark-submit job.
+    It requires that the "spark-submit" binary is in the PATH or the spark_home to be
+    supplied.
+
+    :param conf: Arbitrary Spark configuration properties
+    :param spark_conn_id: refer `spark connection id <howto/connection:spark>` as configured
+        in Airflow administration. When an invalid connection_id is supplied, it will default
+        to yarn.
+    :param files: Upload additional files to the executor running the job, separated by a
+        comma. Files will be placed in the working directory of each executor.
+        For example, serialized objects.
+    :param py_files: Additional python files used by the job, can be .zip, .egg or .py.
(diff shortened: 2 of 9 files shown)


astronomer/providers/apache/spark/example_dags/example_spark_dag.py
```diff
@@ -0,0 +1,37 @@
...
+    tags=["example", "async", "spark"],
+) as dag:
+
+    # [START howto_Dummy_operator_start]
+    start = DummyOperator(task_id="start")
+    # [END howto_Dummy_operator_start]
+
+    # [START howto_operator_spark_submit_async]
+    submit_job_sync = SparkSubmitOperatorAsync(
+        application="/opt/spark-3.1.3-bin-hadoop3.2/examples/src/main/python/pi.py",
```
pankajastro, line 28: How do we create pi.py. This will need to be available in staging to run the integration test

bharanidharan14: We don't need to create this it will be part of the spark example folders

astronomer/providers/apache/spark/example_dags/example_spark_dag.py
```diff
@@ -0,0 +1,37 @@
...
+    catchup=False,
+    tags=["example", "async", "spark"],
+) as dag:
+
+    # [START howto_Dummy_operator_start]
+    start = DummyOperator(task_id="start")
+    # [END howto_Dummy_operator_start]
+
+    # [START howto_operator_spark_submit_async]
+    submit_job_sync = SparkSubmitOperatorAsync(
```
pankajastro, line 27: I think this operator has optional  `conn_id` param that we should take from env variable

bharanidharan14: Added

astronomer/providers/apache/spark/hooks/spark_submit.py
```diff
@@ -0,0 +1,157 @@
...
+            line = await stream.readline()
+            if line:
+                line = line.decode().strip()
+                line = line.strip()
+                self.log.info(line)
+                # TODO
+                # Need to implement the cluster mode yarn, standalone cluster mode, kubernetes
+                # Need to get the application id if cluster_mode is in yarn mode
+                # Need to get the driver id if cluster_mode is in standalone and kubernetes
+            else:
```
pankajastro, line 114: Can we remove else stmt something like
```
if line:
    line = line.decode().strip()
     line = line.strip()
     self.log.info(line)
     # TODO
     # Need to implement the cluster mode yarn, standalone cluster mode, kubernetes
     # Need to get the application id if cluster_mode is in yarn mode
     # Need to get the driver id if cluster_mode is in standalone and kubernetes

break

File: astronomer/providers/apache/spark/hooks/spark_submit.py [hunk 4]
```diff
@@ -0,0 +1,157 @@
...
+        """
+        driver_status = self.track_driver_status()
+        if driver_status not in ["FINISHED", "UNKNOWN", "KILLED", "FAILED", "ERROR"]:
+            return True
+        return False
+
+    def get_driver_status(self) -> Dict[str, Optional[str]]:
+        """Get the driver status and based on the driver status returns the message and status"""
+        msg = ""
+        if self._driver_status == "SUBMITTED":
```
pankajastro, line 141: This is also ok, I'm just thinking if we can use some pattern matching like https://docs.python.org/3.10/whatsnew/3.10.html#pep-634-structural-pattern-matching

astronomer/providers/apache/spark/triggers/spark_submit.py
```diff
@@ -0,0 +1,194 @@
...
+                "application_args": self.application_args,
+                "env_vars": self.env_vars,
+                "verbose": self.verbose,
+                "spark_binary": self.spark_binary,
+            },
+        )
+
+    async def run(self) -> AsyncIterator["TriggerEvent"]:  # type: ignore[override]
+        """
+        Makes a series of connections to snowflake to get the status of the query
```
pankajastro, line 147: why snowflake ?

astronomer/providers/apache/spark/hooks/spark_submit.py
bharanidharan14: python doesn't have any switch case by default it and elif acts as switch case in python, whatever you suggested I am not sure on the lower python version

astronomer/providers/apache/spark/hooks/spark_submit.py
bharanidharan14: No, we can't break outside of the if condition I guess because its stream of output , its like executing the shell command it will the function will send stream of output

astronomer/providers/apache/spark/triggers/spark_submit.py
bharanidharan14: Fixed

astronomer/providers/apache/spark/hooks/spark_submit.py
pankajastro: ok, then let's keep it as it is now.

astronomer/providers/apache/spark/hooks/spark_submit.py
pankajastro: got it

astronomer/providers/apache/spark/triggers/spark_submit.py
```diff
@@ -0,0 +1,194 @@
...
+        )
+
+    async def run(self) -> AsyncIterator["TriggerEvent"]:  # type: ignore[override]
+        """
+        Makes a call to SparkSubmitHookAsync class and execute the spark-submit job in async mode
+        and based on the status it polls
+        """
+        hook = self._get_async_hook()
+        try:
+            await hook.spark_submit_job(self.application)
```
phanikumv, line 152: I think we should not submit the job on the Triggerer. We should only track it. Job submission should happen on the Operator.

dev/Dockerfile
```diff
@@ -4,6 +4,49 @@ FROM ${IMAGE_NAME}
...
+
+#ENV JAVA_HOME=/usr/lib/jvm/java-8-openjdk-arm64/
+ENV JAVA_HOME=/usr/lib/jvm/adoptopenjdk-8-hotspot-amd64/jre/
+
+
+# --------------------------------------------------------
+# SPARK
+# --------------------------------------------------------
+
+ENV SPARK_VERSION=spark-3.1.3-bin-hadoop3.2
```
phanikumv, line 37: This will make our dev environment heavy. Lets remove the logic for spinning up of Spark cluster from the Dockerfile, after we figure out how to make connections to remote Spark clusters

astronomer/providers/apache/spark/example_dags/example_spark_dag.py
```diff
@@ -0,0 +1,41 @@
...
+from datetime import datetime
+
+from airflow import DAG
+from airflow.operators.dummy import DummyOperator
+
+from astronomer.providers.apache.spark.operators.spark_sumbit import (
+    SparkSubmitOperatorAsync,
+)
+
+SPARK_CONN_ID = os.environ.get("SPARK_CONN_ID", "spark_default")
```
pankajkoti, line 15:
```suggestion
SPARK_CONN_ID = os.getenv("SPARK_CONN_ID", "spark_default")
```

astronomer/providers/apache/spark/hooks/spark_submit.py
```diff
@@ -0,0 +1,157 @@
...
+            if line:
+                line = line.decode().strip()
+                line = line.strip()
+                self.log.info(line)
+                # TODO
+                # Need to implement the cluster mode yarn, standalone cluster mode, kubernetes
+                # Need to get the application id if cluster_mode is in yarn mode
+                # Need to get the driver id if cluster_mode is in standalone and kubernetes
+            else:
+                break
```
pankajkoti, line 115:
```suggestion
        line = await stream.readline()
        while line:
            line = line.decode().strip()
            self.log.info(line)
            # TODO
            # Need to implement the cluster mode yarn, standalone cluster mode, kubernetes
            # Need to get the application id if cluster_mode is in yarn mode
            # Need to get the driver id if cluster_mode is in standalone and kubernetes
            line = await stream.readline()
```


In general, while True is not recommended as make the code and breaking condition unclear and also potentially a risk of the code running in an infinite loop.

astronomer/providers/apache/spark/hooks/spark_submit.py
```diff
@@ -0,0 +1,157 @@
...
+
+    def track_driver_status(self) -> Optional[str]:
+        """Track driver status by driver id"""
+        if self._should_track_driver_status:
+            if self._driver_id is None:
```
pankajkoti, line 120:
```suggestion
            if not self._driver_id:
```
Can we use this instead? Python will ensure that it has non-null, non-empty value (in case of string: empty strings, in case of boolen: False and so on)

astronomer/providers/apache/spark/hooks/spark_submit.py
pankajkoti: We can have a dictionary/map of status to message and then get the value for the _driver_status key from the map just in case if we want to avoid so many if-elifs. Also, make it easier to add/remove keys from the map and makes it quick & easier to read.

astronomer/providers/apache/spark/hooks/spark_submit.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        Processes the log and extracts useful information out of it.
+
+        :param stream: subprocess stream bytes
+        """
+        while True:
+            line = await stream.readline()
+            if line:
+                line = line.decode().strip()
+                line = line.strip()
```
pankajkoti, line 108: Aren't we already stripping on the previous line?

astronomer/providers/apache/spark/operators/spark_sumbit.py
```diff
@@ -0,0 +1,105 @@
...
+        """
+        Callback for when the trigger fires - returns immediately.
+        Relies on trigger to throw an exception, otherwise it assumes execution was
+        successful.
+        """
+        if event:
+            if event["status"] == "ERROR":
+                raise AirflowException(event["message"])
+        self.log.info("%s completed successfully.", self.task_id)
+        return None
```
pankajkoti, line 105: Is this a convention we follow across our code to have an explicit `return None` at the end of a function? Python by default returns None if not specified.

pankajkoti (commented):
Minor suggestions. Thanks!

bharanidharan14:
Currently it is hold, Async spark submit job is done only for local execution

PR #195, opened 2022-04-06
Update CONTRIBUTING.rst to show  trigger's  JSON-serializable value limitation.

rajaths010494:
- example to show class object shouldn't be passed to trigger from operator because of JSON-serializable.


phanikumv (commented):
We need to mention that the first step to implementing an async operator is to check whether  it is possible to retrieve an execution handle or a unique ID through which we can poll in the trigger.

rajaths010494:
> We need to mention that the first step to implementing an async operator is to check whether it is possible to retrieve an execution handle or a unique ID through which we can poll in the trigger.

Done

CONTRIBUTING.rst
```diff
@@ -268,6 +268,23 @@ Considerations while writing Async or Deferrable Operator
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 - Writing a deferrable or async operator takes a bit more work. There are some main points to consider:
     - Deferrable Operators & Triggers rely on more recent asyncio features, and as a result only work on Python 3.7 or higher.
+    - Any Deferrable Operator implementation must use an execution handle in order to poll for the status in the Trigger.
```
kaxil, line 271:
```suggestion
    - Any Deferrable **Operator** implementation needs the API used to give you a unique identifier in order to poll for the status in the Trigger. This does not affect creating an async Sensor as "sensors" are just poll-based whereas "Operators" are "Submit + Poll" operation.
```

CONTRIBUTING.rst
```diff
@@ -268,6 +268,23 @@ Considerations while writing Async or Deferrable Operator
...
+                timeout=self.execution_timeout,
+                trigger=BigQueryGetDataTrigger(
+                    conn_id=self.gcp_conn_id,
+                    job_id=self.job_id,
+                    dataset_id=self.dataset_id,
+                    table_id=self.table_id,
+                    project_id=hook.project_id,
+                ),
+                method_name="execute_complete",
+            )
```
kaxil, line 287: .. code-block:: python

        job = self._submit_job(hook, configuration=configuration)
        self.job_id = job.job_id
        self.defer(
            timeout=self.execution_timeout,
            trigger=BigQueryGetDataTrigger(
                conn_id=self.gcp_conn_id,
                job_id=self.job_id,
                dataset_id=self.dataset_id,
                table_id=self.table_id,
                project_id=hook.project_id,
            ),
        )

kaxil (approved):

PR #212, opened 2022-04-09
Add `flake8` check for logging in Pre-commit Hooks

kaxil:
This commit adds flake8 plugin for logging - https://github.com/globality-corp/flake8-logging-format in pre-commit Hooks so that we always use Lazy loading of args in logs.


pankajkoti (commented):
Amazing! ❤️

PR #220, opened 2022-04-11
Fix the pre-commit dependency check

### LINKED ISSUE #218: Add apache.livy in all extras and add a pre-commit hook
Fix the pre-commit dependency check script to catch missing dependencies

phanikumv:
This PR fixes a bug in the pre commit dependency config check script. The current logic uses a one side diff between two sets (`expected_all_extra` - `found_all_extra`).However it won't work in cases when the set `found_all_extra` contains an additional element which is not in set `expected_all_extra`.

Closes #218


kaxil:
Good work on this @phanikumv

PR #232, opened 2022-04-14
Add implementation for OpenLineage custom extractor for BigQuery operator

pankajkoti:
Export job_id as XCOM for the BigQueryInsertJobOperatorAsync operator
and add an OpenLineage custom extractor for the same.

The custom extractor is an extension of [OpenLineage BigQuery extractor](https://github.com/OpenLineage/OpenLineage/blob/main/integration/airflow/openlineage/airflow/extractors/bigquery_extractor.py) to include 
`BigQueryInsertJobOperatorAsync`. 
The OpenLineage extractor has a [common implementation](https://github.com/OpenLineage/OpenLineage/blob/cb2687aeb723b1bde0ec335e6ec46ea40b111a6f/integration/common/openlineage/common/provider/bigquery.py#L141) to fetch statistics for 
big query with the given `job_id` which is fetched from `XCOM` in the case of Airflow tasks.
So, I believe, all we need to do is push the job_id as `XCOM` for the task 
which is pulled by OpenLineage and the metadata is fetched & extracted 
by the common implementation.

Some additional OpenLineage specific environment variables need to be set 
which are included as part of the `docker-compose.yaml` file  change in the commit.

Also, adding a flask based local Lineage server where we can forward extraction
events for local development.


pankajkoti:
I believe the changes we see in the PR are the needed changes for the custom extractor. However, I don't see relevant statistics available in the Datakin UI with this setup. I have a call with Maciej Obuchowski and will try to understand from him if I am missing something.

pankajkoti:
Also, I have pushed the commit without fixing the static checks. I will fix those once I confirm that this is the right approach to go with. Din't want to spend time on fixing the static checks initially.

astronomer/providers/google/cloud/extractors/bigquery_extractor.py
```diff
@@ -0,0 +1,94 @@
+# SPDX-License-Identifier: Apache-2.0.
```
phanikumv, line 1: do we need this license to be added? we are currently not adding any license for providers repo files.

astronomer/providers/google/cloud/extractors/bigquery_extractor.py
```diff
@@ -0,0 +1,94 @@
...
+
+    @classmethod
+    def get_operator_classnames(cls) -> List[str]:
+        return ["BigQueryOperator", "BigQueryExecuteQueryOperator", "BigQueryInsertJobOperatorAsync"]
+
+    def extract(self) -> Optional[TaskMetadata]:
+        return None
+
+    def extract_on_complete(self, task_instance) -> Optional[TaskMetadata]:
+        log.debug(f"extract_on_complete({task_instance})")
```
phanikumv, line 45: can you add docstrings for each method explaining what it does, what parameters it accepts etc.?

astronomer/providers/google/cloud/extractors/bigquery_extractor.py
```diff
@@ -0,0 +1,94 @@
...
+    """Internal SQL context for holding query parser results"""
+
+    sql: str = attr.ib()
+    inputs: Optional[str] = attr.ib(default=None)
+    outputs: Optional[str] = attr.ib(default=None)
+    parser_error: Optional[str] = attr.ib(default=None)
+
+
+class BigQueryExtractor(BaseExtractor):
+    def __init__(self, operator):
```
phanikumv, line 34: add docstrings to explain what this class is supposed to do?

astronomer/providers/google/cloud/triggers/bigquery.py
```diff
@@ -68,6 +68,7 @@ async def run(self) -> AsyncIterator["TriggerEvent"]:  # type: ignore[override]
                 if response_from_hook == "success":
                     yield TriggerEvent(
                         {
+                            "jobId": self.job_id,
```
phanikumv, line 71: do we need to modify the unit tests to account for this additional parameter?

dev/local_flask_lineage_server.py
```diff
@@ -0,0 +1,31 @@
...
+        content = json.dumps(js, sort_keys=True, indent=4)
+
+        date = parse(js["eventTime"]).date()
+        job_name = js["job"]["name"]
+        event_name = f"{date}-{job_name}"
+    except Exception:
+        content = str(request.data, "UTF-8")
+    file_path = f"{DUMPS_DIR}/{event_name}.json"
+
+    print(f"Written event {event_name} to file {file_path}")
```
pankajastro, line 28: How about using log instead of print ?

astronomer/providers/google/cloud/extractors/bigquery_insertjoboperatorasync_extractor.py
```diff
@@ -0,0 +1,93 @@
...
+        inputs = stats.inputs
+        output = stats.output
+        run_facets = stats.run_facets
+
+        return TaskMetadata(
+            name=get_job_name(task=self.operator),
+            inputs=[ds.to_openlineage_dataset() for ds in inputs],
+            outputs=[output.to_openlineage_dataset()] if output else [],
+            run_facets=run_facets,
+        )
```
phanikumv, line 93: add unit tests for this code

astronomer/providers/google/cloud/extractors/bigquery_extractor.py
pankajkoti: Removed.

astronomer/providers/google/cloud/extractors/bigquery_extractor.py
pankajkoti: done.

astronomer/providers/google/cloud/triggers/bigquery.py
pankajkoti: Yes, accommodated the required change.

dev/local_flask_lineage_server.py
pankajkoti: done.

astronomer/providers/google/cloud/extractors/bigquery_insertjoboperatorasync_extractor.py
pankajkoti: Added.

astronomer/providers/google/cloud/extractors/bigquery_extractor.py
pankajkoti: done.

pankajkoti:
Looks like something weird causing tests to fail for code out of scope of this PR. Closing the PR and opened a new one with a new branch cut from `main` : #290

PR #233, opened 2022-04-14
Capitalise module level constants

pankajkoti:


docs/conf.py
```diff
@@ -18,37 +18,37 @@
...
 
 sys.path.insert(0, os.path.abspath(".."))
-repo_dir = Path(__file__).parent.parent
+REPO_DIR = Path(__file__).parent.parent
 
 # -- Project information -----------------------------------------------------
 
-project = "Astronomer Providers"
-author = "Astronomer Inc."
+PROJECT = "Astronomer Providers"
```
kaxil, line 25: We can't do the others though as they are using by Sphinx.

We only can do that are used internally, like `REPO_DIR`. Feel free to make it internal if you want to by adding `_` i.e. `_REPO_DIR`

docs/conf.py
```diff
@@ -34,7 +34,7 @@
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = [
+_EXTENSIONS = [
```
kaxil, line 37:
```suggestion
extensions = [
```

docs/conf.py
```diff
@@ -161,5 +161,5 @@ def _inclue_init_in_docs(app, what, name, obj, would_skip, options):
 
 def setup(app):
     """Sphinx Application API"""
-    if "autoapi.extension" in extensions:
+    if "autoapi.extension" in _EXTENSIONS:
```
kaxil, line 164:
```suggestion
    if "autoapi.extension" in extensions:
```

docs/conf.py
kaxil: This is also used by Sphinx for extensions

kaxil: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-extensions

kaxil (approved):

PR #236, opened 2022-04-14
Add pre-commit hooks for detecting private key & AWS Creds

kaxil:
This is not bullet proof but will prevent us adding Secrets in some scenarios which is good enough.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,7 +6,7 @@ repos:
       - id: check-useless-excludes
 
   - repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v4.1.0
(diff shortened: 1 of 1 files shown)


pankajastro:
Just wondering if it also restricts in case of dummy value and env key match to AWS. That may be problematic

kaxil:
> Just wondering if it also restricts in case of dummy value and env key match to AWS. That may be problematic

It is actually smart about it - https://github.com/pre-commit/pre-commit-hooks/blob/main/pre_commit_hooks/detect_aws_credentials.py

It only detects the only that are set locally too in Environment Variables or AWS Cred files :)

phanikumv (approved):

Issue #237, opened 2022-04-14
Fix `LivyOperatorAsync` example DAG


phanikumv:
Refer to https://github.com/astronomer/astronomer-providers/pull/235 and implement additional suggestions as part of the PR

PR #252, opened 2022-04-20
Fix LivyOperatorAsync example DAG

pankajkoti:
The LivyOperatorAsync example DAG is failing as part of the integration
tests on staging environment. The reason for the failure is that the DAG is
referring to AWS environment variables different than what has been
configured in the staging deployment. This change renames such AWS 
environment variables to be consistent with other example DAG and the
same have been set in the staging deployment environment.

Closes: https://github.com/astronomer/astronomer-providers/issues/237

### DIFF
--- a/astronomer/providers/apache/livy/example_dags/example_livy.py
+++ b/astronomer/providers/apache/livy/example_dags/example_livy.py
@@ -30,9 +30,9 @@ PEM_FILENAME = os.environ.get("PEM_FILENAME", "providers_team_keypair")
 PRIVATE_KEY = Variable.get("providers_team_keypair")
 
 AWS_S3_CREDS = {
-    "aws_access_key_id": os.environ.get("AWS_ACCESS_KEY", "sample_aws_access_key_id"),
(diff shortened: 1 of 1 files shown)


pankajkoti:
Resolved initial issues where inconsistent variable names were causing connection error.

We still are facing issues, with the following error being next:
`{'Code': 'VALIDATION_ERROR', 'Message': 'The given SSH key name was invalid'}` Will look into this next.

pankajkoti:
> {'Code': 'VALIDATION_ERROR', 'Message': 'The given SSH key name was invalid'}

This issue was due to inconsistent regions. @sunank200 helped pointing this out and the example DAG runs fine now with `us-east-2`

sunank200:
`aws-region-name` has to be `us-east-2`. This issue should be fixed now.

pankajastro (approved):

rajaths010494 (approved):

phanikumv (changes requested):
Please update the description to summarize what we are doing

phanikumv (approved):

Issue #98, opened 2022-03-07
Implement Async `AzureDataFactoryPipelineRunStatusSensor`


phanikumv:
Async version of `AzureDataFactoryPipelineRunStatusSensor`: https://airflow.apache.org/docs/apache-airflow-providers-microsoft-azure/stable/_api/airflow/providers/microsoft/azure/sensors/data_factory/index.html#airflow.providers.microsoft.azure.sensors.data_factory.AzureDataFactoryPipelineRunStatusSensor

**Acceptance Criteria**:
- [x] Unit Tests coverage in the PR (80% Code Coverage -- We will need to add CodeCov separately to measure code cov) with all of them passing 
- [x] Example DAG using the async Operator that can be used to run Integration tests that are parametrized via Environment variables. Example - https://github.com/apache/airflow/blob/8a03a505e1df0f9de276038c5509135ac569a667/airflow/providers/google/cloud/example_dags/example_bigquery_to_gcs.py#L33-L35
- [x] Add proper docstrings for each of the methods and functions including Example DAG on how it should be used (populate 
- [x] Exception Handling in case of errors 
- [ ] Improve the OSS Docs to make sure it covers the following:
   - [x] Has an example DAG for the sync version
   - [ ] How to add a connection via Environment Variable & explain each of the fields. Example - https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/connections/postgres.html
   - [ ] How to use Guide for the Operator - example: https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/operators/postgres_operator_howto_guide.html

bharanidharan14:
Raised Draft PR 
https://github.com/astronomer/astronomer-providers/pull/253

- Add async Sensor, Hook, Trigger
- Add Example Dag
- Add unit Test case
- cover code coverage above 95%
- cover mypy
- Add doc string

bharanidharan14:
working on the self sufficient DAG to create pipeline and remove script for data factory, group, storage. currently it has creating script for

- creating resource group and if not present
- creating  Data factory
- create storage linked service 
- create blob dataset(input, output)

bharanidharan14:
Finished with self sufficient DAG 

created resource group and if not present,  Data factory,  storage linked service, blob dataset(input, output) via script and run the pipeline and delete whatever resource is been created via script

PR #253, opened 2022-04-20
Implement Async Azure data factory pipeline run status sensor

bharanidharan14:
closes: #98 

- Add async Sensor, Hook, Trigger
- Add Example DAG
- Add unit Test case


astronomer/providers/microsoft/azure/hooks/data_factory.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    An Async Hook connects to Azure DataFactory to perform pipeline operations
+
+    :param azure_data_factory_conn_id: The :ref:`Azure Data Factory connection id<howto/connection:adf>`.
+    """
+
+    def __init__(self, azure_data_factory_conn_id: str):
+        self._async_conn: DataFactoryManagementClient = None
+        self.conn_id = azure_data_factory_conn_id
+        super().__init__()
```
pankajastro, line 23: we should pass azure_data_factory_conn_id to super class as well

phanikumv (changes requested):
Please correct the description , summarize what is being done as part of this PR, correct the typos and it is not referring the issue correctly using `Closes #id`

phanikumv (approved):
looks good to me

Issue #257, opened 2022-04-21
Fix failing example_emr_eks_pi_job example DAG


pankajkoti:
The example_emr_eks_pi_job is failing as part of our intergration tests run.
As part of the investigation logs, the shell script Bash Operator seems to having some errors not being raised up and letting the task state shown to be successful.
Understand the DAG, debug the issue across tasks in the DAG and make it working.

pankajkoti:
Have set failure raise for shell script in the PR: https://github.com/astronomer/astronomer-providers/pull/259

However, cloud formation template is not succeeding with below failure reason:
` AWS::EKS::Nodegroup/ManagedNodeGroup: CREATE_FAILED – "Resource handler returned message: \"[Issue(Code=AsgInstanceLaunchFailures, Message=You've reached your quota for maximum Fleet Requests for this account. Launching EC2 instance failed., ResourceIds=[eks-ng-a0731f88-2ec026b0-8620-cc3d-82e3-dee3ed01e00a]), Issue(Code=NodeCreationFailure, Message=Instances failed to join the kubernetes cluster, ResourceIds=[DUMMY_11448ab7-f2d5-42a7-a6cc-0d74ccb2e8e9, DUMMY_1bb6f31e-92c8-43d5-9ca3-8e815cbff947, DUMMY_2a5d16f9-291c-4d58-bf35-2eb5c0db88f5, DUMMY_4e4b490f-ef51-4d5b-87a5-3a5141ac6f9c, DUMMY_77a3e8d7-3af0-49e6-afa5-7c5ebc9a0eeb, DUMMY_7e88602e-c7f8-4d6a-943b-5f5078a4eb3f, DUMMY_a129f26a-1a81-4b5e-9c84-2285d8734a8f, DUMMY_abb67e19-d841-4bd0-ad05-5418f73149d1, DUMMY_ae850f0d-0693-4772-9d9b-4f4d442bab0e, DUMMY_ff345f8f-3468-4f2b-b8cf-ac1560c4150d])] (Service: null, Status Code: 0, Request ID: null, Extended Request ID: null)\" (RequestToken: 98ef1611-2341-fd34-ef08-4e8970c34e47, HandlerErrorCode: GeneralServiceException)"`

pankajkoti:
Adding @bharanidharan14 as assignee too as he is trying this on his local.

pankajkoti:
[...]

PR #259, opened 2022-04-21
Fix EMR EKS example DAG

pankajkoti:
The PR addresses following changes:
1. Fail the shell script if any of the commands fail
        Currently, the shell script is silently ignoring failure of
        commands. As a result, the Airflow task calling this shell script
        is marked as 'success' which should not happen because the DAG
        proceeds to execute subsequent tasks resulting in late failures.
        With this change, we want to report error in any command upfront
        so that the task is marked 'failed' and the DAG does not continue
        executing downstream tasks.
2. Add separate Dockerfile for EMR-EKS example DAG image
        The current Dockerfile for local development is missing a few libraries
        required for the EMR-EKS example DAG. So, create a separate Dockerfiles
        with the required libraries to help aid faster local development.

Closes: #257


Makefile
```diff
@@ -27,6 +27,9 @@ clean: ## Remove all the containers along with volumes
 build: ## Build the Docker image (ignoring cache)
 	docker build -f dev/Dockerfile . -t astronomer-providers-dev:latest --no-cache
 
+build-emr_eks_container_example_dag-image: ## Build the Docker image for EMR EKS containers example DAG (ignoring cache)
```
pankajastro, line 30: Instead of only emr_eks what about doing for integration test i.e all examples?

pankajkoti: good idea, we can do it in a separate PR as this issue & PR is related to EMR-EKS DAG failure.

bharanidharan14 (approved):
Overall looks good to me

phanikumv (changes requested):
still  needs fix to make it run end to end successfully.

pankajkoti:
Cloudformation template is failing with below reason:
`
AWS::EKS::Nodegroup/ManagedNodeGroup: CREATE_FAILED – "Resource handler returned message: \"[Issue(Code=AsgInstanceLaunchFailures, Message=You've reached your quota for maximum Fleet Requests for this account. Launching EC2 instance failed., ResourceIds=[eks-ng-a0731f88-2ec026b0-8620-cc3d-82e3-dee3ed01e00a]), Issue(Code=NodeCreationFailure, Message=Instances failed to join the kubernetes cluster, ResourceIds=[DUMMY_11448ab7-f2d5-42a7-a6cc-0d74ccb2e8e9, DUMMY_1bb6f31e-92c8-43d5-9ca3-8e815cbff947, DUMMY_2a5d16f9-291c-4d58-bf35-2eb5c0db88f5, DUMMY_4e4b490f-ef51-4d5b-87a5-3a5141ac6f9c, DUMMY_77a3e8d7-3af0-49e6-afa5-7c5ebc9a0eeb, DUMMY_7e88602e-c7f8-4d6a-943b-5f5078a4eb3f, DUMMY_a129f26a-1a81-4b5e-9c84-2285d8734a8f, DUMMY_abb67e19-d841-4bd0-ad05-5418f73149d1, DUMMY_ae850f0d-0693-4772-9d9b-4f4d442bab0e, DUMMY_ff345f8f-3468-4f2b-b8cf-ac1560c4150d])] (Service: null, Status Code: 0, Request ID: null, Extended Request ID: null)\" (RequestToken: 98ef1611-2341-fd34-ef08-4e8970c34e47, HandlerErrorCode: GeneralServiceException)"`

I tried setting instance_type to m4.xlarge both is staging deployment environment as well as running a local command:
`eksctl create cluster --name pankaj-test --with-oidc --ssh-access --ssh-public-key providers_team_keypair --instance-types=m4.xlarge`. But, it keeps on creating an instance with m4.large type and then fails with the above error. 

I checked IAM roles/policies in our account to see if they have some limitation on instance type. But, could not find anything like that.

pankajkoti:
> still needs fix to make it run end to end successfully.

@phanikumv I believe the DAG is failing due to restrictions on our AWS account and we need no change on the code to fix the failure.

phanikumv:
> > still needs fix to make it run end to end successfully.
> 
> @phanikumv I believe the DAG is failing due to restrictions on our AWS account and we need no change on the code to fix the failure.
> 
> Edit: We need some code change to pass variables to the shell script via BashOperator from the DAG

The permission issue should be resolved as we are now using our AWS account

phanikumv (approved):

PR #269, opened 2022-04-25
Use Exception sub class in azure datafactory example DAG

pankajastro:
Removed generalised exception type `Exception ` and catching  `azure.core.exceptions.HttpResponseError` which is raise by azure SDK. Changed `logger.info` to `logger. exception` get more exception info.


phanikumv (commented):
Looks good

phanikumv (approved):

Issue #284, opened 2022-04-27
Fix example docs tag


pankajastro:
Some example has a typo i.e it has a start tag instead of end tag

PR #285, opened 2022-04-27
Fix example docs tag

pankajastro:
Fix example docstring tag typo.

Closes: #284


rajaths010494 (approved):

phanikumv (commented):
not fully clear on the impact this has on the documentation, can you explain please.

pankajastro:
> not fully clear on the impact this has on the documentation, can you explain please.

Right, currently no impact, since we are not using these tags to generate any documentation. But, since we are writing this tag I think we should keep it in the correct form.

PR #290, opened 2022-04-29
Add implementation for OpenLineage custom extractor for BigQueryInsertJobOperatorAsync

pankajkoti:
Export job_id as XCOM for the BigQueryInsertJobOperatorAsync operator
and add an OpenLineage custom extractor for the same.

The custom extractor is an extension of OpenLineage BigQuery extractor to include
BigQueryInsertJobOperatorAsync.
The OpenLineage extractor has a common implementation to fetch statistics for
big query with the given job_id which is fetched from XCOM in the case of Airflow tasks.
So, I believe, all we need to do is push the job_id as XCOM for the task
which is pulled by OpenLineage and the metadata is fetched & extracted
by the common implementation.

Some additional OpenLineage specific environment variables need to be set
which are included as part of the docker-compose.yaml file change in the commit.

Also, adding a flask based local Lineage server where we can forward extraction
events for local development.

PR #292, opened 2022-04-29
Fix failing Databricks tests

pankajkoti:
Following tests for Databricks are failing and observed in a non-related PR: #290

1. test_databricks_submit_run_operator_async
2. test_databricks_run_now_operator_async

The reason for failure is that the execute method called in these
tests is trying to push an XCOM against task instance 'ti' of
the context. Previously, context was an empty dictionary missing
the 'ti' key. We now are using the create_context utility to
generate a context object which includes the 'ti' key against which
the XCOM is getting pushed.

### DIFF
--- a/tests/databricks/operators/test_databricks.py
+++ b/tests/databricks/operators/test_databricks.py
@@ -56,7 +56,7 @@ def test_databricks_submit_run_operator_async(
     )
 
     with pytest.raises(TaskDeferred) as exc:
-        operator.execute(context)
+        operator.execute(context=create_context(operator))
 
     assert isinstance(exc.value.trigger, DatabricksTrigger), "Trigger is not a DatabricksTrigger"
 
@@ -80,7 +80,7 @@ def test_databricks_run_now_operator_async(
     )
 
     with pytest.raises(TaskDeferred) as exc:
-        operator.execute(context)
+        operator.execute(context=create_context(operator))
 
     assert isinstance(exc.value.trigger, DatabricksTrigger), "Trigger is not a DatabricksTrigger"
 


pankajkoti:
I observed Databricks tests failing as part of this PR: https://github.com/astronomer/astronomer-providers/pull/290 and hence created a separate issue and PR to fix these here. Once this is merged to main, I will get these changes into my branch.

pankajastro:
By default, the value of do_xcom_push param is false of DatabricksSubmitRunOperator, DatabricksRunNowOperator and also I can see in these particular tests we are not setting param do_xcom_push so it is weird that the test is failing.

pankajkoti:
> By default, the value of `do_xcom_push` param is false of `DatabricksSubmitRunOperator` and `DatabricksRunNowOperator` and also I can see in these particular tests we are not setting param `do_xcom_push `so it is weird that the test is failing.

Yes, weird why it's failing only in my PR: https://app.circleci.com/pipelines/github/astronomer/astronomer-providers/1445/workflows/6a04ddcb-3480-4f37-b321-0e735af0f7a4/jobs/4527

pankajastro:
> 



> > By default, the value of `do_xcom_push` param is false of `DatabricksSubmitRunOperator` and `DatabricksRunNowOperator` and also I can see in these particular tests we are not setting param `do_xcom_push `so it is weird that the test is failing.
> 
> Yes, weird why it's failing only in my PR: https://app.circleci.com/pipelines/github/astronomer/astronomer-providers/1445/workflows/6a04ddcb-3480-4f37-b321-0e735af0f7a4/jobs/4527
> 
> But, I believe, we need this fix anyway, right @pankajastro ?

wait, In my local airflow code I can see the default value for do_xocm_push param is false but in airflow main, I see it is true https://github.com/apache/airflow/blob/main/airflow/providers/databricks/operators/databricks.py#L315  @kaxil

phanikumv:
> > 
> 
> > > By default, the value of `do_xcom_push` param is false of `DatabricksSubmitRunOperator` and `DatabricksRunNowOperator` and also I can see in these particular tests we are not setting param `do_xcom_push `so it is weird that the test is failing.
> > 
> > 
> > Yes, weird why it's failing only in my PR: https://app.circleci.com/pipelines/github/astronomer/astronomer-providers/1445/workflows/6a04ddcb-3480-4f37-b321-0e735af0f7a4/jobs/4527
> > But, I believe, we need this fix anyway, right @pankajastro ?
> 
> wait, In my local airflow code I can see the default value for do_xocm_push param is false but in airflow main, I see it is true https://github.com/apache/airflow/blob/main/airflow/providers/databricks/operators/databricks.py#L315 @kaxil @phanikumv

Ideally the behaviour should be same as that of apache airflow

kaxil:
Looks like it was changed recently - https://github.com/apache/airflow/pull/22541 on the main branch of Airflow which is also released as part of databircks provider 2.6.0 - https://airflow.apache.org/docs/apache-airflow-providers-databricks/stable/index.html#id1

pankajastro:
> Looks like it was changed recently - [apache/airflow#22541](https://github.com/apache/airflow/pull/22541) on the main branch of Airflow which is also released as part of databircks provider 2.6.0 - https://airflow.apache.org/docs/apache-airflow-providers-databricks/stable/index.html#id1

Yes, we are having this issue because of airflow databricks operator behaviour change. But, because of this change, there will be a backward compatibility issue for others also.

pankajkoti:
I think in general for Airflow, most places I see the value set to True. Irrespective of the value, I believe we should have the fix in this PR for the tests by creating the needed context. Please let me know if you have alternative thoughts.

phanikumv (approved):

Issue #302, opened 2022-05-02
Fix mypy issues and  fix failing tests


pankajkoti:
The context argument has been declared with type Dict[str, Any].
However, with the new release of mypy 0.950 it has started complaining
that the type should be Context in alignment with the base classes.

Some tests are failing for BigQuery Operator due to change in error messages.
Update check for error message text to fix those failing tests.

PR #303, opened 2022-05-02
Fix mypy type annotation issues

pankajkoti:
The context argument has been declared with type Dict[str, Any].
However, with the new release of mypy 0.950 it has started complaining
that the type should be Context in alignment with the base classes.

Some tests are failing for BigQuery Operator due to change in error message texts.
Update check for error message text to fix those failing tests.

Closes: #302


phanikumv (commented):
looks good

phanikumv (approved):

Issue #307, opened 2022-05-04
Use consistent method to fetch env variable Value


pankajastro:
we should either use `os.environ.get` or `os.getenv` across the repo to fetch the env variable value, not both

pankajkoti:
I would have a slight preference towards using os.getenv if everyone agrees to it.
No specific reason, just less characters.

pankajastro:
> I would have a slight preference towards using os.getenv if everyone agrees to it. No specific reason, just less characters.

Ok, I'll make it to use `os.getenv`

PR #308, opened 2022-05-04
Replace `os.environ.get` to `os.getenv`

pankajastro:
This PR will update `os.environ.get` to `os.getenv` so that fetching env will be consistent across the reop.

Closes: https://github.com/astronomer/astronomer-providers/issues/307


phanikumv (approved):

pankajkoti (commented):
@pankajastro Thank you for the clean up!

PR #313, opened 2022-05-06
Include timeout on EmrContainerSensorTrigger

pankajastro:
EmrContainerSensor currently does not respect max_retries parameter. After this changes, it will timeout after max_retries limit exceed 

closes: #241

### DIFF
--- a/astronomer/providers/amazon/aws/triggers/emr.py
+++ b/astronomer/providers/amazon/aws/triggers/emr.py
@@ -26,9 +26,9 @@ class EmrContainerSensorTrigger(BaseTrigger):
         self,
         virtual_cluster_id: str,
         job_id: str,
-        max_retries: Optional[int],
-        aws_conn_id: str,
-        poll_interval: int,
+        max_retries: Optional[int] = None,
+        aws_conn_id: str = "aws_default",
+        poll_interval: int = 10,
     ):
         super().__init__()
         self.virtual_cluster_id = virtual_cluster_id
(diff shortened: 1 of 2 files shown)


tests/amazon/aws/triggers/test_emr.py
```diff
@@ -171,6 +171,26 @@ async def test_emr_container_sensors_trigger_exception(mock_query_status):
     assert TriggerEvent({"status": "error", "message": "Test exception"}) in task
 
 
+@pytest.mark.asyncio
+@mock.patch("astronomer.providers.amazon.aws.hooks.emr.EmrContainerHookAsync.check_job_status")
+async def test_emr_container_sensor_trigger_timeout(mock_query_status):
+    """Assert EmrContainerOperatorTrigger max_tries exceed"""
```
phanikumv, line 177: Please correct the docstrings 
```suggestion
    """Asserts that the  EmrContainerSensorTrigger triggers correct event in case of timeout"""
```

astronomer/providers/amazon/aws/triggers/emr.py
```diff
@@ -64,6 +65,17 @@ async def run(self) -> AsyncIterator["TriggerEvent"]:  # type: ignore[override]
                 else:
                     msg = "EMR Containers sensors completed"
                     yield TriggerEvent({"status": "success", "message": msg})
+
+                if self.max_retries and try_number >= self.max_retries:
```
pankajkoti, line 69: If `max_retries` is an optional param, it has to come as a keyword param with default value as `None` and after positional params in the constructor. https://github.com/astronomer/astronomer-providers/pull/313/files#diff-0a94dcae11767f8098ecf337281ccae74019ebc185787234098d24bd1f34b3e8R29

pankajkoti (changes requested):

tests/amazon/aws/triggers/test_emr.py
pankajastro: fixed

astronomer/providers/amazon/aws/triggers/emr.py
pankajastro: Added the default value

pankajkoti (approved):

rajaths010494 (approved):

phanikumv (approved):

phanikumv (approved):

Issue #304, opened 2022-05-03
Add missing docstrings for some of the Triggers


kaxil:
We don't have docstrings for the following Triggers

```
astronomer/providers/core/triggers/external_task.py:14:class TaskStateTrigger(BaseTrigger):  # noqa: D101
astronomer/providers/core/triggers/external_task.py:72:class DagStateTrigger(BaseTrigger):  # noqa: D101
astronomer/providers/databricks/triggers/databricks.py:9:class DatabricksTrigger(BaseTrigger):  # noqa: D101
```

kaxil:
We should fix this and get it in 1.3.0 - @phanikumv

PR #316, opened 2022-05-06
Add missing docstring

pankajastro:
Add missing trigger docstring for operator/sensor/trigger
Removed `# noqa: D101`

closes: #304


astronomer/providers/databricks/operators/databricks.py
```diff
@@ -14,7 +14,149 @@
     from airflow.utils.context import Context
 
 
-class DatabricksSubmitRunOperatorAsync(DatabricksSubmitRunOperator):  # noqa: D101
+class DatabricksSubmitRunOperatorAsync(DatabricksSubmitRunOperator):
```
kaxil, line 17: This noqa is fine as we don't have an __init__ method where we add new args/kwargs

kaxil: So we can remove the docstrings below it

kaxil (approved):

astronomer/providers/databricks/operators/databricks.py
pankajastro: Thanks for fixing it.

PR #325, opened 2022-05-06
Remove Flask from requirements

kaxil:
Flask is a dev onoly requirement and is also a core requirement for Airflow.


pankajkoti:
Thank you @kaxil for informing the implicit availability of the dependency and cleaning this up :)

PR #327, opened 2022-05-06
CI: Show installed deps before running tests

kaxil:
This will help us to debug/check when CI starts failing because a new version of the library was released.


pankajkoti:
What does the CircleCI error mean here? Is it that we merged the branch, and it got deleted before the static run started? If that is the case then fine, and if not would like to be aware of the error reasons of CircleCI for my knowledge.

kaxil:
Yup that's correct

PR #335, opened 2022-05-10
Add back implementation for OpenLineage extractor for BigQueryInsertJobAsync Operator

pankajkoti:
This reverts commit 39102d01bc95fdb74635462ff3ee896b092699d8  to add back implementation 
for OpenLineage extractor for BigQueryInsertJobAsyncOperator. 
It was temporarily removed for release 1.3.0

Also, addressed Kaxil's comments on PR #290

### DIFF
--- a/astronomer/providers/google/cloud/extractors/bigquery_async_extractor.py
+++ b/astronomer/providers/google/cloud/extractors/bigquery_async_extractor.py
@@ -1,0 +1,81 @@
+from typing import Any, List, Optional
+
+from airflow.exceptions import AirflowException
+from airflow.models.taskinstance import TaskInstance
+from airflow.utils.log.logging_mixin import LoggingMixin
+from google.cloud.bigquery import Client
+from openlineage.airflow.extractors.base import BaseExtractor, TaskMetadata
+from openlineage.airflow.utils import get_job_name
+from openlineage.common.provider.bigquery import BigQueryDatasetsProvider
+
+from astronomer.providers.google.cloud.operators.bigquery import (
+    BigQueryInsertJobOperatorAsync,
+)
+
+
+class BigQueryAsyncExtractor(BaseExtractor, LoggingMixin):
+    """
+    This extractor provides visibility on the metadata of a BigQuery Insert Job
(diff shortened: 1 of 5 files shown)


bharanidharan14 (approved):

phanikumv (approved):

astronomer/providers/google/cloud/extractors/bigquery_async_extractor.py
```diff
@@ -0,0 +1,85 @@
...
+    def _get_big_query_client(self) -> Client:
+        """
+        Gets the BigQuery client to fetch job metadata.
+        The method checks whether a connection hook is available with the Airflow configuration for the operator, and
+        if yes, returns the same connection. Otherwise, returns the Client instance of``google.cloud.bigquery``.
+        """
+        hook = self.operator.hook
+        return hook.get_client(project_id=hook.project_id, location=hook.location)
+
+    def _get_xcom_bigquery_job_id(self, task_instance: TaskInstance) -> Any:
```
pankajastro, line 35: Can we keep return type `str`?

pankajkoti: The method return the ouput of `xcom_pull` method which declares the return type to `Any`. I had kept the return type to `str` as in our case the jobID will be string, but mypy complained that I have to keep it aligned with the method's return type and make it to `Any`

setup.cfg
```diff
@@ -34,7 +34,7 @@ packages = find_namespace:
 include_package_data = true
 namespace_packages = astronomer,astronomer.providers
 install_requires =
-    apache-airflow>=2.2.0
+    apache-airflow>=2.3.0
```
kaxil, line 37: Does this not work with Airflow 2.2.x @pankajkoti @phanikumv -- Bumping the pin means the providers won't be installable for users on 2.2.x which is what we don't want.

pankajkoti: We observed an issue where some piece of code was calling xcom.execution_date which(the column) does not exist in the upgraded DB for 2.3.0 of our base image.

PR #339, opened 2022-05-11
Fix aws nuke schedule

phanikumv:


pankajastro:
Why hive related docker change in this branch?

pankajkoti:
> Why hive related docker change in this branch?

Is this branch cut from `hive-setup` branch and not from main? Perhaps, that is the reason why we're seeing the changes here.

Issue #298, opened 2022-05-02
Livy example DAG failing


phanikumv:
livy_java_task is failing with below error.

`[2022-05-02, 09:43:01 UTC] {livy.py:151} INFO - Submitting job {"file": "/spark-examples.jar", "className": "org.apache.spark.examples.SparkPi", "numExecutors": 1, "conf": {"spark.shuffle.compress": "false"}} to 
[2022-05-02, 09:43:01 UTC] {base.py:68} INFO - Using connection ID 'livy_default' for task execution.
[2022-05-02, 09:43:01 UTC] {http.py:129} INFO - Sending 'POST' to url: http://ec2-18-119-156-7.us-east-2.compute.amazonaws.com:8998/batches
[2022-05-02, 09:45:10 UTC] {http.py:187} WARNING - HTTPConnectionPool(host='ec2-18-119-156-7.us-east-2.compute.amazonaws.com', port=8998): Max retries exceeded with url: /batches (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7f4d1caf9940>: Failed to establish a new connection: [Errno 110] Connection timed out')) Tenacity will retry to execute the operation
[2022-05-02, 09:45:10 UTC] {taskinstance.py:1774} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/urllib3/connection.py", line 174, in _new_conn
    conn = connection.create_connection(
  File "/usr/local/lib/python3.9/site-packages/urllib3/util/connection.py", line 95, in create_connection
    raise err
  File "/usr/local/lib/python3.9/site-packages/urllib3/util/connection.py", line 85, in create_connection
    sock.connect(sa)
TimeoutError: [Errno 110] Connection timed out

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/urllib3/connectionpool.py", line 703, in urlopen
    httplib_response = self._make_request(
  File "/usr/local/lib/python3.9/site-packages/urllib3/connectionpool.py", line 398, in _make_request
    conn.request(method, url, **httplib_request_kw)
  File "/usr/local/lib/python3.9/site-packages/urllib3/connection.py", line 239, in request
    super(HTTPConnection, self).request(method, url, body=body, headers=headers)
  File "/usr/local/lib/python3.9/http/client.py", line 1285, in request
    self._send_request(method, url, body, headers, encode_chunked)
[...]

PR #342, opened 2022-05-11
Fix ingress port authorizations for Livy & Hive example DAGs

pankajkoti:
The previous implementation was checking only for the presence
of the machine's IP address for existence of ALLOW permission
in the security group. As a result, when both Livy and Hive example
DAGs run concurrently, one of them adds the machine's IP along with
its ports to be authorized, but the second DAG sees that the IP
is already added and then skips authorizing its port. To elaborate
with an example, let's suppose the Hive example DAG runs first. It
will authorize and open port '22' and '10000' for the machine's IP.
When the Livy operator runs, it will see that rules against the
machine's IP already exist in the security group and as a result
it will skip authorizing its required port '8998' in the security
group.

To fix this, the commit removes the check on IP in the security
group and instead tries authorising the ports directly. In case,
the rule already exists, boto will throw a duplicate permission error
which we catch. Other errors are reported back to ensure that the
task is marked failed. Previously, since, the port was getting
skipped to be open, we were getting the error mentioned in issue #298 

Closes: #298

### DIFF
--- a/astronomer/providers/apache/hive/example_dags/example_hive.py
+++ b/astronomer/providers/apache/hive/example_dags/example_hive.py
@@ -15,6 +15,7 @@ from airflow.providers.amazon.aws.operators.emr import (
     EmrTerminateJobFlowOperator,
 )
 from airflow.providers.apache.hive.operators.hive import HiveOperator
+from botocore.exceptions import ClientError
 from requests import get
 
 from astronomer.providers.apache.hive.sensors.hive_partition import (
@@ -29,9 +30,10 @@ AWS_S3_CREDS = {
     "aws_secret_access_key": os.getenv("AWS_SECRET_ACCESS_KEY", "aws_secret_key"),
     "region_name": os.getenv("AWS_DEFAULT_REGION", "us-east-2"),
 }
-
+BOTO_DUPLICATE_PERMISSION_ERROR = "InvalidPermission.Duplicate"
 PEM_FILENAME = os.getenv("PEM_FILENAME", "providers_team_keypair")
 PRIVATE_KEY = Variable.get("providers_team_keypair")
+HIVE_OPERATOR_INGRESS_PORT = int(os.getenv("HIVE_OPERATOR_INGRESS_PORT", 10000))
 HIVE_SCHEMA = os.getenv("HIVE_SCHEMA", "default")
 HIVE_TABLE = os.getenv("HIVE_TABLE", "zipcode")
 HIVE_PARTITION = os.getenv("HIVE_PARTITION", "state='FL'")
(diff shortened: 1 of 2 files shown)


astronomer/providers/apache/livy/example_dags/example_livy.py
```diff
@@ -160,6 +154,11 @@ def add_inbound_rule_for_security_group(task_instance: Any) -> None:
                 }
             ],
         )
+    except ClientError as error:
+        if error.response.get("Error", {}).get("Code", "") == BOTO_DUPLICATE_PERMISSION_ERROR:
+            logging.info("Ingress for port 22 already authorized")
```
bharanidharan14, line 159: It has to be exception instead of info ?

astronomer/providers/apache/hive/example_dags/example_hive.py
```diff
@@ -198,6 +191,11 @@ def add_inbound_rule_for_security_group(task_instance: Any) -> None:
                 }
             ],
         )
+    except ClientError as error:
+        if error.response.get("Error", {}).get("Code", "") == BOTO_DUPLICATE_PERMISSION_ERROR:
+            logging.info("Ingress for port 22 already authorized")
```
bharanidharan14, line 196: It has to be exception instead of info ?

astronomer/providers/apache/hive/example_dags/example_hive.py
```diff
@@ -153,38 +155,29 @@ def add_inbound_rule_for_security_group(task_instance: Any) -> None:
...
+                    "ToPort": HIVE_OPERATOR_INGRESS_PORT,
                     "IpRanges": [{"CidrIp": str(current_docker_ip) + "/32"}],
                 }
             ],
         )
-
-        # Allow SSH traffic on port 22 and copy file to HDFS
+    except ClientError as error:
+        if error.response.get("Error", {}).get("Code", "") == BOTO_DUPLICATE_PERMISSION_ERROR:
+            logging.info("Ingress for port %s already authorized", HIVE_OPERATOR_INGRESS_PORT)
```
bharanidharan14, line 175: same here ..It has to be exception instead of info ?

astronomer/providers/apache/livy/example_dags/example_livy.py
pankajkoti: Since, we're catching the exception and is known I put it to info. We can discuss this with the team on what the log level could be here.

pankajkoti: I guess I will put it log level ERROR instead of EXCEPTION since we don't need the stacktrace here

astronomer/providers/apache/hive/example_dags/example_hive.py
pankajastro: what about logging the actual error messages? it should be in the error object.

rajaths010494 (approved):

bharanidharan14 (approved):

astronomer/providers/apache/hive/example_dags/example_hive.py
pankajkoti: Addressed. Included in the log.

Issue #279, opened 2022-04-26
Fix Redshift example DAG


pankajkoti:
1.Replace `create_redshift_cluster` in the example DAG with the CreateRedshiftClusterOperator

2.I see the `delete_redshift_cluster_snapshot` task of the DAG has always raised an Exception on a daily basis but we're catching that in the code and as a result the task is getting marked success. Below is the exception,

```
[2022-04-25, 15:07:59 IST] {taskinstance.py:1270} INFO - Executing <Task(PythonOperator): delete_redshift_cluster_snapshot> on 2022-04-25 09:32:12.843775+00:00
[2022-04-25, 15:07:59 IST] {standard_task_runner.py:52} INFO - Started process 71 to run task
[2022-04-25, 15:07:59 IST] {standard_task_runner.py:79} INFO - Running: ['airflow', 'tasks', 'run', 'example_async_redshift_cluster_management', 'delete_redshift_cluster_snapshot', 'manual__2022-04-25T09:32:12.843775+00:00', '--job-id', '7687', '--raw', '--subdir', 'DAGS_FOLDER/example_redshift_cluster_management.py', '--cfg-path', '/tmp/tmpxk89w10_', '--error-file', '/tmp/tmpxopbepnn']
[2022-04-25, 15:07:59 IST] {standard_task_runner.py:80} INFO - Job 7687: Subtask delete_redshift_cluster_snapshot
[2022-04-25, 15:07:59 IST] {logging_mixin.py:109} INFO - Running <TaskInstance: example_async_redshift_cluster_management.delete_redshift_cluster_snapshot manual__2022-04-25T09:32:12.843775+00:00 [running]> on host cometary-asterism-3668-worker-c49bfd594-pfmbv
[2022-04-25, 15:07:59 IST] {taskinstance.py:1446} INFO - Exporting the following env vars:
AIRFLOW_CTX_DAG_OWNER=airflow
AIRFLOW_CTX_DAG_ID=example_async_redshift_cluster_management
AIRFLOW_CTX_TASK_ID=delete_redshift_cluster_snapshot
AIRFLOW_CTX_EXECUTION_DATE=2022-04-25T09:32:12.843775+00:00
AIRFLOW_CTX_DAG_RUN_ID=manual__2022-04-25T09:32:12.843775+00:00
[2022-04-25, 15:07:59 IST] {credentials.py:1111} INFO - Found credentials in environment variables.
[...]

PR #348, opened 2022-05-13
Fix Redshift DAGs to catch appropriate exceptions

pankajkoti:
Cluster delete and snapshot delete tasks keep on waiting while
the status is in `deleting` state. Since, the status get is part
of a while loop, the last iteration when they get deleted raises
an exception that the corresponding resource is not found and it
marks the task as failed. We are fixing this by catching the relevant
status code and other exceptions are re-raised.

Additionally, it is observed quite often that the DAG tasks fail
without any logs. @ephraimbuddy suggested that this could be due
to DAG processing timeouts occurring due to time spent in importing
heavy libraries (although we could not find the dag processing logs in Astro cloud). 
So taking this guess & the reference of
https://airflow.apache.org/docs/apache-airflow/stable/best-practices.html#top-level-python-code
we're delaying the import of `boto` to the task execution stage and
avoiding to import it at top module level.

We're also also renaming the operator's python reference variable
names to be consistent with the `task_id`.

Closes: #279

### DIFF
--- a/astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
@@ -82,6 +82,7 @@ def create_redshift_cluster_callable() -> None:
     )
 
     while get_cluster_status() != "available":
+        logging.info("Waiting for cluster to be available. Sleeping for 30 seconds.")
         time.sleep(30)
 
 
@@ -99,11 +100,13 @@ def create_redshift_cluster_snapshot_callable() -> None:
     )
 
     while get_snapshot_status() != "available":
+        logging.info("Waiting for cluster snapshot to be available. Sleeping for 30 seconds.")
         time.sleep(30)
 
     # Introduce sleep to wait for the cluster to be available back from 'modifying' state before proceeding on to the
     # downstream task.
     while get_cluster_status() != "available":
+        logging.info("Waiting for cluster to be available. Sleeping for 30 seconds.")
         time.sleep(30)
 
 
@@ -121,6 +124,7 @@ def delete_redshift_cluster_snapshot_callable() -> None:
         )
 
         while get_snapshot_status() == "deleting":
+            logging.info("Waiting for cluster snapshot to be deleted. Sleeping for 30 seconds.")
             time.sleep(30)
 
     except ClientError as exception:
@@ -136,6 +140,7 @@ def delete_redshift_cluster_snapshot_callable() -> None:
     # Introduce sleep to wait for the cluster to be available back after snapshot activity before proceeding on to the
     # downstream task.
     while get_cluster_status() != "available":
+        logging.info("Waiting for cluster to be available. Sleeping for 30 seconds.")
         time.sleep(30)
 
 
@@ -153,6 +158,7 @@ def delete_redshift_cluster_callable() -> None:
         )
 
         while get_cluster_status() == "deleting":
+            logging.info("Waiting for cluster to be deleted. Sleeping for 30 seconds.")
             time.sleep(30)
 
     except ClientError as exception:
--- a/astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
@@ -3,12 +3,10 @@ import os
 import time
 from datetime import datetime, timedelta
 
-import boto3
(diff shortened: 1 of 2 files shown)


astronomer/providers/amazon/aws/example_dags/example_redshift_sql.py
```diff
@@ -54,17 +57,24 @@ def delete_redshift_cluster() -> None:
...
-        logging.exception("Error deleting redshift cluster")
-        raise exception
+        if exception.response.get("Error", {}).get("Code", "") == "ClusterNotFound":
+            logging.error(
+                "Cluster might have already been deleted. Error message is: %s",
+                exception.response["Error"]["Message"],
+            )
+        else:
+            logging.exception("Error deleting redshift cluster")
+            raise
```
pankajastro, line 71: what about raising `raise exception` this might have more info?

astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
```diff
@@ -94,20 +98,20 @@ def create_redshift_cluster_snapshot() -> None:
         ],
     )
 
-    while True:
-        if get_snapshot_status() == "available":
-            break
+    while get_snapshot_status() != "available":
         time.sleep(30)
```
kaxil, line 102: probably worth adding a log line, to say we are sleeping, otherwise it will look like if the job is hanging

astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
```diff
@@ -116,22 +120,30 @@ def delete_redshift_cluster_snapshot() -> None:
             SnapshotClusterIdentifier=REDSHIFT_CLUSTER_IDENTIFIER,
         )
 
-        while True:
-            if get_snapshot_status() == "deleting":
-                time.sleep(30)
-                continue
+        while get_snapshot_status() == "deleting":
+            time.sleep(30)
```
kaxil, line 124: Same for all the sleep's

astronomer/providers/amazon/aws/example_dags/example_redshift_sql.py
```diff
@@ -54,17 +57,24 @@ def delete_redshift_cluster() -> None:
...
-            if get_cluster_status() == "deleting":
-                time.sleep(30)
-                continue
+        while get_cluster_status() == "deleting":
+            time.sleep(30)
+
     except ClientError as exception:
-        logging.exception("Error deleting redshift cluster")
-        raise exception
+        if exception.response.get("Error", {}).get("Code", "") == "ClusterNotFound":
```
kaxil, line 64: Should we add a check on L55 to see if the cluster exists or not? try..except is good too :) but just thinking out loud

astronomer/providers/amazon/aws/example_dags/example_redshift_sql.py
kaxil: `logging.exception` will contain that details

pankajkoti: in this case, `raise` will have the same impact as `raise exception`; it's just a short hand for that.

Screenshot 2022-05-16 at 12 50 25 PM

astronomer/providers/amazon/aws/example_dags/example_redshift_sql.py
pankajkoti: yes, like the idea of being proactively checking rather than reacting with a try-catch. We have the except more for the while loop on L60 where the cluster remains in deleting state for a while and then throws this error. In my opinion it makes sense for L55 to have that check as it will be only once; but for L60 it will then make 2 API calls each time in the loop. 
However, I do not seem to find a relevant method available to check beforehand that the cluster exists https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/redshift.html

astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
pankajkoti: yes, makes sense, added everywhere a log for the same.

bharanidharan14 (commented):
Overall looks good to me

pankajastro:
LGTM

pankajastro (approved):

phanikumv (approved):

PR #349, opened 2022-05-13
Enhance command to be run inside KubernetesPodOperator

pankajkoti:
The previous command had just a plain sleep of 30 seconds.
Enhance it to write some log every one second and go back to sleep.


bharanidharan14 (approved):

pankajastro:
How long does it take to complete? if more than 1 minute then I feel 1 sec is too frequent.

pankajkoti:
> How long does it take to complete? if more than 1 minute then I feel 1 sec is too frequent.

It takes about 33 seconds

Issue #351, opened 2022-05-16
Tasks failing without writing any logs


pankajkoti:
Some tasks are failing without writing any logs.

Possible reason might be 
Imports of heavy libraries at module level might cause DAG processing timeout errors.
To avoid this, we need to delay of imports to the specific operator scope.

Reference: https://airflow.apache.org/docs/apache-airflow/stable/best-practices.html#top-level-python-code

However we need to investigate in detail.

pankajkoti:
We have a Zendesk ticket created for this issue: https://astronomer.zendesk.com/agent/tickets/9862

bharanidharan14:
Since issue is not being observed for weeks, so closing the issue

PR #354, opened 2022-05-16
Delay import of libraries to task scope

pankajkoti:
DAGs processing might timeout due to import of heavy libraries
at top module level. Since, the DAGs are refreshed often, we want
to avoid such timeouts in processing. Hence, delay the import of
heavy libraries to task definition scope.
Reference: https://airflow.apache.org/docs/apache-airflow/stable/best-practices.html#top-level-python-code

part of: #351


phanikumv (commented):
@pankajkoti Appreciate the detailed descriptions that you are putting in along with commit messages. 💯

Issue #294, opened 2022-04-29
[POC] Improve unit test quality


pankajastro:
Investigate how we can make the unit tests more readable and reusable.

- class-based test (This will help to group tests together in a better way)
- util module inside test which will have all shared function/method
- move third party api mocked response to util so that it can be shared between tests (example: https://github.com/astronomer/astronomer-providers/blob/main/tests/amazon/aws/hooks/test_emr.py#L324)
- better meaningful docs string 
- move all the test constants to a separate class of util module

PR #368, opened 2022-05-20
Unit Test Cleanup

pankajastro:
This is an attempt to improve the unit test case this PR will add the below value
- We have to not write context` fixture in each test file 
- We can move mocked third-party API response that we want to test out of test case files
- We can reuse the `create_context` method which is currently redundant at dozen of place 
- We can move the constant like conn_name, poll_interval and other repeated param out of the test file and reuse it
- test cases name will be smaller that would be easy to read
- It would be easy to read the case since now it will have a nice grouping which mapped to the actual class
- It would be easy to add a new test case and improve/change the existing test case since it would be grouped in a class 
- Running of test would be easier using pytest I can run a group of tests as well as individual test

Since this PR have only changes for core providers so above points are currently  applicable on  core provider only

Closes: #294


pankajastro:
> Need to test parametrize in class based approach

This should be a problem, I have checked in the apache repo

kaxil:
What's the status of this PR?

bharanidharan14:
> > Need to test parametrize in class based approach
> 
> This should be a problem, I have checked in the apache repo

We can't have parametrize based function ?

pankajastro:
> > > Need to test parametrize in class based approach
> > 
> > 
> > This should be a problem, I have checked in the apache repo
> 
> We can't have parametrize based function ?

It is doable, it's a typo in the last comment, updated it. https://github.com/apache/airflow/blob/main/tests/providers/amazon/aws/hooks/test_base_aws.py#L704

pankajastro:
> What's the status of this PR?

@kaxil This was one of my suggestions to improve unit test reusability and readability during our retro call. I came up with some improvements for one module `core`.

bharanidharan14 (commented):
Overall LGTM, only thing can you add more description what is done on the cleanup activity, so that it can be avoided in the upcoming PR

bharanidharan14:
@pankajastro can you please fix the Description

bharanidharan14 (approved):

bharanidharan14:
@pankajastro can we merge this ? Is anything need to be done ?

pankajastro:
I want to merge this but before going ahead I would like to know what other people think about these changes. WDYT @phanikumv @kaxil @pankajkoti @bharanidharan14 @rajaths010494 ?

kaxil (approved):
I would like to hear opinions from others :)

pankajkoti (commented):
The test coverage is dropping by 0.95% . Do we need to worry about it?

tests/core/sensors/test_external_task.py
```diff
@@ -11,147 +11,122 @@
...
-    assert len(execution_dates) == 2
-    sensor.execution_date_fn = None  # Reset this field for next test
-
-    # Case #3 both sensor.execution_delta and sensor.execution_date_fn are set to None
-    execution_dates = sensor.get_execution_dates(context)
-    assert execution_dates[0] == datetime(2015, 1, 1)
-    assert len(execution_dates) == 1
+
+class TestExternalTaskSensorAsync:
+    TASK_ID = "external_task_sensor_check"
```
bharanidharan14, line 16: can we add this variables inside setup_class like class variables?

pankajastro:
> The test coverage is dropping by 0.95% . Do we need to worry about it?

If you see the delta changes https://github.com/astronomer/astronomer-providers/pull/368#issuecomment-1132732770 the coverage has dropped for most of the files which are not related to this PR so I don't think it matters much matter we still have all tests old test here

tests/core/sensors/test_external_task.py
pankajastro: do you mean I should create a setup class and keep this variable in that class? Can you please elaborate a little bit?

bharanidharan14: Yes you do like this 

```
class Test:
    def setup_class(self) -> None:
        self.xxxx: str = ""
        self.yyyyy: str = ''
        self.connection_string = (
        )

```

bharanidharan14: or you can create def test_init() inside the class

pankajastro: these variables are just single lines and getting used in this class only so moving outside to a different class not sure how much value will add.  regarding variables like conn which might get used in multiple test classes, I'm planning to keep them in `util`

pankajastro:
Created a follow-up ticket https://github.com/astronomer/astronomer-providers/issues/568

Issue #369, opened 2022-05-20
ExternalTaskSensorAsync Example should defer  


pankajastro:
Currently, it waits on the dummy operator so won't get a chance to defer

Related to: https://github.com/astronomer/astronomer-providers/pull/361

PR #378, opened 2022-05-23
Remove test_external_task_async_waits_for_me from master DAG

pankajastro:
Removed `test_external_task_async_waits_for_me_dag_from_master_dag` from `master_dag` since this dag does not test any of our async implantations but the purpose of this was to run `ExternalTaskSensorAsync` example DAG so no need to trigger this via master DAG. 

Now, [astronomer/providers/core/example_dags/example_external_task.py](https://github.com/astronomer/astronomer-providers/pull/370/files#diff-b9ba255046aec454155394e74f499084ba8ed2683b77f7316331b201da786656) trigger this instead of master DAG

related to: https://github.com/astronomer/astronomer-providers/issues/369

### DIFF
--- a/.circleci/integration-tests/master_dag.py
+++ b/.circleci/integration-tests/master_dag.py
@@ -122,7 +122,6 @@ with DAG(
 
     # Core DAG
     core_task_info = [
-        {"external_task_wait_dag": "test_external_task_async_waits_for_me"},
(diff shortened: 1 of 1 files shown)


phanikumv (approved):

phanikumv:
@pankajastro can you take a look at why the external task integration DAG is not starting?

PR #382, opened 2022-05-25
Revert Openlineage Bigquery Extractor

kaxil:
We are reverting this change to release 1.4.0 but they will be available in 1.5.0


pankajkoti (approved):
Should we also remove the OpenLineage specific environment variables from docker-compose.yaml? Otherwise, looks good to me.

dev/docker-compose.yaml
```diff
@@ -22,8 +22,6 @@ x-airflow-common:
     OPENLINEAGE_URL: http://host.docker.internal:5050/
     OPENLINEAGE_API_KEY: <YOUR_OPENLINEAGE_API_KEY>
     OPENLINEAGE_NAMESPACE: dev
```
pankajkoti, line 24:
```suggestion
```

PR #398, opened 2022-05-30
Add `retries` arg in the default params for example DAGs

pankajkoti:
Adds `retries` argument along with `retry_delay` to reattempt
failed tasks in example DAGs.

### DIFF
--- a/astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
@@ -13,6 +13,8 @@ EXECUTION_TIMEOUT = int(os.getenv("EXECUTION_TIMEOUT", 6))
 
 default_args = {
     "execution_timeout": timedelta(hours=EXECUTION_TIMEOUT),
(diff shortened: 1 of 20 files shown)


astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
```diff
@@ -13,6 +13,8 @@
 
 default_args = {
     "execution_timeout": timedelta(hours=EXECUTION_TIMEOUT),
+    "retries": int(os.getenv("DEFAULT_TASK_RETRIES", 2)),
```
pankajastro, line 16: Does it start from the failed task or from the first task when the retry runs?

pankajastro: No, retry param you are setting at DAG level so let's say I have three tasks in DAG (A -> B -> C) Now B failed for some reason then retry will resume re-run from task B or task A?

phanikumv: retry should be set at task level only

pankajastro (approved):

Issue #366, opened 2022-05-20
Microsoft ADF task failing with timeout error


pankajastro:
AzureDataFactoryRunPipelineOperatorAsync failing with timeout error

```
[2022-05-20, 08:17:03 UTC] {task_command.py:369} INFO - Running <TaskInstance: example_async_adf_run_pipeline.run_pipeline_wait manual__2022-05-20T08:16:43.588216+00:00 [running]> on host 10.1.4.43
[2022-05-20, 08:17:04 UTC] {taskinstance.py:1888} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/astronomer/providers/microsoft/azure/operators/data_factory.py", line 89, in execute_complete
    raise AirflowException(event["message"])
airflow.exceptions.AirflowException: Timeout: The pipeline run 389780b0-d815-11ec-9848-4eb583b4c754 has InProgress.
```

phanikumv:
Change has been deployed and today's execution on astro-cloud went through fine, will observe for this week and then close it in case of no more issues.

pankajkoti:
The re-ordering of tasks seem to have helped and the DAG has run fine for the last 3 days post deployment of this change and hence closing the ticket.

phanikumv:
Closing as the latest issue is tracked in #461

PR #405, opened 2022-05-31
Reorder sequence for execution of tasks

pankajkoti:
I am unable to reproduce issue #366 locally after trying multiple
times. Looking at the code, I am understanding that the
`run_pipeline_no_wait` and `run_pipeline_wait` try to execute the
same pipeline with same name and resources. The `run_pipeline_no_wait`
tasks starts the pipeline and does not wait for the pipeline to finish
before marking the task as success. My hunch here is that since
the pipeline launched by `run_pipeline_no_wait` is still in progress,
another immediate attempt by `run_pipeline_wait` task to start the
same pipeline is the reason for the failure of the task observed in
issue #366. Hence, I am reordering the task sequence so that
`run_pipeline_wait` task runs first; the task waits for the pipeline
to finish before marking the task as success and then we launch
the `run_pipeline_no_wait` task as the downstream task.

I would like to push this commit &  merge to main so that we can
observe in our Astro Cloud deployment's master DAG run on whether this
solves the problem. If the problem still persists, we will then need
to spend additional efforts in investigating the issue.

### DIFF
--- a/astronomer/providers/microsoft/azure/example_dags/example_adf_run_pipeline.py
+++ b/astronomer/providers/microsoft/azure/example_dags/example_adf_run_pipeline.py
@@ -185,17 +185,17 @@ with DAG(
     # [END howto_create_resource_group]
 
     # [START howto_operator_adf_run_pipeline]
-    run_pipeline_no_wait = AzureDataFactoryRunPipelineOperatorAsync(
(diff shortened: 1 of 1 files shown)


phanikumv (changes requested):
Please include the example DAG name in the title

phanikumv (approved):

Issue #406, opened 2022-05-31
Dataproc : remove location in favour of region


pankajastro:
OSS PR
- https://github.com/apache/airflow/pull/23250
- https://github.com/apache/airflow/pull/23350

phanikumv:
To do more analysis in order to check whether this change can be included in the upcoming release.

cc @kaxil

PR #407, opened 2022-05-31
Remove location from ``DataprocSubmitJobOperatorAsync`` docstring

pankajastro:
Remove `location` from `DataprocSubmitJobOperatorAsync` since it's been removed from the parent class i.e `DataprocSubmitJobOperator`

Closes: #406

### DIFF
--- a/astronomer/providers/google/cloud/operators/dataproc.py
+++ b/astronomer/providers/google/cloud/operators/dataproc.py
@@ -15,7 +15,6 @@ class DataprocSubmitJobOperatorAsync(DataprocSubmitJobOperator):
 
     :param project_id: Optional. The ID of the Google Cloud project that the job belongs to.
     :param region: Required. The Cloud Dataproc region in which to handle the request.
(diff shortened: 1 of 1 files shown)


pankajkoti (commented):
Since the parameter is no longer present, it would mean that this is backward-incompatible and would we need to include this in a major release? 
Reference: https://docs.astronomer.io/astro/runtime-version-lifecycle-policy#versioning-scheme

pankajastro:
> Since the parameter is no longer present, it would mean that this is backward-incompatible. Would we need to include this in a major release? Reference: https://docs.astronomer.io/astro/runtime-version-lifecycle-policy#versioning-scheme

Agree, Ideally we should follow this but in this particular scenario backwards compatibility does not depend on this reop version rather it depends on https://pypi.org/project/apache-airflow-providers-google/.  Here, we only have docs change but I'm ok to park it for major release too.

kaxil (approved):
Since this is just a doc change I am fine with it.

PR #447, opened 2022-06-15
Add async implementation for Redshift cluster deletion

pankajkoti:
The PR addresses the following:
- Implements necessary operator, trigger, hook changes to allow  async deletion of a Redshift cluster.
- Uses this new operator `RedshiftPauseClusterOperatorAsync` in the example DAG.
- Fixes existing tests that fail due to method signature change.
- Add a make target to build images with `aws-cli` installed required for local testing of AWS example DAGs

closes: #403

### DIFF
--- a/astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
@@ -17,17 +17,21 @@ from astronomer.providers.amazon.aws.sensors.redshift_cluster import (
     RedshiftClusterSensorAsync,
 )
 
+AWS_CONN_ID = os.getenv("ASTRO_AWS_CONN_ID", "aws_default")
+AWS_DEFAULT_REGION = os.getenv("AWS_DEFAULT_REGION", "us-east-2")
+AWS_ACCESS_KEY_ID = os.getenv("AWS_ACCESS_KEY_ID", "**********")
+AWS_SECRET_ACCESS_KEY = os.getenv("AWS_SECRET_ACCESS_KEY", "***********")
+EXECUTION_TIMEOUT = int(os.getenv("EXECUTION_TIMEOUT", 6))
+REDSHIFT_CLUSTER_DB_NAME = os.getenv("REDSHIFT_CLUSTER_DB_NAME", "astro_dev")
 REDSHIFT_CLUSTER_IDENTIFIER = os.getenv("REDSHIFT_CLUSTER_IDENTIFIER", "astro-providers-cluster")
 REDSHIFT_CLUSTER_MASTER_USER = os.getenv("REDSHIFT_CLUSTER_MASTER_USER", "awsuser")
 REDSHIFT_CLUSTER_MASTER_PASSWORD = os.getenv("REDSHIFT_CLUSTER_MASTER_PASSWORD", "********")
-REDSHIFT_CLUSTER_DB_NAME = os.getenv("REDSHIFT_CLUSTER_DB_NAME", "astro_dev")
-REDSHIFT_CLUSTER_TYPE = os.getenv("REDSHIFT_CLUSTER_TYPE", "single-node")
 REDSHIFT_CLUSTER_NODE_TYPE = os.getenv("REDSHIFT_CLUSTER_NODE_TYPE", "dc2.large")
-AWS_ACCESS_KEY_ID = os.getenv("AWS_ACCESS_KEY_ID", "**********")
-AWS_SECRET_ACCESS_KEY = os.getenv("AWS_SECRET_ACCESS_KEY", "***********")
-AWS_DEFAULT_REGION = os.getenv("AWS_DEFAULT_REGION", "us-east-2")
-AWS_CONN_ID = os.getenv("ASTRO_AWS_CONN_ID", "aws_default")
-EXECUTION_TIMEOUT = int(os.getenv("EXECUTION_TIMEOUT", 6))
+REDSHIFT_CLUSTER_STATUS_FETCH_INTERVAL_SECONDS = int(
+    os.getenv("REDSHIFT_CLUSTER_STATUS_FETCH_INTERVAL_SECONDS", 10)
+)
+REDSHIFT_CLUSTER_TYPE = os.getenv("REDSHIFT_CLUSTER_TYPE", "single-node")
+
 
 default_args = {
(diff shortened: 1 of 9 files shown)


astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
```diff
@@ -232,10 +233,13 @@ def delete_redshift_cluster_callable() -> None:
         trigger_rule="all_done",
     )
```
bharanidharan14, line 235: can you add the doc tag to this operator

bharanidharan14 (commented):
Fix code coverage and test case for the delete operator

astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
```diff
@@ -232,11 +233,16 @@ def delete_redshift_cluster_callable() -> None:
...
 
-    delete_redshift_cluster = PythonOperator(
+    # [START howto_operator_redshift_delete_cluster_async]
+    delete_redshift_cluster = RedshiftDeleteClusterOperatorAsync(
         task_id="delete_redshift_cluster",
-        python_callable=delete_redshift_cluster_callable,
```
pankajastro, line 237: If not other places we are using `delete_redshift_cluster_callable` please remove it.

astronomer/providers/amazon/aws/hooks/redshift_cluster.py
```diff
@@ -29,8 +30,47 @@ async def cluster_status(self, cluster_identifier: str) -> Dict[str, Any]:
...
+                    FinalClusterSnapshotIdentifier=final_cluster_snapshot_identifier,
+                )
+                status = response["Cluster"]["ClusterStatus"] if response and response["Cluster"] else None
+                if status == "deleting":
+                    flag = asyncio.Event()
+                    while True:
+                        expected_response = await asyncio.create_task(
+                            self.get_cluster_status(cluster_identifier, "cluster_not_found", flag, True)
+                        )
+                        await asyncio.sleep(10)
```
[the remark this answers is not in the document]
pankajkoti: Getting from operator now.

astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
pankajkoti: Removed.

astronomer/providers/amazon/aws/operators/redshift_cluster.py
```diff
@@ -15,6 +16,80 @@
...
+    """
+
+    def __init__(
+        self,
+        *,
+        skip_final_cluster_snapshot: bool = True,
+        final_cluster_snapshot_identifier: Optional[str] = None,
+        cluster_status_fetch_interval_seconds: int = 10,
+        aws_conn_id: str = "aws_default",
+        poll_interval: float = 5.0,
```
pankajastro, line 37: what is the difference between cluster_status_fetch_interval_seconds and poll_interval? I think we usually use poll_interval to fetch the status.

astronomer/providers/amazon/aws/operators/redshift_cluster.py
```diff
@@ -15,6 +16,80 @@
...
+                    cluster_identifier=self.cluster_identifier,
+                    operation_type="delete_cluster",
+                    skip_final_cluster_snapshot=self.skip_final_cluster_snapshot,
+                    final_cluster_snapshot_identifier=self.final_cluster_snapshot_identifier,
+                    cluster_status_fetch_interval_seconds=self.cluster_status_fetch_interval_seconds,
+                ),
+                method_name="execute_complete",
+            )
+        else:
+            self.log.warning(
```
pankajastro, line 70: just thinking either we should log a warning or raise an error here

astronomer/providers/amazon/aws/operators/redshift_cluster.py
```diff
@@ -15,6 +16,80 @@
...
+            if "status" in event and event["status"] == "error":
+                msg = "{0}: {1}".format(event["status"], event["message"])
+                raise AirflowException(msg)
+            elif "status" in event and event["status"] == "success":
+                self.log.info("%s completed successfully.", self.task_id)
+                self.log.info("Deleted cluster successfully")
+                return None
+        else:
+            self.log.info("%s completed successfully.", self.task_id)
+            return None
```
pankajastro, line 90: this return stmt should not require

astronomer/providers/amazon/aws/operators/redshift_cluster.py
```diff
@@ -15,6 +16,80 @@
...
+        successful.
+        """
+        if event:
+            if "status" in event and event["status"] == "error":
+                msg = "{0}: {1}".format(event["status"], event["message"])
+                raise AirflowException(msg)
+            elif "status" in event and event["status"] == "success":
+                self.log.info("%s completed successfully.", self.task_id)
+                self.log.info("Deleted cluster successfully")
+                return None
```
pankajastro, line 87: you can remove this return stmt

astronomer/providers/amazon/aws/operators/redshift_cluster.py
```diff
@@ -15,6 +16,80 @@
...
+        else:
```
pankajastro, line 88: if you won't receive any event from the trigger that should be an error case, what do you think?

astronomer/providers/amazon/aws/operators/redshift_cluster.py
pankajkoti: using poll_interval now.

astronomer/providers/amazon/aws/operators/redshift_cluster.py
pankajkoti: It may happen that the task failed after deleting the cluster and it is retired. At that time we would get cluster_not_found and we would not want to task to fail then. Made a change to check the status; logging a warning if cluster is not found and otherwise, raising an exception like you suggested.

astronomer/providers/amazon/aws/operators/redshift_cluster.py
pankajkoti: Raising an exception now.

pankajastro:
LGTM

pankajastro (approved):

PR #456, opened 2022-06-20
Upgrade airflow and astro cloud image

phanikumv:
Upgrade the integration tests to use `astro-runtime:5.0.4-base` and  `quay.io/astronomer/ap-airflow:2.3.2`


pankajastro (approved):

dev/Dockerfile
```diff
@@ -1,4 +1,4 @@
-ARG IMAGE_NAME="quay.io/astronomer/astro-runtime:5.0.2-base"
+ARG IMAGE_NAME="quay.io/astronomer/astro-runtime:5.0.4-base"
```
rajaths010494, line 1: Can we also upgrade in Dockerfile.aws and Dockerfile.emr_eks_container?

PR #464, opened 2022-06-22
Fix: Keep the Python Filename Consistent 

pankajastro:
closes: #466 

Renaming the files to maintain consistency
- removed async, extract, operator, trigger words from python filename

This change will not affect any customer interfaces since BigQueryAsyncExtractor and SnowflakeAsyncExtractor we have not released yet


tests/cncf/kubernetes/hooks/test_kubernetes.py
```diff
@@ -49,7 +47,7 @@ async def test_kubernetes__load_config(
 
 
 @pytest.mark.asyncio
-@mock.patch("astronomer.providers.cncf.kubernetes.hooks.kubernetes_async.KubernetesHookAsync._load_config")
+@mock.patch("astronomer.providers.cncf.kubernetes.hooks.kubernetes.KubernetesHookAsync._load_config")
```
phanikumv, line 50: why do we need to do this? What if a customer uses the old import path?

pankajastro: I believe customers will be using only the operator/sensor/extractor, not the hook. WDYT?

tests/google/cloud/triggers/test_kubernetes_engine.py
```diff
@@ -97,9 +95,7 @@ async def test_run(mock_tmp, get_api_client_async, wait_for_pod_start):
 @mock.patch(
     "astronomer.providers.cncf.kubernetes.triggers.wait_container.WaitContainerTrigger.wait_for_pod_start"
 )
-@mock.patch(
-    "astronomer.providers.cncf.kubernetes.hooks.kubernetes_async.KubernetesHookAsync.get_api_client_async"
-)
+@mock.patch("astronomer.providers.cncf.kubernetes.hooks.kubernetes.KubernetesHookAsync.get_api_client_async")
```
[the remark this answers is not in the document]
pankajastro: I agree. but the sync version also uses same pattern https://github.com/apache/airflow/blob/main/airflow/providers/cncf/kubernetes/hooks/kubernetes.py generally we keep the filename same as sync version and also I'm not able to think the better name please let me know if I can improve it.

pankajastro:
> Did you test the example DAG's?

No, I'll test it once the people will agree with this change.

tests/cncf/kubernetes/hooks/test_kubernetes.py
phanikumv: - Assume one of the PROD DAG of a customer has 

`import astronomer.providers.cncf.kubernetes.hooks.kubernetes_async.KubernetesHookAsync`
wouldnt their DAG break? 

- There is evidence of some customers using hook directly, one of our customers  uses TrinoHook directly. I don't know whether there is a possibility of other users doing the same. But it's always better to be on safe side.

- We'll need to redirect the user to use the latest path if we still want to take the risk..something like below where we warn the users to not use deprecated paths

` warnings.warn(
                "The import path parameter <> has been deprecated. Use the import path <> instead.",
                DeprecationWarning,
                stacklevel=2,
 )`

phanikumv: cc @kaxil to comment on this

kaxil: yeah changes to hook is considered a breaking change. Add a backwards-compatible path with deprecation warning for old path

pankajastro: thank you @phanikumv @kaxil for pointing it out, I'll change it accordingly.

pankajastro: I have tried to deprecate it in c2d42a78f686c4342c9d48cfb8205acc90cbd85a commit. Please let me know how it is looking

astronomer/providers/amazon/aws/hooks/base_aws_async.py
```diff
@@ -1,73 +1,9 @@
...
-            aws_access_key_id=aws_access_key_id,
-            aws_session_token=aws_session_token,
-            verify=self.verify,
-            config=self.config,
-        )
+warnings.warn(
+    "This module is deprecated. Please use `astronomer.providers.amazon.aws.hooks.base_aws`.",
+    DeprecationWarning,
+    stacklevel=2,
+)
```
phanikumv, line 9: Perfect ! Can you test once by using the deprecated path and let me know the result

tests/google/cloud/triggers/test_kubernetes_engine.py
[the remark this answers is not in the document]
phanikumv: if we are in sync with apache airflow, its fine

tests/cncf/kubernetes/hooks/test_kubernetes.py
phanikumv: Perfect ! Can you test once by using the deprecated path and let me know the result

astronomer/providers/amazon/aws/hooks/base_aws.py
```diff
@@ -0,0 +1,73 @@
...
+        roles provided in the Airflow connection extra args via aiobotocore is not supported by the
+        library yet.
+
+    :param aws_conn_id: The Airflow connection used for AWS credentials.
+        If this is None or empty then the default boto3 behaviour is used. If
+        running Airflow in a distributed manner and aws_conn_id is None or
+        empty, then default boto3 configuration would be used (and must be
+        maintained on each worker node).
+    :param verify: Whether or not to verify SSL certificates.
+        https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html
```
kaxil, line 23: You can use the seealso Sphinx directive for this link - https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-seealso

astronomer/providers/amazon/aws/hooks/base_aws.py
```diff
@@ -0,0 +1,73 @@
...
+    :param region_name: AWS region_name. If not specified then the default boto3 behaviour is used.
+    :param client_type: boto3.client client_type. Eg 's3', 'emr' etc
+    :param resource_type: boto3.resource resource_type. Eg 'dynamodb' etc
+    :param config: Configuration for botocore client.
+        (https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html)
```
kaxil, line 28: same here - https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-seealso

kaxil (approved):

astronomer/providers/amazon/aws/hooks/base_aws.py
bharanidharan14, line 28: .. seealso::
        For more information on Session reference, take a look at the link:
:ref:`https://boto3.amazonaws.com/v1/documentation/api/latest/reference/core/session.html`

astronomer/providers/amazon/aws/hooks/base_aws.py
pankajastro: Disable flake8 `RS303` check since it was giving false positive and moved the documentation ref at end of param since multiple param pointing to the same link

astronomer/providers/amazon/aws/hooks/base_aws_async.py
pankajastro: Fix the code coverage for Kubernetes but for aws having some difficulty to mock `sync_to_async` so ignoring it for time being

pankajastro:
Look like we will have to merge this https://github.com/astronomer/astronomer-providers/pull/471 @rajaths010494 where we are on this?

astronomer/providers/amazon/aws/hooks/base_aws.py
pankajastro: Docs screenshot

Screenshot 2022-06-24 at 11 21 03 PM

rajaths010494 (approved):

bharanidharan14 (approved):

phanikumv (approved):

PR #471, opened 2022-06-24
Fix BigQueryInsertJobOperatorAsync failure after the google provider upgrade

### LINKED ISSUE #470: Fix BigQueryInsertJobOperatorAsync failure
BigQueryInsertJobOperatorAsync is failing with below error. Might be due to a breaking change on google provider in OSS

AttributeError: 'BigQueryInsertJobOperatorAsync' object has no attribute '_job_id'

rajaths010494:
- Sync version has changed and removed _job_id in operator which was used in our repo,
- Code has been moved to hooks with certain changes.
- Implement the code change needed for it in our repo

closes #470

### DIFF
--- a/astronomer/providers/google/cloud/operators/bigquery.py
+++ b/astronomer/providers/google/cloud/operators/bigquery.py
@@ -86,7 +86,14 @@ class BigQueryInsertJobOperatorAsync(BigQueryInsertJobOperator, BaseOperator):
         hook = _BigQueryHook(gcp_conn_id=self.gcp_conn_id)
 
         self.hook = hook
-        job_id = self._job_id(context)
+        job_id = self.hook.generate_job_id(
+            job_id=self.job_id,
+            dag_id=self.dag_id,
+            task_id=self.task_id,
+            logical_date=context["logical_date"],
+            configuration=self.configuration,
+            force_rerun=self.force_rerun,
+        )
 
         try:
             job = self._submit_job(hook, job_id)
--- a/astronomer/providers/google/cloud/operators/bigquery.py
+++ b/astronomer/providers/google/cloud/operators/bigquery.py
@@ -86,7 +86,7 @@ class BigQueryInsertJobOperatorAsync(BigQueryInsertJobOperator, BaseOperator):
         hook = _BigQueryHook(gcp_conn_id=self.gcp_conn_id)
 
         self.hook = hook
-        job_id = self.hook.generate_job_id(  # type: ignore[attr-defined]
+        job_id = self.hook.generate_job_id(
             job_id=self.job_id,
             dag_id=self.dag_id,
             task_id=self.task_id,
--- a/astronomer/providers/google/cloud/operators/bigquery.py
+++ b/astronomer/providers/google/cloud/operators/bigquery.py
@@ -93,7 +93,7 @@ class BigQueryInsertJobOperatorAsync(BigQueryInsertJobOperator, BaseOperator):
             logical_date=context["logical_date"],
             configuration=self.configuration,
             force_rerun=self.force_rerun,
-        )
+        )  # type: ignore[attr-defined]
 
         try:
             job = self._submit_job(hook, job_id)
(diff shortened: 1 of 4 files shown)


pankajastro (approved):

phanikumv (approved):

phanikumv (commented):
Let's test in astro cloud after merging this.

pankajkoti:
> Let's test in astro cloud after merging this.

We should also be able to test this locally with the latest version of the provider.

astronomer/providers/google/cloud/operators/bigquery.py
```diff
@@ -86,7 +86,14 @@ def execute(self, context: Context) -> None:  # noqa: D102
         hook = _BigQueryHook(gcp_conn_id=self.gcp_conn_id)
 
         self.hook = hook
-        job_id = self._job_id(context)
+        job_id = self.hook.generate_job_id(  # type: ignore[attr-defined]
```
kaxil, line 89: Will this work with older version of the google provider?

kaxil: Does the older version of the provider has `generate_job_id` in the Hook?

tests/google/cloud/operators/test_bigquery.py
```diff
@@ -142,37 +143,12 @@ def test_bigquery_insert_job_operator_execute_complete():
     mock_log_info.assert_called_with("%s completed with response %s ", "insert_query_job", "Job completed")
 
 
-@mock.patch("airflow.providers.google.cloud.operators.bigquery.hashlib.md5")
-@pytest.mark.parametrize(
-    "test_dag_id, expected_job_id",
-    [("test-dag-id-1.1", "airflow_test_dag_id_1_1_test_job_id_2020_01_23T00_00_00_00_00_hash")],
-    ids=["test-dag-id-1.1"],
-)
-def test_job_id_validity(mock_md5, test_dag_id, expected_job_id):
```
kaxil, line 151: We should be able to test that `self.hook.generate_job_id(` is called with correct arguments

astronomer/providers/google/cloud/operators/bigquery.py
kaxil: Why did we have to add the type ignore?

rajaths010494: The older version had a `self._job_id(context)` method that has been moved to hook `generate_job_id`.
type ignore has been added because mypy is unable to get attributes in inherited classes. it's running fine in `make run-mypy` but on circle ci it fails.

tests/google/cloud/operators/test_bigquery.py
rajaths010494: Ok Will add a test case

astronomer/providers/google/cloud/operators/bigquery.py
kaxil: But with these changes in this PR, if someone used older version of the provider, this will fail, right?

kaxil: >type ignore has been added because mypy is unable to get attributes in inherited classes. it's running fine in make run-mypy but on circle ci it fails.

Are you sure about that?  It is possible that the CI is using the old provider where this method doesn't exist

rajaths010494: > But with these changes in this PR, if someone used older version of the provider, this will fail, right?

I have updated the google provider version in setup.cfg

kaxil: see now CI failed too :) - https://app.circleci.com/pipelines/github/astronomer/astronomer-providers/2175/workflows/c4283223-3bee-4964-8d92-9882be8790d3/jobs/8259 saying you don't need a typeignore

rajaths010494: > see now CI failed too :) - https://app.circleci.com/pipelines/github/astronomer/astronomer-providers/2175/workflows/c4283223-3bee-4964-8d92-9882be8790d3/jobs/8259 saying you don't need a typeignore

Understood with change in google provider version mypy passed since `generate_job_id` exists.

tests/google/cloud/operators/test_bigquery.py
rajaths010494: Added test case

PR #487, opened 2022-06-30
Fix Hadoop and Create Bucket Task Dependency 

pankajastro:
- Fix dependency between Hadoop task and create_bucket task i.e Hadoop task will depend on create_bucket
- Add an end task that will report the correct status of DAG run

### DIFF
--- a/astronomer/providers/google/cloud/example_dags/example_dataproc.py
+++ b/astronomer/providers/google/cloud/example_dags/example_dataproc.py
@@ -172,6 +172,5 @@ with models.DAG(
     )
     # [END how_to_cloud_dataproc_delete_cluster_operator]
 
-    create_cluster >> update_cluster
     create_cluster >> update_cluster >> pig_task >> hive_task >> delete_cluster
     (create_cluster >> update_cluster >> spark_task >> spark_sql_task >> hadoop_task >> delete_cluster)
--- a/astronomer/providers/google/cloud/example_dags/example_dataproc.py
+++ b/astronomer/providers/google/cloud/example_dags/example_dataproc.py
@@ -4,6 +4,11 @@ import os
 from datetime import datetime, timedelta
 
 from airflow import models
+from airflow.operators.empty import EmptyOperator
+from airflow.providers.google.cloud.operators.gcs import (
+    GCSCreateBucketOperator,
+    GCSDeleteBucketOperator,
+)
 
 from astronomer.providers.google.cloud.operators.dataproc import (
     DataprocCreateClusterOperatorAsync,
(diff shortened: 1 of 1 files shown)


rajaths010494:
> It seems that the Hadoop job needs an output location here: https://github.com/astronomer/astronomer-providers/pull/487/files#diff-bda2a1da7c8133be14f3184e8bf721ada183b9c4ea8595412fca5d7eaeabeb36R87
> 
> which I suppose is this bucket that we create for and later delete. So, I guess we need it and could not be removed from the self-sufficient DAG.
> 
> @rajaths010494 Can you confirm this?

Yes, @pankajastro  Hadoop job needs a location to write output. So we need a bucket and we delete it later.

pankajastro:
> Yes, @pankajastro Hadoop job needs a location to write output. So we need a bucket and we delete it later.

Aha, I see.
But the current task dependency is not correct i.e hadoop_task should depend on create_bucket
```
create_cluster >> update_cluster >> create_bucket
    create_cluster >> update_cluster >> pig_task >> hive_task >> delete_cluster >> delete_bucket
    (
        create_cluster
        >> update_cluster
        >> spark_task
        >> spark_sql_task
        >> hadoop_task
        >> delete_cluster
        >> delete_bucket
    )
```

It should be like 

```
create_cluster >> update_cluster >> pig_task >> hive_task >> delete_cluster >> delete_bucket
    (
        create_cluster
        >> update_cluster
        >> create_bucket
        >> spark_task
        >> spark_sql_task
        >> hadoop_task
        >> delete_cluster
        >> delete_bucket
    )
```
Ok, I'll fix this.

astronomer/providers/google/cloud/example_dags/example_dataproc.py
```diff
@@ -198,14 +199,17 @@
     )
     # [END howto_delete_buckettask]
 
-    create_cluster >> update_cluster >> create_bucket
-    create_cluster >> update_cluster >> pig_task >> hive_task >> delete_cluster >> delete_bucket
+    end = EmptyOperator(task_id="end")
+
+    create_cluster >> update_cluster >> hive_task >> spark_task >> spark_sql_task >> delete_cluster
```
pankajkoti, line 204: Is spark_task dependent upon hive_task? If not, it can run in parallel to the hive_task.

Also, previously it looks like hive_task is dependent on pig_task. Is that not a case? @rajaths010494 can you confirm?

rajaths010494: No it's not dependent we can run in parallel. Hive task is dependent on creating a bucket

pankajastro:
```
HIVE_JOB = {
    "reference": {"project_id": PROJECT_ID},
    "placement": {"cluster_name": CLUSTER_NAME},
    "hive_job": {"query_list": {"queries": ["SHOW DATABASES;"]}},
}
```

rajaths010494: This looks good 👍 .

pankajkoti (approved):

Issue #243, opened 2022-04-18
Add base trigger for Emr


pankajastro:
Add a base trigger for EmrContainerOperatorAsync, EmrContainerSensorAsync and reuse the code

PR #488, opened 2022-06-30
Add EMR Container Base Trigger

pankajastro:
closes: https://github.com/astronomer/astronomer-providers/issues/243

### DIFF
--- a/astronomer/providers/amazon/aws/triggers/emr.py
+++ b/astronomer/providers/amazon/aws/triggers/emr.py
@@ -17,7 +17,7 @@ class EmrContainerBaseTrigger(BaseTrigger, ABC):
 
     :param virtual_cluster_id: Reference Emr cluster id
     :param job_id:  job_id to check the state
-    :param max_tries: maximum retry for poll for the status
+    :param max_tries: maximum try attempts for polling the status
     :param aws_conn_id: Reference to AWS connection id
     :param poll_interval: polling period in seconds to check for the status
     """
@@ -88,6 +88,11 @@ class EmrContainerSensorTrigger(EmrContainerBaseTrigger):
 class EmrContainerOperatorTrigger(EmrContainerBaseTrigger):
     """Poll for the status of EMR container until reaches terminal state"""
 
+    INTERMEDIATE_STATES: List[str] = ["PENDING", "SUBMITTED", "RUNNING"]
+    FAILURE_STATES: List[str] = ["FAILED", "CANCELLED", "CANCEL_PENDING"]
+    SUCCESS_STATES: List[str] = ["COMPLETED"]
+    TERMINAL_STATES: List[str] = ["COMPLETED", "FAILED", "CANCELLED", "CANCEL_PENDING"]
+
     def serialize(self) -> Tuple[str, Dict[str, Any]]:
         """Serializes EmrContainerOperatorTrigger arguments and classpath."""
         return (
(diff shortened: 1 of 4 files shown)


astronomer/providers/amazon/aws/triggers/emr.py
```diff
@@ -176,7 +148,7 @@ async def run(self) -> AsyncIterator["TriggerEvent"]:  # type: ignore[override]
                         "Try %s: Query is still in non-terminal state - %s", try_number, query_state
                     )
                     await asyncio.sleep(self.poll_interval)
-                if self.max_tries and try_number >= self.max_tries:
+                if self.max_retries and try_number >= self.max_retries:
```
pankajkoti, line 151: This implementation looks incorrect to me. Retries are meant for failures. But, looking at the logic here it means that every time a poll is made as part of the `while True` loop, when the query is in non-terminal state, the `try_number` is incremented. The variable is functioning more like `max_poll_attempts`

pankajastro: Let me verify with sync version

pankajastro: the behaviour looks correct to me maybe the variable name is a bit misleading here
https://github.com/apache/airflow/blob/main/airflow/providers/amazon/aws/operators/emr.py#L209
https://github.com/apache/airflow/blob/main/airflow/providers/amazon/aws/sensors/emr.py#L165

pankajkoti: okay. the docstring has the variable name `max_tries`. Can we keep it the same then? Retry would mean something failed and hence we're retrying. 
Screenshot 2022-07-01 at 3 50 49 PM

pankajastro: rename here b539cd168b71bda61d69c0542367d7d7e9d353b0

astronomer/providers/amazon/aws/triggers/emr.py
```diff
@@ -10,14 +11,13 @@
...
+class EmrContainerBaseTrigger(BaseTrigger, ABC):
     """
-    The EmrContainerSensorTrigger is triggered when EMR container is created, it polls for the AWS EMR EKS Virtual
-    Cluster Job status. It is fired as deferred class with params to run the task in trigger worker
+    Poll for the status of EMR container until reaches terminal state
 
     :param virtual_cluster_id: Reference Emr cluster id
     :param job_id:  job_id to check the state
-    :param max_retries: maximum retry for poll for the status
+    :param max_tries: maximum retry for poll for the status
```
pankajkoti, line 20:
```suggestion
    :param max_tries: maximum try attempts for polling the status
```

astronomer/providers/amazon/aws/triggers/emr.py
```diff
@@ -80,60 +85,27 @@ async def run(self) -> AsyncIterator["TriggerEvent"]:  # type: ignore[override]
...
                 "max_tries": self.max_tries,
                 "poll_interval": self.poll_interval,
             },
         )
 
+    INTERMEDIATE_STATES: List[str] = ["PENDING", "SUBMITTED", "RUNNING"]
+    FAILURE_STATES: List[str] = ["FAILED", "CANCELLED", "CANCEL_PENDING"]
+    SUCCESS_STATES: List[str] = ["COMPLETED"]
+    TERMINAL_STATES: List[str] = ["COMPLETED", "FAILED", "CANCELLED", "CANCEL_PENDING"]
+
```
pankajkoti, line 108: Would be good if we move to this before the first method of the class at the top for readability since these are class variables.

pankajkoti (approved):

pankajkoti:
Please update the description to include what and why.

astronomer/providers/amazon/aws/triggers/emr.py
pankajastro: yeah, somehow I missed it. Fixed here cdea7279867152df20182404e5602934a2cb3c61

PR #489, opened 2022-07-01
Update FileSensorAsync Example to Defer

pankajastro:
Since the file was already present before triggering `file_sensor_task` so it was not waiting and defering this PR make sure that the current will example defer  
- Add start and end task
- add bash task to wait for 10 sec and create a file
- add a task to delete the created file
- Deleted that hardcoded file example_file_async_sensor.txt

Screenshot 2022-07-01 at 3 12 07 PM

### DIFF
--- a/astronomer/providers/core/example_dags/example_file_sensor.py
+++ b/astronomer/providers/core/example_dags/example_file_sensor.py
@@ -2,6 +2,8 @@ import os
 from datetime import timedelta
 
 from airflow import DAG
+from airflow.operators.bash import BashOperator
+from airflow.operators.empty import EmptyOperator
 from airflow.utils.timezone import datetime
 
 from astronomer.providers.core.sensors.filesystem import FileSensorAsync
@@ -24,10 +26,30 @@ with DAG(
     default_args=default_args,
     tags=["example", "async", "core"],
 ) as dag:
+    start = EmptyOperator(task_id="start")
+
+    create_file = BashOperator(
+        task_id="create_file",
+        bash_command="sleep 10 && touch /usr/local/airflow/dags/example_file_async_sensor.txt",
+    )
+
     # [START howto_sensor_filesystem_async]
     file_sensor_task = FileSensorAsync(
         task_id="file_sensor_task",
-        filepath="example_file_async_sensor.txt",
+        filepath="/usr/local/airflow/dags/example_file_async_sensor.txt",
         fs_conn_id=FS_CONN_ID,
+        poke_interval=3,
     )
     # [END howto_sensor_filesystem_async]
+
+    delete_file = BashOperator(
+        task_id="delete_file",
+        bash_command="rm /usr/local/airflow/dags/example_file_async_sensor.txt",
+        trigger_rule="all_done",
+    )
+
+    end = EmptyOperator(task_id="end")
+
+    start >> [file_sensor_task, create_file]
+    [create_file, file_sensor_task] >> delete_file
+    [file_sensor_task, create_file, delete_file] >> end


astronomer/providers/core/example_dags/example_file_sensor.py
```diff
@@ -24,10 +26,30 @@
     default_args=default_args,
     tags=["example", "async", "core"],
 ) as dag:
+    start = EmptyOperator(task_id="start")
+
+    create_file = BashOperator(
+        task_id="create_file",
+        bash_command="sleep 10 && touch /usr/local/airflow/dags/example_file_async_sensor.txt",
```
pankajkoti, line 33: Since we're creating and sensing this file, perhaps we don't need to create it in the DAGs directory. How about creating it in the /tmp/ directory?
```suggestion
        bash_command="sleep 10 && touch /tmp/example_file_async_sensor.txt",
```

astronomer/providers/core/example_dags/example_file_sensor.py
```diff
@@ -24,10 +26,30 @@
...
+    )
+
     # [START howto_sensor_filesystem_async]
     file_sensor_task = FileSensorAsync(
         task_id="file_sensor_task",
-        filepath="example_file_async_sensor.txt",
+        filepath="/usr/local/airflow/dags/example_file_async_sensor.txt",
```
pankajkoti, line 39:
```suggestion
        filepath="/tmp/example_file_async_sensor.txt",
```
If the above suggestion can be considered.

astronomer/providers/core/example_dags/example_file_sensor.py
```diff
@@ -24,10 +26,30 @@
...
         fs_conn_id=FS_CONN_ID,
+        poke_interval=3,
     )
     # [END howto_sensor_filesystem_async]
+
+    delete_file = BashOperator(
+        task_id="delete_file",
+        bash_command="rm /usr/local/airflow/dags/example_file_async_sensor.txt",
```
pankajkoti, line 47:
```suggestion
        bash_command="rm /tmp/example_file_async_sensor.txt",
```

pankajkoti (approved):
Looks like a good improvement. Have a minor suggestion if you would like to consider.

astronomer/providers/core/example_dags/example_file_sensor.py
pankajastro: This file needs to be accessable from the trigger since we are checking file status from the trigger.
so if I'll create a file with bash operator in /tmp/ folder then it would be available in the worker /tmp/ folder which trigger can't access but if I'll create in dags then it can access  since dags folder is a shared folder

pankajkoti: Is that only the DAGs folder is synced across Airflow components or the whole directory /usr/local/airflow synced? Maybe we can keep it there then. Only concern is that the dag processor would try to process this file and then find that it is not a DAG; not sure if we need to worry for that optimisation, but just a thought.

@kaxil do you have some inputs here?

pankajastro: In future maybe either we can add `AIRFLOW__CORE__DAG_IGNORE_FILE_SYNTAX` in dockerfile or we can create a `. airflowignore`

pankajastro: tracking here https://github.com/astronomer/astronomer-providers/issues/491

kaxil: Any file apart from `.py` files will be ignored when parsing for DAG files so not a problem

Issue #495, opened 2022-07-04
Fix example_async_bigquery_queries dependency on delete_dataset task


pankajkoti:
It is observed that the delete_dataset task is called even before
the dependent tasks finish processing which depend on the dataset
to be queried upon. 

 Attaching the current incorrect dependency screenshot.![Screenshot 2022-07-04 at 3.45.07 PM.png](https://images.zenhubusercontent.com/624476b10ffe431c7bb98647/65c2a9b0-b43e-43e0-814d-8b5df435646c)

Following is the log where the  select_query_job relies on the dataset but the dataset gets deleted even before the task finished processing.
```
[2022-07-04, 06:14:51 UTC] {base.py:68} INFO - Using connection ID 'google_cloud_default' for task execution.
[2022-07-04, 06:14:51 UTC] {bigquery.py:67} INFO - Inserting job airflow_example_async_bigquery_queries_select_query_job_2022_07_04T06_12_30_471149_00_00_2463898085b7a59b21423852ed291692
[2022-07-04, 06:14:51 UTC] {taskinstance.py:1889} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/astronomer/providers/google/cloud/operators/bigquery.py", line 100, in execute
    self._handle_job_error(job)
  File "/usr/local/lib/python3.9/site-packages/airflow/providers/google/cloud/operators/bigquery.py", line 2135, in _handle_job_error
    raise AirflowException(f"BigQuery job {job.job_id} failed: {job.error_result}")
airflow.exceptions.AirflowException: BigQuery job airflow_example_async_bigquery_queries_select_query_job_2022_07_04T06_12_30_471149_00_00_2463898085b7a59b21423852ed291692 failed: {'reason': 'notFound', 'message': 'Not found: Dataset astronomer-airflow-providers:astro_dataset was not found in location US'}

```

PR #496, opened 2022-07-04
Fix example_async_bigquery_queries dependency

pankajkoti:
It is observed that the `delete_dataset` task is called even before
the dependent tasks finish processing which depend on the dataset
to be queried upon. The commit fixes this issue by adding appropriate
dependecny in the DAG chain.

Also, since the `delete_dataset` task is called with the `trigger_rule`
`all_done`, we're adding an end task to report the DAG status correctly,
i.e. fail the DAG if any of the tasks fail.

closes: #495

### DIFF
--- a/astronomer/providers/google/cloud/example_dags/example_bigquery_queries.py
+++ b/astronomer/providers/google/cloud/example_dags/example_bigquery_queries.py
@@ -7,6 +7,7 @@ from datetime import datetime, timedelta
 
 from airflow import DAG
 from airflow.operators.bash import BashOperator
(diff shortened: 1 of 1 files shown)


pankajkoti:
Attaching before and after DAG dependency chain screenshots. <br>
Before:
Screenshot 2022-07-04 at 2 45 00 PM

<br>
After:
Screenshot 2022-07-04 at 3 48 19 PM

bharanidharan14 (approved):
LGTM

pankajastro (approved):

PR #509, opened 2022-07-12
Upgrade base Airflow image to latest release

pankajkoti:
astro-runtime:5.0.6 and ap-airflow:2.3.3-1 is released on
11th July 2022. The commit updates our Dockerfile to reference these
as the base images.


rajaths010494 (approved):

phanikumv (changes requested):
We should try to change the release versions at one place and read it in all the files, rather than changing in all file

pankajkoti:
@phanikumv I have made changes to have a common variable for `make` targets to use the image name. 

With this we will need to update the images in only 2 places now, 1 for the `make` targets and 1 for the CircleCI deployment image.

phanikumv (approved):

PR #525, opened 2022-07-15
Remove EMR Conn from example

pankajastro:
Now `emr_conn_id` is optional in EmrCreateJobFlowOperator since the configuration we pass using param `job_flow_overrides` so the `emr_conn_id` is no longer require. We can merge this once the provider for June is released `apache-airflow-providers-amazon==4.1.0`
OSS PR: https://github.com/apache/airflow/pull/24306

### DIFF
--- a/astronomer/providers/amazon/aws/example_dags/example_emr_sensor.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_emr_sensor.py
@@ -18,7 +18,6 @@ from astronomer.providers.amazon.aws.sensors.emr import (
 JOB_FLOW_ROLE = os.getenv("EMR_JOB_FLOW_ROLE", "EMR_EC2_DefaultRole")
 SERVICE_ROLE = os.getenv("EMR_SERVICE_ROLE", "EMR_DefaultRole")
 AWS_CONN_ID = os.getenv("ASTRO_AWS_CONN_ID", "aws_default")
-EMR_CONN_ID = os.getenv("ASTRO_EMR_CONN_ID", "emr_default")
 EXECUTION_TIMEOUT = int(os.getenv("EXECUTION_TIMEOUT", 6))
 
 SPARK_STEPS = [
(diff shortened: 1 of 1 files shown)


astronomer/providers/amazon/aws/example_dags/example_emr_sensor.py
```diff
@@ -72,7 +71,7 @@
     cluster_creator = EmrCreateJobFlowOperator(
         task_id="create_job_flow",
         job_flow_overrides=JOB_FLOW_OVERRIDES,
-        emr_conn_id=EMR_CONN_ID,
+        aws_conn_id=AWS_CONN_ID,
```
pankajkoti, line 74: will this fail for apache-airflow-providers-amazon < 4.1.0 ? We should make it fail safe for previous versions of the provider then. Something similar to this, with a try-except block catching the exception: https://github.com/astronomer/astronomer-providers/pull/471#discussion_r907184079

pankajastro: hmm, it would fail for apache-airflow-providers-amazon < 4.1.0. Does it make sense to safe fail here since this is an integration test?

pankajkoti: @phanikumv @bharanidharan14 what is your opinion on this?

rajaths010494: I am not sure if we can use try-catch to an argument being passed to an operator.

rajaths010494: > I am not sure if we can use try-catch for an argument being passed to the operator.

pankajkoti: Perhaps we can add a comment in the example DAG that based on the versions of the provider what arguments are expected

bharanidharan14: @pankajastro adding comments in the example DAG is reasonable to me, so it helps user to know what arguements needs to be passed

bharanidharan14: @pankajastro Is this review comment is addressed ?

bharanidharan14 (approved):

PR #543, opened 2022-07-22
Change EMR EKS instance type

phanikumv:
Use `m5.xlarge`  instance type which provides more memory for EMR EKS cluster to be able to execute Spark jobs successfully.


astronomer/providers/amazon/aws/example_dags/example_emr_eks_containers_job.py
```diff
@@ -22,7 +22,7 @@
 DEBUGGING_MONITORING_POLICY = os.getenv("DEBUGGING_MONITORING_POLICY", "test_debugging_monitoring_policy")
 EKS_CLUSTER_NAME = os.getenv("EKS_CLUSTER_NAME", "providers-team-eks-cluster")
 EKS_NAMESPACE = os.getenv("EKS_NAMESPACE", "providers-team-eks-namespace")
-INSTANCE_TYPE = os.getenv("INSTANCE_TYPE", "m4.large")
+INSTANCE_TYPE = os.getenv("INSTANCE_TYPE", "m5.xlarge")
```
pankajkoti, line 25: Let's rename this variable to have some additional prefix e.g. EMR_INSTANCE_TYPE as the current name caused some conflict with Airflow Operator variable with the same name.

Also, we can create this environment variable in our Astro Cloud deployment if we wish to change the value later.

phanikumv: Please review now, I have changed it

pankajkoti (approved):

PR #551, opened 2022-07-25
Werkzeug version downgrade

phanikumv:
Pin the werkzeug version down < 2.2.0 to resolve the below error.

`from werkzeug.routing import parse_rule
E   ImportError: cannot import name 'parse_rule' from 'werkzeug.routing' (/home/circleci/.pyenv/versions/3.8.13/lib/python3.8/site-packages/werkzeug/routing/__init__.py)
____ ERROR at setup of test_get_conn_params_exception[connection_details3-] ____`

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,16 +1,6 @@
 Changelog
 =========
 
-1.7.1 (2022-07-25)
-------------------
-
-Bug Fixes
-"""""""""
-
-- Bump up  ``MarkupSafe`` version as per Airflow 2.3.3 constraints
-  (`#542 <https://github.com/astronomer/astronomer-providers/pull/542>`_)
-
-
 1.7.0 (2022-07-19)
 ------------------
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -1,6 +1,16 @@
 Changelog
 =========
 
+1.7.1 (2022-07-25)
```
pankajkoti, line 4: Should this branch be cut from main instead? Looks like this is cut from release branch. If we merge this, then this would go in too right?

setup.cfg
```diff
@@ -40,6 +40,7 @@ install_requires =
     asgiref
     typing_extensions; python_version < "3.8"
     markupsafe>=1.1.1
+    werkzeug<2.2.0
```
pankajkoti, line 43: Can we add a comment on why we are pinning it to <2.2.0 ?

pankajkoti (approved):

setup.cfg
phanikumv: Added the comments

pankajastro (approved):

PR #552, opened 2022-07-25
Fix hardcoded poll interval

bharanidharan14:
- Fixed hardcoded value for poll interval now, passed it from trigger
- Fix test for the changes made
- Changed the `polling_period_seconds` name to `poll_interval`


astronomer/providers/snowflake/triggers/snowflake_trigger.py
```diff
@@ -62,7 +62,7 @@ async def run(self) -> AsyncIterator["TriggerEvent"]:  # type: ignore[override]
         """
         hook = get_db_hook(self.snowflake_conn_id)
         try:
-            run_state = await hook.get_query_status(self.query_ids)
+            run_state = await hook.get_query_status(self.query_ids, self.polling_period_seconds)
```
pankajastro, line 65: I think we should keep variable names consistent in operator/sensor/hook/trigger. So please consider renaming `polling_period_seconds` to `polling_inverval` since the operator has `polling_inverval` and at most of the places for polling we have `polling_inverval` variable name

pankajastro (approved):

astronomer/providers/snowflake/hooks/snowflake.py
```diff
@@ -91,7 +91,9 @@ def check_query_output(self, query_ids: List[str]) -> None:
                     self.log.info("Rows affected: %s", cur.rowcount)
                     self.log.info("Snowflake query id: %s", query_id)
 
-    async def get_query_status(self, query_ids: List[str]) -> Dict[str, Union[str, List[str]]]:
+    async def get_query_status(
+        self, query_ids: List[str], poll_intervals: float
```
pankajkoti, line 95: This should be poll_interval and not poll_intervals as it is as single value

bharanidharan14: Got it

astronomer/providers/snowflake/triggers/snowflake_trigger.py
bharanidharan14: Changed it to `poll_interval`

Issue #566, opened 2022-08-02
Broken links for class definitions in the Operators list page


pankajkoti:
The links on https://astronomer-providers.readthedocs.io/en/stable/providers/operators_and_sensors_list.html for class definitions are not working.

PR #567, opened 2022-08-02
Update base url for API docs to fix broken links

pankajkoti:
Correct the base URL for https://astronomer-providers.readthedocs.io/
so that class def links from operators & sensors list page work
correctly.

closes: #566

### DIFF
--- a/docs/_ext/traverse_operators_sensors.py
+++ b/docs/_ext/traverse_operators_sensors.py
@@ -17,7 +17,7 @@ BASE_IMPORT_PATH = "astronomer.providers"
 CURRENT_FILE_PATH = Path(__file__)
 ASTRONOMER_PROVIDERS_PATH = CURRENT_FILE_PATH.parent.parent.parent / "astronomer" / "providers"
 
-BASE_CLASS_DEF_URL = "/astronomer-providers/docs/_build/html/_api/"
+BASE_CLASS_DEF_URL = "/en/stable/_api/"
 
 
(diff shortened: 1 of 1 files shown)


bharanidharan14 (approved):

docs/_ext/traverse_operators_sensors.py
```diff
@@ -17,7 +17,7 @@
 CURRENT_FILE_PATH = Path(__file__)
 ASTRONOMER_PROVIDERS_PATH = CURRENT_FILE_PATH.parent.parent.parent / "astronomer" / "providers"
 
-BASE_CLASS_DEF_URL = "/astronomer-providers/docs/_build/html/_api/"
+BASE_CLASS_DEF_URL = "/en/stable/_api/"
```
kaxil, line 20: Is there a way we could dynamically get the base URL? Otherwise this would point to "stable" docs even for "latest".

Example: https://astronomer-providers.readthedocs.io/en/latest/providers/operators_and_sensors_list.html

pankajkoti: Created another PR to address the change needed to use the right version of the document: https://github.com/astronomer/astronomer-providers/pull/569

We're in need of using javascript here until we move everything to native Sphinx with the work on this ticket: https://app.zenhub.com/workspaces/airflow-providers-team-61a89ee7fbf048001dc8378d/issues/astronomer/astronomer-providers/508

PR #587, opened 2022-08-16
Exclude files containing dead links

feluelle:


rajaths010494 (approved):

.pre-commit-config.yaml
```diff
@@ -39,6 +39,8 @@ repos:
     rev: v3.10.2
     hooks:
       - id: markdown-link-check
+        # Ignoring files containing dead links which are actually not dead.
+        exclude: dev/README.md|CODE_OF_CONDUCT.md
```
pankajkoti, line 43: I believe we should not exclude these files here as that would beat the purpose of the `markdown-link-check` pre-commit check then. Can we check if something else is wrong if the links are not actually dead or if there is some issue with the pre-commit hook itself?

feluelle: I agree. This was only being done to unblock other PRs. We should find a way to fix the markdown issues.

PR #588, opened 2022-08-16
Release 1.8.0

rajaths010494:
Update `setup.cfg`, `docs/conf.py` and `CHANGELOG.rst` with the changes pertaining to release 1.8.0

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -61,6 +61,7 @@ cncf.kubernetes =
     kubernetes_asyncio
 databricks =
     apache-airflow-providers-databricks>=2.2.0
+    databricks-sql-connector>=2.0.4;python_version>='3.10'
 google =
     apache-airflow-providers-google>=8.1.0
     gcloud-aio-storage
@@ -123,6 +124,7 @@ all =
     apache-airflow-providers-http
     apache-airflow-providers-snowflake
     apache-airflow-providers-microsoft-azure
+    databricks-sql-connector>=2.0.4;python_version>='3.10'
     gcloud-aio-bigquery
     gcloud-aio-storage
     kubernetes_asyncio
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,6 @@
 [metadata]
 name = astronomer-providers
-version = 1.8.0.dev1
+version = 1.8.0
 url = https://github.com/astronomer/astronomer-providers/
 author = Astronomer
 author_email = humans@astronomer.io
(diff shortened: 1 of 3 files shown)


phanikumv (changes requested):
Tests are failing. Please resolve it

uranusjr:
@rajaths010494 and I just debugged through this. The root problem of the CI failure is [databricks-sql-python does not support Python 3.10](https://github.com/databricks/databricks-sql-python/issues/26), this means that `pip install astronomer-providers[databricks]` cannot succeed on 3.10. There are a couple of solutions here:

1. Just ignore the CI failure, merge this PR and release 1.8.0 as-is. This means the package will be installable in most scenarios, _except specifically with the `databricks` extra on 3.10_, where pip will emit an error like the CI curerntly does (not possible to find dependency resolution).
2. Remove 3.10 support until databricks adds 3.10 support. This ensures that no users will see any dependency error, but the downside is the release will not be installable on 3.10, even if the user is not using databricks.

Personally I’m leaning toward the first solution, but it does need to involve some human judgement (not fully automated). So I would definitely understand if y’all decide to go with the second solution instead.

uranusjr (approved):
Personally I would ignore the CI failure and do the release.

kaxil:
We should keep an eye out on https://github.com/databricks/databricks-sql-python/pull/32 and https://github.com/databricks/databricks-sql-python/pull/31 that might tackle it 🤷

kaxil:
Looks like they are going to release a new version supporting python 3.10 in a few hours 🎉 

https://github.com/databricks/databricks-sql-python/pull/33#issuecomment-1218375892

We can just wait for a day for that :)

kaxil:
They have released 2.0.4 🎉 and I have added the changes to this PR, we are good to go

kaxil (approved):

bharanidharan14 (approved):

bharanidharan14 (approved):

setup.cfg
```diff
@@ -61,6 +61,7 @@ cncf.kubernetes =
     kubernetes_asyncio
 databricks =
     apache-airflow-providers-databricks>=2.2.0
+    databricks-sql-connector>=2.0.4;python_version>='3.10'
```
phanikumv, line 64: Shouldnt we revert back this change back to 1.7.1 , now that we have 3.10 support?
```suggestion
    apache-airflow-providers-databricks>=2.2.0
```

rajaths010494: Yes pip is able to look for compatible versions needed for python3.10 and installs the `databricks-sql-connector>=2.0.4`.
Will make the necessary changes.

rajaths010494:
> * Tests are failing. Please resolve it
> * Include the details on the PR #589 in the `CHANGELOG.rst`

Added the details of PR #589 in  `CHANGELOG.rst`

phanikumv (approved):

PR #591, opened 2022-08-17
Fix issues for 1.8

rajaths010494:


setup.cfg
```diff
@@ -40,6 +40,8 @@ install_requires =
     asgiref
     typing_extensions; python_version < "3.8"
     markupsafe>=1.1.1
+    pyarrow==6.0.1
+    databricks-sql-connector==2.0.2
```
uranusjr, line 44:
```suggestion
    pyarrow>=6
    databricks-sql-connector
```

This is better if it “works” (in the sense that pip allows it… no idea if it would actually “work”).

rajaths010494:
Closing this PR as `databricks-sql-connector>=2.0.4` has released providing support for python3.10

Issue #606, opened 2022-08-22
Fix the snowflake extractor import in dev docker 


pankajastro:
Getting below error in container 

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/airflow/plugins_manager.py", line 227, in load_entrypoint_plugins
    plugin_class = entry_point.load()
  File "/usr/local/lib/python3.9/site-packages/importlib_metadata/__init__.py", line 203, in load
    module = import_module(match.group('module'))
  File "/usr/local/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 850, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/usr/local/lib/python3.9/site-packages/openlineage/airflow/plugin.py", line 25, in <module>
    from openlineage.airflow import listener
  File "/usr/local/lib/python3.9/site-packages/openlineage/airflow/listener.py", line 74, in <module>
    extractor_manager = ExtractorManager()
  File "/usr/local/lib/python3.9/site-packages/openlineage/airflow/extractors/manager.py", line 15, in __init__
    self.task_to_extractor = Extractors()
  File "/usr/local/lib/python3.9/site-packages/openlineage/airflow/extractors/extractors.py", line 80, in __init__
    extractor = import_from_string(extractor)
  File "/usr/local/lib/python3.9/site-packages/openlineage/airflow/utils.py", line 303, in import_from_string
    raise ImportError(f"Failed to import {path}") from e
ImportError: Failed to import  astronomer.providers.snowflake.extractors.snowflake.SnowflakeAsyncExtractor
```

pankajastro:
<img width="1036" alt="Screenshot 2022-08-23 at 3 27 56 AM" src="https://user-images.githubusercontent.com/98807258/186026028-c92e060c-1c95-4cb9-a055-4b996a63d65b.png">

PR #607, opened 2022-08-22
Fix import error for snowflake extractor

pankajastro:
look like `>-` in YAML put a space at the end for multiline string because of that import was failing in dev container for the snowflake extractor 

closes: https://github.com/astronomer/astronomer-providers/issues/606

### DIFF
--- a/dev/docker-compose.yaml
+++ b/dev/docker-compose.yaml
@@ -22,9 +22,8 @@ x-airflow-common:
     ASTRONOMER_ENVIRONMENT: local
     AIRFLOW__LINEAGE__BACKEND: openlineage.lineage_backend.OpenLineageBackend
     OPENLINEAGE_URL: http://host.docker.internal:5050/
-    OPENLINEAGE_EXTRACTORS: >-
-      astronomer.providers.google.cloud.extractors.bigquery.BigQueryAsyncExtractor;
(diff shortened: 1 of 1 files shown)


pankajkoti:
Does this approach work @pankajastro . Mean to ask with this does it no longer complain for import error and that both the extractors are loaded and run fine? I believe @rajaths010494 had checked previously with Maciej from GetInData and he had mentioned that this is an open issue on OL where multiple extractors load is failing for some reason.

pankajastro:
> Does this approach work @pankajastro . Mean to ask with this does it no longer complain for import error and that both the extractors are loaded and run fine? I believe @rajaths010494 had checked previously with Maciej from GetInData and he had mentioned that this is an open issue on OL where multiple extractors load is failing for some reason

This more of looks like related YAML string parsing for example if I'm keeping the env like 
`OPENLINEAGE_EXTRACTORS: astronomer.providers.google.cloud.extractors.bigquery.BigQueryAsyncExtractor;astronomer.providers.snowflake.extractors.snowflake.SnowflakeAsyncExtractor` i.e in a single like then it at least not complaining in UI.

Coming to your question, at least it not complaining in UI and also the error in not in the container.  Do I need to do more testing like running some examples etc?
Screenshot 2022-08-23 at 11 43 54 AM

pankajkoti (approved):

PR #608, opened 2022-08-25
Compare timeouts in ADF triggerer realtime

preethamrajk:
Currently, the design assumes that the triggerer process and the worker process are working in the same execution context(pod/vm etc).

While monotonic time functions are the right way to compare timeouts, they are required to have the same reference point during time comparison.

For this reason, this PR looks at moving away from the `end_time` parameter in the Triggerer and passes the task's timeout value directly to the Triggerer so that the time comparison can be done in the triggerer process.

### DIFF
--- a/astronomer/providers/microsoft/azure/operators/data_factory.py
+++ b/astronomer/providers/microsoft/azure/operators/data_factory.py
@@ -1,4 +1,3 @@
-import time
 from typing import Any, Dict
 
 from airflow.exceptions import AirflowException
--- a/astronomer/providers/microsoft/azure/operators/data_factory.py
+++ b/astronomer/providers/microsoft/azure/operators/data_factory.py
@@ -69,7 +69,7 @@ class AzureDataFactoryRunPipelineOperatorAsync(AzureDataFactoryRunPipelineOperat
                 resource_group_name=self.resource_group_name,
                 factory_name=self.factory_name,
                 check_interval=self.check_interval,
-                end_time=end_time,
+                end_time=self.timeout,
             ),
             method_name="execute_complete",
         )
--- a/astronomer/providers/microsoft/azure/operators/data_factory.py
+++ b/astronomer/providers/microsoft/azure/operators/data_factory.py
@@ -68,7 +68,7 @@ class AzureDataFactoryRunPipelineOperatorAsync(AzureDataFactoryRunPipelineOperat
                 resource_group_name=self.resource_group_name,
                 factory_name=self.factory_name,
                 check_interval=self.check_interval,
-                end_time=self.timeout,
+                timeout=self.timeout,
             ),
             method_name="execute_complete",
         )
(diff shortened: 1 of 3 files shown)


preethamrajk:
Hi @bharanidharan14 @phanikumv @rajaths010494 @pankajastro @pankajkoti 

Please review the PR to fix time comparisons in the Data Factory Triggerer.

WRT the tests that are failing, they don't seem to be related to the code that is this folder.

pankajkoti:
Hi @preethamrajk . Thanks a lot for creating this PR and interest in contributing to the project.
We did observe that monotonic time causing problems due to worker and triggerer running in different contexts like you have mentioned in the PR description. We removed the use of montonic time and are using unix timestamp now with the PR: https://github.com/astronomer/astronomer-providers/pull/602 . This will be out in the next release. 

Your approach also does look good to me. I will like to hear opinion from others on this one.

preethamrajk:
> Hi @preethamrajk . Thanks a lot for creating this PR and interest in contributing to the project. We did observe that monotonic time causing problems due to worker and triggerer running in different contexts like you have mentioned in the PR description. We removed the use of montonic time and are using unix timestamp now with the PR: #602. This will be out in the next release.
> 
> Your approach also does look good to me. I will like to hear opinion from others too on this one.

Thanks for the response @pankajkoti. Curious to know when the next release is. And what is the release cycle is for the providers?

rajaths010494:
Triggerer when it restarts every time it sets a new end_time rather than a fixed end_time when a timeout should have happened.
As @pankajkoti suggested we have fixed the issue. Please do wait for the release.

preethamrajk:
That's a fair observation @rajaths010494.

That lead me to think a bit more..

I find it fundamentally incorrect trying to fail the task if the ADF Pipeline is taking too long. Shouldn't the pipeline continue to execute even if the it's taking too long? This will allow for another mechanism such as SLA callbacks to handle long running ADF pipelines without failing the data flows. Thoughts?

Happy to close this PR with my two cents.. thanks for reviewing promptly.

rajaths010494:
> That's a fair observation @rajaths010494.
> 
> That lead me to think a bit more..
> 
> I find it fundamentally incorrect trying to fail the task if the ADF Pipeline is taking too long. Shouldn't the pipeline continue to execute even if the it's taking too long? This will allow for another mechanism such as SLA callbacks to handle long running ADF pipelines without failing the data flows. Thoughts?
> 
> Happy to close this PR with my two cents.. thanks for reviewing promptly.

Users can define end_time using the operator for the long-running tasks so that it waits for a longer period if you feel end_time is unsure how much longer it can be then you can use `AzureDataFactoryPipelineRunStatusSensorAsync` given the run_id of the pipeline to poll until it reaches the desired state.
An example of the same can be found [here](https://github.com/astronomer/astronomer-providers/blob/main/astronomer/providers/microsoft/azure/example_dags/example_adf_run_pipeline.py#L188-L207)

pankajkoti:
Hi @preethamrajk New release https://github.com/astronomer/astronomer-providers/releases/tag/1.8.1 is out which has the needed change as discussed above. Please check and hope to hear back your from you soon on this and other contributions :) Thank you again!

preethamrajk:
Thanks @pankajkoti. Have tested this in my setup and works fine.

PR #615, opened 2022-09-01
Remove werkzeug dep

kaxil:
Since https://github.com/maxcountryman/flask-login/issues/686 is fixed and Airflow 2.4 (alpha at the moment) won't need it, we are good to remove it from our setup.cfg


kaxil:
I merged it with broken `mypy` so that I could fix it in the following PR so we can release 1.8.1

PR #619, opened 2022-09-03
Add compat module for typing `airflow.utils.context.Context`

josh-fell:
The `airflow.utils.context` module was not introduced to OSS Airflow until 2.2.3. Importing this module as a top-level import sets an implicit, minimum requirement for Airflow 2.2.3 which is higher than the minumum `apache-airflow` requirement set for Astronomer Providers. This PR adds a typing compat module to be used throughout the repo for consistent typing of `context`.


kaxil (approved):

astronomer/providers/amazon/aws/operators/batch.py
```diff
@@ -7,14 +7,17 @@
...
 
 from airflow.exceptions import AirflowException
 from airflow.providers.amazon.aws.operators.batch import BatchOperator
-from airflow.utils.context import Context
 
 from astronomer.providers.amazon.aws.triggers.batch import BatchOperatorTrigger
 
+if TYPE_CHECKING:
+    # TODO: This can be removed from TYPE_CHECKING once there is a minimum requirement of Airflow 2.2.3+
+    from airflow.utils.context import Context
```
kaxil, line 19: We have one more option:

1. Have a `compat` module in this repo and import `airflow.utils.context.Context` if it is available, or just set `Context = Dict` if it is not available, so it is consistent.


wydt?

phanikumv: I like the idea of having dynamic check using `compat` module.

@josh-fell could you please make the change and fix the code coverage.

josh-fell: > Have a compat module in this repo and import airflow.utils.context.Context if it is available, or just set Context = Dict if it is not available, so it is consistent.

+1 ❤️  this idea as well. I'll add a `compat` module and unify the typing `Context` across the repo.

astronomer/providers/utils/typing_compat.py
```diff
@@ -0,0 +1,19 @@
...
+# typically used as the typing for the ``context`` arg in operators and sensors. However, using this class for
+# typing outside of TYPE_CHECKING in modules sets an implicit, minimum requirement for Apache Airflow 2.2.3
+# which is currently more recent than the current minimum requirement of Apache Airflow 2.2.0.
+#
+# TODO: Remove this once the repo has a minimum Apache Airflow requirement of 2.2.3+.
+try:
+    from airflow.utils.context import Context
+except ModuleNotFoundError:
+
+    class Context(MutableMapping[str, Any]):  # type: ignore[no-redef]
```
josh-fell, line 13: With the approach of `Context = MutableMapping[str, Any]`, mypy was complaining about not being able to assign to a type and the variable assignment types not matching. Instead of ignoring those mypy errors (which felt icky), I decided to redefine the typing class and ignore the redefinition mypy error.

Does this seem reasonable? Any other mypy workarounds?

josh-fell:
The pre-commit.ci check is failing for `Executable `.circleci/scripts/pre_commit_context_typing_compat.py` is not executable`. I'm not sure how to resolve this. Both `make run-static-checks` and `pre-commit run --all-files` execute just fine locally for me. Any pointers?

josh-fell:
Of course, if there is a preference of having unit test over a pre-commit hook, I can change that no problem!

bharanidharan14:
@josh-fell Still code coverage is failing

josh-fell:
> @josh-fell Still code coverage is failing

Ah yes. Thanks for the poke. Just pushed a unit test in [e2a9ade](https://github.com/astronomer/astronomer-providers/pull/619/commits/e2a9ade35590ae165ed14a7f4833ba2603f099d0) which hopefully helps.

josh-fell:
[Good to know the pre-commit hook works](https://results.pre-commit.ci/run/github/388913563/1663036903.TsAyOpIjQs-PWu6wavydnQ). Updated the dbt Cloud modules.

Issue #620, opened 2022-09-05
Add check for PEM exists for Hive and Livy DAG.


rajaths010494:
While running hive or livy dag where we use ssh to login into the machine pem file was missing in /tmp folder.
This PR checks whether the file exists or not while copying the pem into /tmp.
If it doesn't exist then it will raise an exception.

rajaths010494:
raised the PR https://github.com/astronomer/astronomer-providers/pull/621

PR #621, opened 2022-09-05
Add check for PEM exists for Hive and Livy DAG.

rajaths010494:
While running hive or livy dag where we use ssh to login into the machine pem file was missing in /tmp folder.
Now PEM is downloaded into /tmp in SSH task so PEM is available at the same worker node. 
This PR checks whether the file exists or not while copying the pem into /tmp and downloading before SSH into the machine.
If it doesn't exist then it will raise an exception.

closes: #620

### DIFF
--- a/astronomer/providers/apache/hive/example_dags/example_hive.py
+++ b/astronomer/providers/apache/hive/example_dags/example_hive.py
@@ -206,10 +206,11 @@ def add_inbound_rule_for_security_group(task_instance: Any) -> None:
             raise error
 
 
-def create_key_pair() -> None:
+def ssh_and_run_command(task_instance: Any, **kwargs: Any) -> None:
     """
     Load the private_key from airflow variable and creates a pem_file
-    at /tmp/.
+    at /tmp/. SSH into the machine and execute the bash script from the list
+    of commands.
     """
     # remove the file if it exists
     if os.path.exists(f"/tmp/{PEM_FILENAME}.pem"):
@@ -221,17 +222,11 @@ def create_key_pair() -> None:
 
     # write private key to file with 400 permissions
     os.chmod(f"/tmp/{PEM_FILENAME}.pem", 0o400)
-    # check if the PEM file exists or not.
+    # Check if the PEM file exists or not.
     if not os.path.exists(f"/tmp/{PEM_FILENAME}.pem"):
-        # if it doesn't exists raise an error.
+        # if it doesn't exists raise an error
         raise AirflowException("PEM file wasn't copied properly.")
 
-
-def ssh_and_run_command(task_instance: Any, **kwargs: Any) -> None:
-    """
-    SSH into the machine and execute the bash script from the list
-    of commands.
-    """
     import paramiko
 
     key = paramiko.RSAKey.from_private_key_file(kwargs["path_to_pem_file"])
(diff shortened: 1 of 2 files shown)


pankajkoti (commented):
I am thinking that the tasks in the DAG could be running on different workers and hence there could be failure as the file system would not be common between them. Can we think if this could be the reason and if yes, how we can resolve this?

rajaths010494:
> I am thinking that the tasks in the DAG could be running on different workers and hence there could be failure as the file system would not be common between them. Can we think if this could be the reason and if yes, how we can resolve this?

image
when I checked in the worker node the pem file wasn't present. So added the check if the file exists or not. 
as you mentioned if a different worker for tasks but I see only one worker node.



image

if that's the case multiple workers are present then files created in one worker should also be made available in the workers as well.

pankajkoti:
> > I am thinking that the tasks in the DAG could be running on different workers and hence there could be failure as the file system would not be common between them. Can we think if this could be the reason and if yes, how we can resolve this?
> 
> image
> 
> when I checked in the worker node the pem file wasn't present. So added the check if the file exists or not. as you mentioned if a different worker for tasks but I see only one worker node.
> image
> 
> if that's the case multiple workers are present then files created in one worker should also be made available in the workers as well.

yes, currently we have one worker so that may not be the problem. However, I believe there is no sync mechanism available which makes files created on one worker node available on other worker nodes. Even if such a sync existed, there would be delays in sync and we should not rely on a such sync with lags.

Although this change is needed and helps us in raising errors earlier, it still won't solve the problem that the DAG is failing when the copy fails. I believe we need something more to solve this problem of intermittent failures.

rajaths010494:
> > > I am thinking that the tasks in the DAG could be running on different workers and hence there could be failure as the file system would not be common between them. Can we think if this could be the reason and if yes, how we can resolve this?
> > 
> > 
> > image
> > when I checked in the worker node the pem file wasn't present. So added the check if the file exists or not. as you mentioned if a different worker for tasks but I see only one worker node.
> > image
> > if that's the case multiple workers are present then files created in one worker should also be made available in the workers as well.
> 
> yes, currently we have one worker so that may not be the problem. However, I believe there is no sync mechanism available which makes files created on one worker node available on other worker nodes. Even if such a sync existed, there would be delays in sync and we should not rely on a such sync with lags.
> 
> Although this change is needed and helps us in raising errors earlier, it still won't solve the problem that the DAG is failing when the copy fails. I believe we need something more to solve this problem of intermittent failures.

yes, we can merge the copy task into the task which uses PEM and make it one so whenever task on any worker it copies the PEM at that time and then ssh into the machine.

rajaths010494:
> > > > I am thinking that the tasks in the DAG could be running on different workers and hence there could be failure as the file system would not be common between them. Can we think if this could be the reason and if yes, how we can resolve this?
> > > 
> > > 
> > > image
> > > when I checked in the worker node the pem file wasn't present. So added the check if the file exists or not. as you mentioned if a different worker for tasks but I see only one worker node.
> > > image
> > > if that's the case multiple workers are present then files created in one worker should also be made available in the workers as well.
> > 
> > 
> > yes, currently we have one worker so that may not be the problem. However, I believe there is no sync mechanism available which makes files created on one worker node available on other worker nodes. Even if such a sync existed, there would be delays in sync and we should not rely on a such sync with lags.
> > Although this change is needed and helps us in raising errors earlier, it still won't solve the problem that the DAG is failing when the copy fails. I believe we need something more to solve this problem of intermittent failures.
> 
> yes, we can merge the copy task into the task which uses PEM and make it one so whenever task on any worker it copies the PEM at that time and then ssh into the machine.

@pankajkoti I have modified the code to download the PEM as part of the SSH task, so now PEM gets downloaded into the same worker where SSH tasks run.

pankajkoti (approved):

Issue #610, opened 2022-08-29
Astronomer Providers not displayed in Providers view in Airflow UI


josh-fell:
**Describe the bug**
In the Airflow UI, there is a Provider view under the Admin menu. This view displayed all of the Airflow provider packages installed in an environment. However, Astronomer Providers is missing from the view even when it is installed (and done so by default with our Astro Runtime image).

For example, let's start a brand new Astro project locally and verify `astronomer-providers` is installed:
<img width="676" alt="image" src="https://user-images.githubusercontent.com/48934154/187112074-ffd9cca9-95e2-4cc3-9023-dd605c8d6770.png">

Then navigate to the Providers view in the Airflow UI via Admin > Providers. Notice Astronomer Providers is absent from the list:
<img width="1053" alt="image" src="https://user-images.githubusercontent.com/48934154/187112165-09d7ae6b-2957-4141-ab98-22b65cd6d3eb.png">

**To Reproduce**
1. Start a new, local Astro project with `astro dev init`.
2. Run `astro dev start` to launch the Astro project.
3. From the command line, verify `astronomer-providers` is installed via `astro dev bash --webserver' then 'pip list | grep astronomer-providers`
4. In the Airflow UI, navigate to the Providers view via Admin > Providers menus.

**Expected behavior**
Astronomer Providers should be displayed in the Providers view within the Airflow UI.

**Screenshots**
See above.

**Desktop (please complete the following information):**
 - OS: MacOS Monterrey
 - Browser: Chrome

**Smartphone (please complete the following information):**
 N/A

**Additional context**
N/A

PR #626, opened 2022-09-07
Add provider entry_points for airflow UI

pankajastro:
This PR is an attempt to add an entry point in the provider so that Airflow can display the astronomer-provider metadata in UI

Screenshot 2022-09-08 at 3 39 47 AM

closes: https://github.com/astronomer/astronomer-providers/issues/610

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -134,6 +134,10 @@ all =
 include =
     astronomer.*
 
+[options.entry_points]
+apache_airflow_provider=
+  provider_info=astronomer.providers.package:get_provider_info
+
 [flake8]
 enable-extensions=G
 exclude = venv/*,tox/*,specs/*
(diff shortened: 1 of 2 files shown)


phanikumv (approved):
Looks good, please link to the issue that josh has raised

setup.cfg
```diff
@@ -134,6 +134,10 @@ all =
 include =
     astronomer.*
 
+[options.entry_points]
+apache_airflow_provider=
```
pankajkoti, line 138: Is the name correct or should we rename it to include astronomer?

pankajastro: I don't think https://airflow.apache.org/docs/apache-airflow-providers/index.html#creating-your-own-providers

PR #628, opened 2022-09-08
Fix redshift extractor path

pankajastro:
Extractor path should be `astronomer.providers.amazon.aws.extractors.redshift.RedshiftAsyncExtractor` instead of  `astronomer.providers.amazon.aws.extractors.redshift .RedshiftAsyncDataExtractor`

### DIFF
--- a/dev/docker-compose.yaml
+++ b/dev/docker-compose.yaml
@@ -24,7 +24,7 @@ x-airflow-common:
     OPENLINEAGE_URL: http://host.docker.internal:5050/
     OPENLINEAGE_EXTRACTORS: "astronomer.providers.google.cloud.extractors.bigquery.BigQueryAsyncExtractor;\
                             astronomer.providers.snowflake.extractors.snowflake.SnowflakeAsyncExtractor;\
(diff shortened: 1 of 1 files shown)


dev/docker-compose.yaml
```diff
@@ -24,7 +24,7 @@ x-airflow-common:
     OPENLINEAGE_URL: http://host.docker.internal:5050/
     OPENLINEAGE_EXTRACTORS: "astronomer.providers.google.cloud.extractors.bigquery.BigQueryAsyncExtractor;\
                             astronomer.providers.snowflake.extractors.snowflake.SnowflakeAsyncExtractor;\
-                            astronomer.providers.amazon.aws.extractors.redshift.RedshiftAsyncDataExtractor"
+                            astronomer.providers.amazon.aws.extractors.redshift.RedshiftAsyncExtractor"
```
phanikumv, line 27: Interesting how it was tested during development with the wrong extractor path.

cc @rajaths010494

pankajkoti (approved):
Thanks for fixing this @pankajastro

dev/docker-compose.yaml
pankajastro: I believe that is the case @pankajkoti

PR #635, opened 2022-09-12
Add fallback value for config metadata

pankajastro:
Recently we encounter that the new astro-cloud instances were failing with an error. Here, I'm adding a fallback for the metadata so that server will not crash.

```
  File "/usr/local/lib/python3.9/site-packages/astronomer/providers/package.py", line 16, in get_provider_info
    "description": config["metadata"]["description"],
  File "/usr/local/lib/python3.9/configparser.py", line 963, in __getitem__
    raise KeyError(key)
KeyError: 'metadata'
```

### DIFF
--- a/astronomer/providers/package.py
+++ b/astronomer/providers/package.py
@@ -13,8 +13,8 @@ def get_provider_info() -> Dict[str, Any]:
         # Required.
         "package-name": "astronomer-providers",
         "name": "Astronomer Providers",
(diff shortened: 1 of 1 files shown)


phanikumv (approved):

pankajkoti (commented):
Thanks for the fix @pankajastro . This helps, pods are coming up well

PR #639, opened 2022-09-14
Bump version to 1.10.0.dev1

rajaths010494:
Since 1.9.0 is released, we should bump up this version

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -27,7 +27,7 @@ project = "Astronomer Providers"
 author = "Astronomer Inc."
 
 # The full version, including alpha/beta/rc tags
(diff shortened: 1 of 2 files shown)


pankajkoti (approved):

docs/conf.py
```diff
@@ -27,7 +27,7 @@
 author = "Astronomer Inc."
 
 # The full version, including alpha/beta/rc tags
-release = "1.9.0"
+release = "1.10.0.dev1"
```
kaxil, line 30: fyi we don't need a `.` before `dev1`. We can have `1.10.0dev1` too

PR #654, opened 2022-09-20
Implement `SFTPSensorAsync`

### LINKED ISSUE #629: SFTP Async Sensor
**Describe the solution you'd like**
Async version of the SFTP sensor now that smart sensors are planned to be removed in 2.4.0.

Ability to watch a path on an SFTP server and complete when a file pattern is matched at the path.

----

I'm currently writing this for internal use at my company following the patterns used in other async sensors in this repo. I plan to contribute this once completed. The sensor is functional, and I'm currently working on the test suite for it.

tseruga:
Addresses #629 

Implements a basic, functional asynchronous SFTPSensor using asyncssh.

First time contributor here, so open to feedback on style and technical decisions around how these operators are generally formatted/created.

closes #629


tseruga:
Will address failures

bharanidharan14:
Hi @tseruga. Thanks for contributing to this project and creating this PR. We will definitely take a look at this PR and share the feedback.

astronomer/providers/sftp/triggers/sftp.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def serialize(self) -> Tuple[str, Dict[str, Any]]:
+        """Serializes SFTPTrigger arguments and classpath"""
+        return (
+            "astronomer.providers.sftp.triggers.sftp.SFTPTrigger",
+            {
+                "path": self.path,
+                "file_pattern": self.file_pattern,
+                "sftp_conn_id": self.sftp_conn_id,
+                "poll_interval": self.poll_interval,
```
bharanidharan14, line 54: "poke_interval": self.poke_interval,

tseruga:
Unsure of how to address the remaining build failures:
`build_docs` is complaining about my usage of `*` in my doc-strings. I annotated the line so the static checker locally doesn't complain about it, but I'm not sure if there's another annotation I should use for `build-docs` to accept it. I can also remove it if it's just easier to link to the `fnmatch` docs instead of having the info in the doc-string itself.

`mypy` is complaining about type hints due to class inheritance. The parent SFTP sensor class expects `Optional[SFTPHook]` but I'm overriding the hook with the new custom async hook `SFTPHookAsync`, which is a child of `SFTPHook`.

There also numerous errors from `mypy` which aren't due to my changes. Not sure what to do about those

bharanidharan14:
For the build docs failure, you need to add the `.rst` file for `sftp sensor` in ` docs/providers` (follow dbt or http sensors)
 and you need to make an entry in `docs/providers/index.rst` as well

bharanidharan14:
@tseruga For docs failure, 
You need to add `sftp` in` docs/providers/index.rst` and you need to create folder sftp as I mentioned in the earlier comment

For mypy failure, which is not related to your changes , that is covered in one of the [PR](https://github.com/astronomer/astronomer-providers/pull/659) once it is merged it will cover

tseruga:
Im clearly missing something obvious with sphinx docs. I've added an entry to the `providers/index.rst` but it still seems to be building the docs incorrectly for sftp. I've looked for an inconsistencies compared to other providers, but can't find it. I'm sure it's something small

pankajastro:
> Im clearly missing something obvious with sphinx docs. I've added an entry to the `providers/index.rst` but it still seems to be building the docs incorrectly for sftp. I've looked for an inconsistencies compared to other providers, but can't find it. I'm sure it's something small

Docs build should be working now. mypy will be fixed by https://github.com/astronomer/astronomer-providers/pull/659

pankajastro:
Any thoughts on how we will run this in our integration tests? I think we may take this in separate PR, I have created an issue to track it https://github.com/astronomer/astronomer-providers/issues/662

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -0,0 +1,127 @@
...
+    :param passphrase: passphrase used with the key_file for authentication to SFTP server
+    """
+
+    conn_name_attr = "ssh_conn_id"
+    default_conn_name = "sftp_default"
+    conn_type = "sftp"
+    hook_name = "SFTP"
+    default_known_hosts = '"~/.ssh/known_hosts"'
+
+    def __init__(  # nosec: B107
```
pankajastro, line 32: why do we need this `# nosec: B107`?

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -0,0 +1,127 @@
...
+    async def _get_conn(self) -> asyncssh.SSHClientConnection:
+        """
+        Asynchronously connect to the SFTP server as an SSH client
+
+        The following parameters are provided either in the extra json object in
+        the SFTP connection definition
+
+        key_file
+        known_hosts
+        passphrase
```
pankajastro, line 61: shall we keep these points as bullet points, how is it looking in docs currently?

astronomer/providers/sftp/sensors/sftp.py
```diff
@@ -0,0 +1,52 @@
...
+            method_name="execute_complete",
+        )
+
+    def execute_complete(self, context: Dict[str, Any], event: Optional[Dict[Any, Any]] = None) -> None:
+        """
+        Callback for when the trigger fires - returns immediately.
+        Relies on trigger to throw an exception, otherwise it assumes execution was
+        successful.
+        """
+        self.log.info("%s completed successfully.", self.task_id)
```
pankajastro, line 51: won't this print "task_id completed successfully always" ??

astronomer/providers/sftp/sensors/sftp.py
```diff
@@ -0,0 +1,52 @@
...
+        return None
```
pankajastro, line 52: this won't required

astronomer/providers/sftp/sensors/sftp.py
pankajastro: I think you should yield the status either error or success from the trigger and based on that you should either log the message or raise the exception. WDYT?

tests/sftp/hooks/test_sftp.py
```diff
@@ -0,0 +1,84 @@
...
+    def __init__(self):
+        pass
+
+    async def start_sftp_client(self):
+        return MockSFTPClient()
+
+
+@patch("astronomer.providers.sftp.hooks.sftp.SFTPHookAsync._get_conn")
+@pytest.mark.asyncio
+async def test_list_directory_path_does_not_exist(mock_hook_get_conn):
```
pankajastro, line 31: can you please convert these tests to class-based tests. you can take a look at either core or cncf tests https://github.com/astronomer/astronomer-providers/blob/main/tests/core/sensors/test_external_task.py

tests/sftp/sensors/test_sftp.py
```diff
@@ -0,0 +1,24 @@
...
+from astronomer.providers.sftp.triggers.sftp import SFTPTrigger
+
+
+def context():
+    """Creates an empty context."""
+    context = {}
+    yield context
+
+
+def test_sftp_run_now_sensor_async():
```
pankajastro, line 14: can you please convert these tests to class-based tests. you can take a look at either core or cncf tests https://github.com/astronomer/astronomer-providers/blob/main/tests/core/sensors/test_external_task.py

tests/sftp/sensors/test_sftp.py
```diff
@@ -0,0 +1,24 @@
+import pytest
+from airflow.exceptions import TaskDeferred
+
+from astronomer.providers.sftp.sensors.sftp import SFTPSensorAsync
+from astronomer.providers.sftp.triggers.sftp import SFTPTrigger
+
+
+def context():
```
pankajastro, line 8: do we need this? what about doing something like https://github.com/astronomer/astronomer-providers/blob/main/tests/core/sensors/test_external_task.py#L20

tests/sftp/triggers/test_sftp.py
```diff
@@ -0,0 +1,16 @@
+from astronomer.providers.sftp.triggers.sftp import SFTPTrigger
+
+
+def test_sftp_trigger_serialization():
```
pankajastro, line 4: can you please convert these tests to class-based tests. you can take a look at either core or cncf tests https://github.com/astronomer/astronomer-providers/blob/main/tests/core/sensors/test_external_task.py

astronomer/providers/sftp/triggers/sftp.py
```diff
@@ -0,0 +1,63 @@
...
+            },
+        )
+
+    async def run(self) -> AsyncIterator["TriggerEvent"]:  # type: ignore[override]
+        """
+        Makes a series of asynchronous calls to sftp servers via async sftp hook. It yields a Trigger if
+        file matching file pattern exists at the specified path, otherwise it throws an exception.
+        """
+        hook = self._get_async_hook()
+        while True:
```
pankajastro, line 55: does it make sense to expose a timeout param in operator WDYT?

astronomer/providers/sftp/hooks/sftp.py
tseruga: Static checker thinks that a password is exposed with:
```
password: str = "",
```

astronomer/providers/sftp/triggers/sftp.py
tseruga: Not sure I understand. We do expose the timeout parameter. Are you implying that we shouldn't?

pankajastro: yes, in `SFTPSensorAsync` class `__init__ ` method

tseruga: Oh just explicitly accept timeout instead of hiding it behind the parent constructor? Can do

astronomer/providers/sftp/triggers/sftp.py
```diff
@@ -0,0 +1,63 @@
...
+            try:
+                await hook.get_file_by_pattern(path=self.path, fnmatch_pattern=self.file_pattern)
+                yield TriggerEvent(True)
+            except AirflowException:
+                await asyncio.sleep(self.poke_interval)
```
pankajastro, line 60: Also, I think if a file is not found then only we should sleep in case of any other error we should throw error. WDYT?

tseruga: This is how it works now, albeit in a roundabout way.

In the hook
```
        try:
            files = await sftp_client.listdir(path)
            return sorted(files)
        except asyncssh.SFTPNoSuchFile:
            raise AirflowException(f"No files at path {path} found — Deferring")
```

we only catch the exception when a file isn't found. This means that any connection errors will fall through and halt the sensor, throwing the raw exception back to Airflow (which isn't then caught, because it's not an `AirflowException`)

pankajastro: Ok, In that case, any other exception than `AirflowException` we should yield the error something like
```
yield TriggerEvent({"status": "error", message: e})
```
then if `status` is an error then raise an exception in SFTPSensorAsync  `execute_complete` method you can access it like
```
if event["status"] == "error":
    raise
```

tseruga: Got it, this makes more sense. Going to rework this a bit so that `AirflowExceptions` aren't being used as heavily for control flow. It was causing the triggerer to get stuck in a loop after the task had finished/failed execution.

pankajastro:
Please Update/rebase your branch.

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -0,0 +1,127 @@
...
+                    conn_config.update(known_hosts=self.known_hosts)
+
+            if self.passphrase:
+                conn_config.update(passphrase=self.passphrase)
+
+            ssh_client = await asyncssh.connect(**conn_config)
+
+            return ssh_client
+
+    async def list_directory(self, path: str = "") -> List[str]:
```
rajaths010494, line 101:
```suggestion
    async def list_directory(self, path: str = "") -> List[str]:
```
async def list_directory(self, path: str = "") -> Optional[List[str]]:

astronomer/providers/sftp/sensors/sftp.py
```diff
@@ -0,0 +1,71 @@
...
+            trigger=SFTPTrigger(
+                path=self.path,
+                file_pattern=self.file_pattern,
+                sftp_conn_id=self.sftp_conn_id,
+                poke_interval=self.poke_interval,
+            ),
+            method_name="execute_complete",
+        )
+
+    def execute_complete(self, context: Dict[str, Any], event: Optional[Dict[Any, Any]] = None) -> None:
```
pankajastro, line 57: mypy is still failing maybe you can change 
```
def execute_complete(self, context: Dict[str, Any], event: Optional[Dict[Any, Any]] = None) -> None:
```
to 
```
def execute_complete(self, context: Dict[str, Any], event: Any = None) -> None:
```

astronomer/providers/sftp/hooks/sftp.py
rajaths010494, line 101:
```suggestion
    async def list_directory(self, path: str = "") -> Optional[List[str]]:
```

astronomer/providers/sftp/sensors/sftp.py
pankajastro, line 57:
```suggestion
    def execute_complete(self, context: Dict[str, Any], event: Any = None) -> None:
```

docs/providers/index.rst
```diff
@@ -15,15 +15,17 @@ Async Provider Example DAG's
...
     Microsoft <microsoft/index>
     Snowflake <snowflake/index>
+    SFTP <sftp/index>
 
 
 * `Amazon AWS <amazon/index.html>`_
 * `Apache <apache/index.html>`_
 * `CNCF <cncf/index.html>`_
 * `Core <core/index.html>`_
-* `Databricks <databricks/index.html>`_
```
pankajastro, line 24: Why these change?

docs/providers/index.rst
```diff
@@ -15,15 +15,17 @@ Async Provider Example DAG's
...
-* `HTTP <http/index.html>`_
-* `Microsoft <microsoft/index.html>`_
-* `Snowflake <snowflake/index.html>`_
+* `Databricks <databricks/index>`_
+* `dbt <dbt/index>`_
+* `Google <google/index>`_
+* `HTTP <http/index>`_
+* `Microsoft <microsoft/index>`_
+* `Snowflake <snowflake/index>`_
+* `SFTP <sftp/index>`_
```
pankajastro, line 31:
```suggestion
* `SFTP <sftp/index.html>`_
```

tseruga:
Ill look through the missed lines and add some tests

tseruga:
Surprised 97.4% coverage still fails the check 😆 I can cover the missing lines

pankajastro:
Thanks @tseruga for addressing the comments. Looking great 🚀

pankajastro (approved):

bharanidharan14 (approved):

rajaths010494:
@tseruga  can you rebase so we can merge it?

pankajastro:
This is awesome, thanks @tseruga  👏 for creating the PR. Appreciate it.

rajaths010494:
Awesome work @tseruga , congrats on your first merged pull request!

tseruga:
Thanks @pankajastro @rajaths010494 @bharanidharan14 !

Hopefully will be contributing more in the future!

PR #671, opened 2022-09-28
Handel the Databricks Operator failure with proper error message

bharanidharan14:
Handled the failure case properly as per in sync version
- In the Jobs API 2.1, we can't call `get_run_output` on the top-level Run ID because it's
not supported by API - we need to call this function on a specific sub-run of the job, even
if it consists of the single task


bharanidharan14:
Got the code unrelated to my changes in this PR, So closing this PR raising another PR for my changes

Issue #643, opened 2022-09-15
Implement `DatabricksOperatorAsync`  changes made in OSS `DatabricksOperator`


bharanidharan14:
- Implement the OSS `DatabricksOperator` changes in `DatabricksOperatorAsync`

PR #674, opened 2022-09-29
Handle the Databricks Operator failures

bharanidharan14:
Handled the failure case properly with proper failure message as per in sync version
- In the Jobs API 2.1, we can't call `get_run_output` on the top-level Run ID because it's
not supported by API - we need to call this function on a specific sub-run of the job, even
if it consists of the single task

closes: #643

### DIFF
--- a/astronomer/providers/databricks/hooks/databricks.py
+++ b/astronomer/providers/databricks/hooks/databricks.py
@@ -8,6 +8,7 @@ from airflow import __version__
 from airflow.exceptions import AirflowException
 from airflow.providers.databricks.hooks.databricks import (
     GET_RUN_ENDPOINT,
+    OUTPUT_RUNS_JOB_ENDPOINT,
     DatabricksHook,
     RunState,
 )
@@ -39,8 +40,7 @@ class DatabricksHookAsync(DatabricksHook):
         :param run_id: id of the run
         :return: state of the run
         """
-        json = {"run_id": run_id}
-        response = await self._do_api_call_async(GET_RUN_ENDPOINT, json)
+        response = await self.get_run_response(run_id)
         state = response["state"]
         life_cycle_state = state["life_cycle_state"]
         # result_state may not be in the state if not terminal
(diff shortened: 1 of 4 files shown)


bharanidharan14:
@rajaths010494 @pankajastro @pankajkoti @utkarsharma2 @sunank200, Please review this PR

astronomer/providers/databricks/triggers/databricks.py
```diff
@@ -82,8 +84,30 @@ async def run(self) -> AsyncIterator["TriggerEvent"]:  # type: ignore[override]
...
+                            error_message = (
+                                f"{self.task_id} failed with terminal state: {run_state} "
+                                f"and with the error {notebook_error}"
+                            )
+                        else:
+                            error_message = (
+                                f"{self.task_id} failed with terminal state: {run_state} "
+                                f"and with the error {run_state.state_message}"
+                            )
+                        yield TriggerEvent({"status": "error", "message": error_message})
```
kaxil, line 110: I don't like so deep nesting, can we simplify this

bharanidharan14: ok will simplify this

bharanidharan14: Fixed it

astronomer/providers/databricks/hooks/databricks.py
```diff
@@ -50,6 +50,26 @@ async def get_run_state_async(self, run_id: str) -> RunState:
 
         return RunState(life_cycle_state, result_state, state_message)
 
+    async def get_run_response(self, run_id: str) -> Dict[str, Any]:
```
rajaths010494, line 53: `get_run_response` and `get_run output response` do the same task if we even send the endpoint as an argument. WDYT?

bharanidharan14: yes, we can do that

rajaths010494 (approved):

Issue #677, opened 2022-09-29
Modify test RC dependency automation to create a PR


pankajkoti:
Currently, we only modify the `setup.cfg` with the RC providers, build the image, deploy it to Astro Cloud and then trigger the master DAG run. However, we do not check if the RC cause any static check failures like `mypy` or if the tests run fine as we do not run the checks which are only triggered once we create a PR. We come to know of such failures only upon new subsequent PRs. So to detect such failures early, change the `test-rc-deps` make target to create a PR by pushing the `setup.cfg` to a new branch.

PR #679, opened 2022-09-29
Modify test RC dependency automation to create a PR

pankajkoti:
Currently, we only modify the setup.cfg with the RC providers, build the image, deploy it to Astro Cloud and then trigger the master DAG run. However, we do not check if the RC causes any static check failures like `mypy` or if the tests run fine as we do not run the checks which are only triggered once we create a PR. We come to know of such failures only upon new subsequent PRs. So to detect such failures early, change the `test-rc-deps `make target to create a PR by pushing the setup.cfg to a new branch.

Also, we observed in the past that it takes sometimes ~20 mins for the new image changes to be applied on Astro Cloud and hence we are increasing the wait time to 30 mins before triggering the master DAG run to test the RCs against our DAGs.

closes: #677

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,15 +74,25 @@ run-local-lineage-server: ## Run flask based local Lineage server
 	FLASK_APP=dev/local_flask_lineage_server.py flask run --host 0.0.0.0 --port 5050
 
 test-rc-deps: ## Test providers RC by building an image with given dependencies and running the master DAG
+	git checkout main && git pull origin main
+	$(eval current_timestamp := $(shell date +%Y-%m-%dT%H-%M-%S%Z))
+	echo "Current timestamp is" $(current_timestamp)
+	$(eval branch_name := "rc-test-$(current_timestamp)")
+	git checkout -b $(branch_name)
+	echo "Updating setup.cfg with RC provider packages"
 	python3 dev/scripts/replace_dependencies.py '$(RC_PROVIDER_PACKAGES)'
+	echo "Building and deploying image to Astro Cloud"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -74,15 +74,25 @@ run-local-lineage-server: ## Run flask based local Lineage server
...
-	echo "Sleeping for 900 seconds (15 mins) allowing the deployed image to be updated across all Airflow components.."
-	sleep 900
+	echo "Sleeping for 1800 seconds (30 minutes) allowing the deployed image to be updated across all Airflow components.."
+	sleep 1800
 	python3 dev/scripts/trigger_master_dag.py '$(DEPLOYMENT_ID)' '$(ASTRONOMER_KEY_ID)' ' $(ASTRONOMER_KEY_SECRET)'
-	git checkout setup.cfg
+	git add setup.cfg
+	git commit -m "Update setup.cfg to use RC provider packages"
+	git push origin $(branch_name)
+	gh pr create --base "main" --title "Test RC provider packages" --fill
```
kaxil, line 95: Some devs might not have `gh` installed

pankajkoti: Yes, how should we tackle it? It should error out that command not found and then the developer can install `gh`. Or is there some proactive way you mean?

pankajkoti: I am explicitly making a pre-check and erroring out if the command is not installed. Does it sound okay?

kaxil: Yeah error out on install "gh" and add a link

kaxil (approved):

PR #681, opened 2022-09-29
Custom XCom GCS

### LINKED ISSUE #596: Add custom XCOM backends - GCS
Refer to https://astronomer.slack.com/archives/CGQSYG25V/p1658957159758039 for the relevant discussion,
and add GCS XCOM Backend into the Google Provider

rajaths010494:
Custom XCOM to upload xcom values to the GCS bucket for the given name.
Adding ENV ` AIRFLOW__CORE__XCOM_BACKEND: astronomer.providers.google.cloud.xcom_backends.gcs.GCSXComBackend` in the `docker-compose.yaml` or where the airflow is running will enable the custom XCOM for GCS.

closes #596

### DIFF
--- a/astronomer/providers/google/cloud/xcom_backends/gcs.py
+++ b/astronomer/providers/google/cloud/xcom_backends/gcs.py
@@ -21,6 +21,7 @@ class GCSXComBackend(BaseXCom):
     PREFIX = os.getenv("PREFIX", "gcs_xcom_")
     GCP_CONN_ID = os.getenv("CONNECTION_NAME", "google_cloud_default")
     BUCKET_NAME = os.getenv("XCOM_BACKEND_BUCKET_NAME", "airflow_xcom_backend_default_bucket")
+    PANDAS_DATAFRAME = "dataframe"
 
     @staticmethod
     def write_and_upload_value(value: Any) -> str:
@@ -29,17 +30,16 @@ class GCSXComBackend(BaseXCom):
         hook = GCSHook(gcp_conn_id=GCSXComBackend.GCP_CONN_ID)
         if conf.getboolean("core", "enable_xcom_pickling"):
             value = pickle.dumps(value)
-        elif isinstance(value, list):
-            value = str(value)
         elif isinstance(value, pd.DataFrame):
             value = value.to_json()
+            key_str = key_str + "_" + GCSXComBackend.PANDAS_DATAFRAME
         else:
             value = json.dumps(value)
         hook.upload(GCSXComBackend.BUCKET_NAME, key_str, data=value)
         return key_str
 
     @staticmethod
-    def read_value(filename: str) -> str:
+    def read_value(filename: str) -> bytes:
         """Download the file from GCS"""
         # Here we download the file from GCS
         hook = GCSHook(gcp_conn_id=GCSXComBackend.GCP_CONN_ID)
@@ -59,11 +59,10 @@ class GCSXComBackend(BaseXCom):
         if isinstance(result, str) and result.startswith(GCSXComBackend.PREFIX):
             value = GCSXComBackend.read_value(result)
             if conf.getboolean("core", "enable_xcom_pickling"):
-                try:
-                    return pickle.loads(value)  # nosec
-                except pickle.UnpicklingError:
-                    return json.loads(value.decode("UTF-8"))
-            return value
+                return pickle.loads(value)  # nosec
+            elif result.endswith(GCSXComBackend.PANDAS_DATAFRAME):
+                return pd.read_json(value)
+            return json.loads(value)
 
     def orm_deserialize_value(self) -> str:
         """
--- a/astronomer/providers/google/cloud/xcom_backends/gcs.py
+++ b/astronomer/providers/google/cloud/xcom_backends/gcs.py
@@ -21,7 +21,7 @@ class GCSXComBackend(BaseXCom):
     @staticmethod
     def serialize_value(value: Any) -> Any:  # type: ignore[override]
         """Custom XCOM for GCS to serialize the data"""
-        value = _GCSXComBackend().write_and_upload_value(value)
+        value = _GCSXComBackend.write_and_upload_value(value)
         return BaseXCom.serialize_value(value)
 
     @staticmethod
--- a/astronomer/providers/google/cloud/xcom_backends/gcs.py
+++ b/astronomer/providers/google/cloud/xcom_backends/gcs.py
@@ -13,6 +13,36 @@ from airflow.providers.google.cloud.hooks.gcs import GCSHook
 
 class GCSXComBackend(BaseXCom):
     """
+    The GCS custom xcom backend is an xcom custom backend wrapper that handles serialization and deserialization
+    of common data types. This overrides the ``TaskInstance.XCom`` object with this wrapper.
+    """
+
+    @staticmethod
+    def serialize_value(value: Any) -> Any:  # type: ignore[override]
+        """Custom XCOM for GCS to serialize the data"""
+        value = _GCSXComBackend().write_and_upload_value(value)
+        return BaseXCom.serialize_value(value)
+
+    @staticmethod
+    def deserialize_value(result: Any) -> Any:
+        """Custom XCOM for GCS to deserialize the data"""
+        result = BaseXCom.deserialize_value(result)
+        if isinstance(result, str) and result.startswith(_GCSXComBackend.PREFIX):
+            result = _GCSXComBackend.download_and_read_value(result)
+        return result
+
+    def orm_deserialize_value(self) -> str:
+        """
+        Deserialize amethod which is used to reconstruct ORM XCom object.
+        This method should be overridden in custom XCom backends to avoid
+        unnecessary request or other resource consuming operations when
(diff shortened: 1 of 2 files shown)


astronomer/providers/google/cloud/xcom_backends/gcs.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @staticmethod
+    def write_and_upload_value(value: Any) -> str:
+        """Convert to string and upload to GCS"""
+        key_str = GCSXComBackend.PREFIX + str(uuid.uuid4())
+        hook = GCSHook(gcp_conn_id=GCSXComBackend.GCP_CONN_ID)
+        if conf.getboolean("core", "enable_xcom_pickling"):
+            value = pickle.dumps(value)
+        elif isinstance(value, list):
+            value = str(value)
```
feluelle, line 33: I think it should be possible to use `json.dumps` for this as well as a list is a valid json.
```suggestion
```

astronomer/providers/google/cloud/xcom_backends/gcs.py
```diff
@@ -0,0 +1,75 @@
...
+    def deserialize_value(result: Any) -> Any:
+        """Custom XCOM for GCS to deserialize the data"""
+        result = BaseXCom.deserialize_value(result)
+        if isinstance(result, str) and result.startswith(GCSXComBackend.PREFIX):
+            value = GCSXComBackend.read_value(result)
+            if conf.getboolean("core", "enable_xcom_pickling"):
+                try:
+                    return pickle.loads(value)  # nosec
+                except pickle.UnpicklingError:
+                    return json.loads(value.decode("UTF-8"))
```
feluelle, line 65: What happens in case of `pd.Dataframe`? I think you should be using `value.from_json()` to return the data frame.

astronomer/providers/google/cloud/xcom_backends/gcs.py
rajaths010494: removed and use `json.dumps`

astronomer/providers/google/cloud/xcom_backends/gcs.py
rajaths010494: yes added the type to key string so i can deserialise to the same datatype

astronomer/providers/google/cloud/xcom_backends/gcs.py
```diff
@@ -0,0 +1,74 @@
...
+class GCSXComBackend(BaseXCom):
+    """
+    Custom XCom persistence class extends base to support various datatypes.
+    To use this XCom Backend, add the environment variable `AIRFLOW__CORE__XCOM_BACKEND`
+    to your environment and set it to
+    `astronomer.providers.google.cloud.xcom_backends.gcs.GCSXComBackend`
+    """
+
+    PREFIX = os.getenv("PREFIX", "gcs_xcom_")
+    GCP_CONN_ID = os.getenv("CONNECTION_NAME", "google_cloud_default")
```
kaxil, line 22: `CONNECTION_NAME` is a very generic name for an env var

astronomer/providers/google/cloud/xcom_backends/gcs.py
```diff
@@ -0,0 +1,74 @@
...
+
+    @staticmethod
+    def write_and_upload_value(value: Any) -> str:
+        """Convert to string and upload to GCS"""
+        key_str = GCSXComBackend.PREFIX + str(uuid.uuid4())
+        hook = GCSHook(gcp_conn_id=GCSXComBackend.GCP_CONN_ID)
+        if conf.getboolean("core", "enable_xcom_pickling"):
+            value = pickle.dumps(value)
+        elif isinstance(value, pd.DataFrame):
+            value = value.to_json()
```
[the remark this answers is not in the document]
rajaths010494: `to_json`can take read bytes and string, whereas if  `to_csv` is used and while deserialising we use `read_csv` needs an io.StringIO(value) and `to_feather` needs a filepath to serialise and deserialise

astronomer/providers/google/cloud/xcom_backends/gcs.py
rajaths010494: Fixed

astronomer/providers/google/cloud/xcom_backends/gcs.py
```diff
@@ -0,0 +1,81 @@
...
+    PREFIX = os.getenv("XCOM_BACKEND_PREFIX", "gcs_xcom_")
+    GCP_CONN_ID = os.getenv("XCOM_BACKEND_CONNECTION_NAME", "google_cloud_default")
+    BUCKET_NAME = os.getenv("XCOM_BACKEND_BUCKET_NAME", "airflow_xcom_backend_default_bucket")
+    PANDAS_DATAFRAME = "dataframe"
+    DATETIME_OBJECT = "datetime"
+
+    @staticmethod
+    def write_and_upload_value(value: Any) -> str:
+        """Convert to string and upload to GCS"""
+        key_str = GCSXComBackend.PREFIX + str(uuid.uuid4())
```
uranusjr, line 31:
```suggestion
        key_str = f"{GCSXComBackend.PREFIX}{uuid.uuid4()}"
```

Same for other `+` lines below. This is more performant.

uranusjr (commented):
I would restructure the implementation like this:

```python
# Only implements the required interface and "forward" logic.
class GCSXCom(BaseXCom):
    def orm_deserialize_value(self) -> str:
        ...

    @staticmethod
    def serialize_value(value: Any, **kwargs) -> Any:
        value = _GCSXComBackend().write_and_upload_value(value, **kwargs)
        return BaseXCom.serialize_value(value)

    @staticmethod
    def deserialize_value(xcom: "XCom") -> Any:
        result = BaseXCom.deserialize_value(xcom)
        if isinstance(result, str) and result.startswith(GCSXComBackend.PREFIX):
            result = _GCSXComBackend().download_and_read_value(result)
        return result

# Implements the actual logic!
class _GCSXComBackend:
    def write_and_upload_value(self, value, **kwargs):
        ...

    def download_and_read_value(self, result):
        ...
```

The advantage to this is we can reference values with `self`, which enables us to cache things (for example the GCSHook) and make the environment variable retrievals lazy properties. This reduces import-time logic (makes startup fast) and can also simplify testing.

rajaths010494:
> I would restructure the implementation like this:
> 
> ```python
> # Only implements the required interface and "forward" logic.
> class GCSXCom(BaseXCom):
>     def orm_deserialize_value(self) -> str:
>         ...
> 
>     @staticmethod
>     def serialize_value(value: Any, **kwargs) -> Any:
>         value = _GCSXComBackend().write_and_upload_value(value, **kwargs)
>         return BaseXCom.serialize_value(value)
> 
>     @staticmethod
>     def deserialize_value(xcom: "XCom") -> Any:
>         result = BaseXCom.deserialize_value(xcom)
>         if isinstance(result, str) and result.startswith(GCSXComBackend.PREFIX):
>             result = _GCSXComBackend().download_and_read_value(result)
>         return result
> 
> # Implements the actual logic!
> class _GCSXComBackend:
>     def write_and_upload_value(self, value, **kwargs):
>         ...
> 
>     def download_and_read_value(self, result):
>         ...
> ```
> 
> The advantage to this is we can reference values with `self`, which enables us to cache things (for example the GCSHook) and make the environment variable retrievals lazy properties. This reduces import-time logic (makes startup fast) and can also simplify testing.

Yes, i am doing it. Will need to modify test as well for the changes

bharanidharan14:
@kaxil @uranusjr @josh-fell Need your review on this PR

astronomer/providers/amazon/aws/example_dags/example_sagemaker.py
```diff
@@ -34,18 +41,41 @@
...
+                    7.0,3.2,4.7,1.4
+                    6.4,3.2,4.5,1.5
+                    """
+TRAIN_DATASET = """
+                2,4.9,3.0,1.4,0.2
+                0,7.3,2.9,6.3,1.8
+                2,5.1,3.5,1.4,0.2
+                0,4.9,2.5,4.5,1.7
+                """
+
```
kaxil, line 54: Unrelated changes?

rajaths010494: Rebased branch

astronomer/providers/google/cloud/xcom_backends/gcs.py
```diff
@@ -0,0 +1,97 @@
...
+class GCSXComBackend(BaseXCom):
+    """
+    The GCS custom xcom backend is an xcom custom backend wrapper that handles serialization and deserialization
+    of common data types. This overrides the ``TaskInstance.XCom`` object with this wrapper.
+    """
+
+    @staticmethod
+    def serialize_value(value: Any) -> Any:  # type: ignore[override]
+        """Custom XCOM for GCS to serialize the data"""
+        value = _GCSXComBackend().write_and_upload_value(value)
```
uranusjr, line 24:
```suggestion
        value = _GCSXComBackend.write_and_upload_value(value)
```

kaxil (approved):

PR #716, opened 2022-10-18
POC Big Query polling ping-pong

uranusjr:
Instead of running a loop in the trigger, this essentially "uses" the triggerer process as the loop. The idea is:

1. The operator gets deferred after submitting the query.
2. The trigger waits and checks the result *once* and reports via an event
3. The operator is resumed by the event to check the status
    3a. If the job is reported as still running, defer to run the trigger *again*. Go back to 2.
    3b. If the job is done (success or error), return or raise.

This kind of achieves the ping-pong effect and emit the relevant logs in the operator's worker process, instead of in the triggerer process. The repeated triggering would probably have some overhead, but should be OK since everything is async.

### DIFF
--- a/astronomer/providers/google/cloud/operators/bigquery.py
+++ b/astronomer/providers/google/cloud/operators/bigquery.py
@@ -1,5 +1,5 @@
 """This module contains Google BigQueryAsync providers."""
-from typing import Any, Dict
+from typing import Any, Dict, Optional, Type
 
 from airflow.exceptions import AirflowException
 from airflow.models.baseoperator import BaseOperator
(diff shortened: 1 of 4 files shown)


uranusjr:
Test failures are expected for now; we need to change all the BQ operators after (if) we decide this is a good idea and rewrite the test.

phanikumv:
> Test failures are expected for now; we need to change all the BQ operators after (if) we decide this is a good idea and rewrite the test.

@uranusjr Just thinking if we can come up with a solution that would work across most of the operators rather than specific ones?

pankajastro:
> ```python
> AsyncProgressTrigger
> ```

@uranusjr excited to see what the template (methods, object interaction etc) for `AsyncProgressTrigger` and `AsyncProgressMixin` would look like. We will have to return/yield from trigger to operator execute_complete method display log and re-defer if needed.

uranusjr:
I pushed an experiemental implementation to extract the ping-pong logic into common base classes. The implementation is only used in `BigQueryInsertJobOperatorAsync` for the moment.

phanikumv:
Closing as this was a POC for the ping-pong functionality.Will use it as a reference for future implementations.

Issue #725, opened 2022-10-25
Token error with Sagemaker Async Operators


fhoda:
**Describe the bug**
Getting errors with the new Sagemaker Async Operators that I don't get with the traditional ones. I'm using a personal Access Key, Secret, and Session Token as I did with the non async operators for auth.

```
botocore.exceptions.ClientError: An error occurred (UnrecognizedClientException) when calling the DescribeTrainingJob operation: The security token included in the request is invalid.
```

**To Reproduce**
Use the SageMaker async operators with user Access Key, Secret, and Session Token

**Expected behavior**
Expect it to not have auth/token errors.

**Additional context**
When I switch back to the traditional operators in the same dag with the same auth creds it works fine.

@kentdanas also had similar issues and her auth was setup a little different.

bharanidharan14:
The issue is with the session token is not considered while the secrete and access key is given in the connection proper field, not in the extra config

PR #730, opened 2022-10-27
Fix session token from extra config

bharanidharan14:
when AWS secrete and access key is passed via the connection UI field the session token is not considered, when it is passed via extra config it is considered

closes: #725

### DIFF
--- a/astronomer/providers/amazon/aws/hooks/base_aws.py
+++ b/astronomer/providers/amazon/aws/hooks/base_aws.py
@@ -65,6 +65,12 @@ class AwsBaseHookAsync(AwsBaseHook):
             self.log.info("Retrieving region_name from Connection.extra_config['region_name']")
             region_name = extra_config["region_name"]  # pragma: no cover
 
+        if "aws_session_token" in extra_config:
+            self.log.info(
+                "session token retrieved from extra, please note you are responsible for renewing these.",
+            )  # pragma: no cover
+            aws_session_token = extra_config.get("aws_session_token")
+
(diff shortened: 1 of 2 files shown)


bharanidharan14:
@rajaths010494 @pankajastro @phanikumv can you review this PR

rajaths010494 (commented):
Overall LGTM.

astronomer/providers/amazon/aws/hooks/base_aws.py
```diff
@@ -65,6 +65,12 @@ async def get_client_async(self) -> AioBaseClient:
             self.log.info("Retrieving region_name from Connection.extra_config['region_name']")
             region_name = extra_config["region_name"]  # pragma: no cover
 
+        if "aws_session_token" in extra_config:
+            self.log.info(
```
pankajastro, line 69: in the sync version, they also log `self.conn_repr` https://github.com/apache/airflow/blob/main/airflow/providers/amazon/aws/utils/connection_wrapper.py#L349 
would it be possible to do similar logging here?

bharanidharan14: whatever the previous async log was there I followed the same. If we want to follow the same in sync version I need to change all the log in this file

pankajastro: As long we have a GitHub issue for it I'm ok.

pankajastro (approved):
LGTM

Issue #736, opened 2022-10-27
XCom Output of Sagemaker Async Operators


fhoda:
**Describe the bug**
XCom return value of `SageMakerTransformOperatorAsync`  and `SageMakerTrainingOperatorAsync` does not produce the expected output.

It seems like some key(s) don't match the non-async operator output.

**To Reproduce**
1. Run a dag with traditional operators
2. Run same dag with Async operators
3. Compare outputs

**Expected behavior**
The Xcom keys and values should match whatever the traditional non-async version of the operators output.

**Screenshots**

fhoda:
@bharanidharan14  I tested locally and seems to be fixed. Thank you!

PR #737, opened 2022-10-27
Fix response key's for SageMaker Training and Transform operator

bharanidharan14:
Fixed the response key for SageMaker Training and Transform Operator

closes: #736

### DIFF
--- a/astronomer/providers/amazon/aws/example_dags/example_sagemaker.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_sagemaker.py
@@ -7,6 +7,7 @@ from datetime import datetime
 from airflow import DAG
 from airflow.decorators import task
 from airflow.models.baseoperator import chain
+from airflow.operators.bash import BashOperator
 from airflow.operators.python import get_current_context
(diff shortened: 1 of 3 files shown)


phanikumv (changes requested):
The title and description do not indicate what and why we are doing this change

rajaths010494 (approved):
LGTM . Just a question, how frequently do we want to update the session tokens with the new connection setup?

bharanidharan14:
> LGTM . Just a question, how frequently do we want to update the session tokens with the new connection setup?

Currently, We are facing some issues on the session token which is causing some steps to fail in the integration, sill I am looking into it.

bharanidharan14:
> > LGTM . Just a question, how frequently do we want to update the session tokens with the new connection setup?
> 
> Currently, We are facing some issues on the session token which is causing some steps to fail in the integration, sill I am looking into it.

I have fixed the issue and created the session token with credentials in this PR #739

bharanidharan14:
@phanikumv @rajaths010494 @pankajastro can you guys review this PR

phanikumv (approved):

PR #742, opened 2022-11-01
Typecast bigquery job response col value

pankajastro:
closes: https://github.com/astronomer/astronomer-providers/issues/743
The Bigquery hook get_records always return a list of string irrespective of the Bigquery table column type. So if even my table column has value 0 the task succeeds since bool("0") returns True.

This PR fixes it by typecasting the column value using `_bq_cast`

### DIFF
--- a/astronomer/providers/google/cloud/hooks/bigquery.py
+++ b/astronomer/providers/google/cloud/hooks/bigquery.py
@@ -2,7 +2,7 @@ from typing import Any, Dict, List, Optional, Union, cast
 
 from aiohttp import ClientSession as ClientSession
 from airflow.exceptions import AirflowException
-from airflow.providers.google.cloud.hooks.bigquery import BigQueryHook
+from airflow.providers.google.cloud.hooks.bigquery import BigQueryHook, _bq_cast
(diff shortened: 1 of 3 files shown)


astronomer/providers/google/cloud/hooks/bigquery.py
```diff
@@ -2,7 +2,7 @@
 
 from aiohttp import ClientSession as ClientSession
 from airflow.exceptions import AirflowException
-from airflow.providers.google.cloud.hooks.bigquery import BigQueryHook
+from airflow.providers.google.cloud.hooks.bigquery import BigQueryHook, _bq_cast
```
pankajastro, line 5: importing private method. Please let me know if it is ok here

kaxil (approved):

rajaths010494 (approved):

astronomer/providers/google/cloud/hooks/bigquery.py
pankajastro: @kaxil https://github.com/apache/airflow/pull/27543

PR #746, opened 2022-11-02
Fix `redshift_data` hook to consider access and secret key from the connections object not from extra

bharanidharan14:
- Consider access and secret key if it is passed in connection login and password field
- Consider AWS session token if it is passed in `extra_config`

closes: #741

### DIFF
--- a/astronomer/providers/amazon/aws/hooks/redshift_data.py
+++ b/astronomer/providers/amazon/aws/hooks/redshift_data.py
@@ -77,6 +77,9 @@ class RedshiftDataHook(AwsBaseHook):
                 if "secret_access_key" in extra_config
                 else extra_config["aws_secret_access_key"]
             )
+        elif connection_object.login:
+            conn_params["aws_access_key_id"] = connection_object.login
+            conn_params["aws_secret_access_key"] = connection_object.password
         else:
             raise AirflowException("Required access_key_id, aws_secret_access_key")
 
--- a/astronomer/providers/amazon/aws/hooks/redshift_data.py
+++ b/astronomer/providers/amazon/aws/hooks/redshift_data.py
@@ -95,7 +95,7 @@ class RedshiftDataHook(AwsBaseHook):
             self.log.info(
                 "session token retrieved from extra, please note you are responsible for renewing these.",
             )
-            conn_params["aws_session_token"] = extra_config.get("aws_session_token")
+            conn_params["aws_session_token"] = extra_config["aws_session_token"]
 
         if "cluster_identifier" in extra_config:
             self.log.info("Retrieving cluster_identifier from Connection.extra_config['cluster_identifier']")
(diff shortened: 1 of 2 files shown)


astronomer/providers/amazon/aws/hooks/redshift_data.py
```diff
@@ -77,6 +77,9 @@ def get_conn_params(self) -> Dict[str, Union[str, int]]:
                 if "secret_access_key" in extra_config
                 else extra_config["aws_secret_access_key"]
             )
+        elif connection_object.login:
+            conn_params["aws_access_key_id"] = connection_object.login
+            conn_params["aws_secret_access_key"] = connection_object.password
```
pankajastro, line 82: between this method andhttps://github.com/astronomer/astronomer-providers/blob/main/astronomer/providers/amazon/aws/hooks/base_aws.py#L33 there is lot of redundant code we should see if we can improve it

tests/amazon/aws/hooks/test_redshift_data.py
```diff
@@ -248,9 +248,11 @@ def test_execute_query_exception(mock_conn, mock_params):
 
 
 @pytest.mark.parametrize(
-    "connection_details, expected_output",
+    "mock_login, mock_pwd, connection_details, expected_output",
     [
         (
+            "",
```
pankajastro, line 254: we should add some value empty does not make sense

tests/amazon/aws/hooks/test_redshift_data.py
```diff
@@ -269,6 +271,8 @@ def test_execute_query_exception(mock_conn, mock_params):
             },
         ),
         (
+            "",
```
pankajastro, line 274: same here

tests/amazon/aws/hooks/test_redshift_data.py
```diff
@@ -286,11 +290,31 @@ def test_execute_query_exception(mock_conn, mock_params):
                 "database": "",
             },
         ),
+        (
+            "test",
+            "test",
+            {
+                "db_user": "test_user",
+                "cluster_identifier": "",
+                "region": "",
```
pankajastro, line 299: same here

astronomer/providers/amazon/aws/hooks/redshift_data.py
bharanidharan14: Yes that's async hook which gets the connection info this `redshift_data` hook  class

bharanidharan14: We can fix this redundant code

bharanidharan14: created ticket https://app.zenhub.com/workspaces/airflow-providers-team-61a89ee7fbf048001dc8378d/issues/astronomer/astronomer-providers/747

bharanidharan14:
@pankajastro @rajaths010494 Addressed your review comments, can you please take a look at it

pankajkoti (approved):

rajaths010494 (approved):

Issue #714, opened 2022-10-17
Operator XCOM - SnowflakeOperator


fritz-astronomer:
**Is your feature request related to a problem? Please describe.**
to keep parity with the sync version - the async operators should return an XCom of results where the synchronous version would return an XCom

PR #752, opened 2022-11-11
Fix `SnowflakeOperatorAsync` Xcom value to return

bharanidharan14:
Return an XCom of results in `SnowflakeOperatorAsync`  as same as the sync version `SnowflakeOperator`

closes: #714

### DIFF
--- a/astronomer/providers/snowflake/hooks/snowflake.py
+++ b/astronomer/providers/snowflake/hooks/snowflake.py
@@ -41,7 +41,7 @@ class SnowflakeHookAsync(SnowflakeHook):
     def run(  # type: ignore[override]
         self,
         sql: Union[str, List[str]],
-        autocommit: bool = False,
+        autocommit: bool = True,
         parameters: Optional[dict] = None,  # type: ignore[type-arg]
     ) -> List[str]:
         """
--- a/astronomer/providers/snowflake/hooks/snowflake.py
+++ b/astronomer/providers/snowflake/hooks/snowflake.py
@@ -1,7 +1,9 @@
+from __future__ import annotations
+
 import asyncio
 from contextlib import closing
 from io import StringIO
-from typing import Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional, Union
 
 from airflow.providers.snowflake.hooks.snowflake import SnowflakeHook
 from asgiref.sync import sync_to_async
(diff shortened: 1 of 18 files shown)


astronomer/providers/snowflake/hooks/snowflake.py
```diff
@@ -81,16 +83,28 @@ def run(  # type: ignore[override]
                 conn.commit()
         return self.query_ids
 
-    def check_query_output(self, query_ids: List[str]) -> None:
+    def check_query_output(
+        self, query_ids: List[str], handler: Callable | None = None, return_last: bool = True
+    ) -> Optional[Union[Any, List[Any]]]:
```
pankajkoti, line 88: Is it possible to specify what `Any` would mean in the Union. Because `List[Any]` would anyway be a subset of` Any`. If it for representing just one record, and not a list can we specify the type of the record so that it gives a better idea of the return type?

astronomer/providers/snowflake/operators/snowflake.py
```diff
@@ -139,9 +142,11 @@ def execute_complete(
...
-                hook.check_query_output(qids)
+                handler = self.handler or fetch_all_handler
+                results = hook.check_query_output(qids, handler)
                 self.log.info("%s completed successfully.", self.task_id)
-                return None
+                if self.do_xcom_push:
+                    return results
         else:
             self.log.info("%s completed successfully.", self.task_id)
             return None
```
pankajkoti, line 152: We can remove this line, right? By default a function return `None`, no?

astronomer/providers/snowflake/hooks/snowflake.py
```diff
@@ -81,16 +83,28 @@ def run(  # type: ignore[override]
...
         """Once the query is finished fetch the result and log it in airflow"""
         with closing(self.get_conn()) as conn:
             with closing(conn.cursor(DictCursor)) as cur:
+                results = []
                 for query_id in query_ids:
                     cur.get_results_from_sfqid(query_id)
-                    cur.fetchall()
+                    if handler is not None:
+                        result = handler(cur)
+                        results.append(result)
```
pankajkoti, line 97: If handler is None, do we need to call `cur.fetchall()`? Or `cur.rowcount` in the below log line is accessible without calling `fetchall()` ?

pankajkoti (commented):
Also, mypy seems to have some failures

astronomer/providers/snowflake/hooks/snowflake.py
```diff
@@ -81,16 +83,28 @@ def run(  # type: ignore[override]
...
                     self.log.info("Rows affected: %s", cur.rowcount)
                     self.log.info("Snowflake query id: %s", query_id)
 
+        if handler is None:
+            return None
+        elif return_last:
+            return results[-1]
```
pankajastro, line 104: this looks weird to me, won't this would break if `handler` is None ?

astronomer/providers/snowflake/hooks/snowflake.py
pankajastro: yeah, `List[Any]` will be cleaner and make sense here

astronomer/providers/snowflake/hooks/snowflake.py
bharanidharan14: if handler is None by default, I am using `fetch_all_handler` which is `cur.fetchall()`

astronomer/providers/snowflake/operators/snowflake.py
pankajastro: in case of no event from the trigger, I think should be an error case, WDYT?

astronomer/providers/snowflake/hooks/snowflake.py
bharanidharan14: It can be list of response or it can be single response or it can be None as well

astronomer/providers/snowflake/operators/snowflake.py
bharanidharan14: Yeah if no event is passed we need to raise an error, now fixed it

astronomer/providers/snowflake/hooks/snowflake.py
bharanidharan14: If the handler is None it won't fail it will return None from this function

PR #763, opened 2022-11-16
Fix issue in SnowflakeOperatorAsync, SnowflakeSqlApiOperatorAsync  and RedshiftSQLOperatorAsync

bharanidharan14:
SnowflakeOperatorAsync, SnowflakeSqlApiOperatorAsync and RedshiftSQLOperatorAsync Operator failed due to new RC version changes apache-airflow-providers-amazon==6.1.0rc1 and apache-airflow-providers-snowflake==4.0.0rc1 the sync version of those operators started inherited from SQLExecuteQueryOperator which doesn't have snowflake_conn_id and redshift_conn_id. This issue is fixed by with backward compatibility


phanikumv (changes requested):
backward compatibility is missing in this code.

PR #777, opened 2022-11-22
Remove mypy ignore specification in `WasbHookAsync`

### LINKED ISSUE #773: Fix mypy issues on micrsoft azure (wasb hook).
Currently, we have skipped the mypy checks on wasb hook.
https://github.com/astronomer/astronomer-providers/blob/main/astronomer/providers/microsoft/azure/hooks/wasb.py#L106-L112
https://github.com/astronomer/astronomer-providers/blob/main/astronomer/providers/microsoft/azure/hooks/wasb.py#L79-L88
https://github.com/astronomer/astronomer-providers/blob/main/astronomer/providers/microsoft/azure/triggers/wasb.py#L129
 and fix this mypy issues properly.

phanikumv:
Closes #773

### DIFF
--- a/astronomer/providers/microsoft/azure/hooks/wasb.py
+++ b/astronomer/providers/microsoft/azure/hooks/wasb.py
@@ -25,7 +25,7 @@ class WasbHookAsync(WasbHook):
         super().__init__()
         self.conn_id = wasb_conn_id
         self.public_read = public_read
-        self.blob_service_client = self.get_conn()
+        self.blob_service_client: BlobServiceClient = self.get_conn()
 
     def get_conn(self) -> BlobServiceClient:
         """Returns the async BlobServiceClient object."""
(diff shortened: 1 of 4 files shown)


astronomer/providers/microsoft/azure/hooks/wasb.py
```diff
@@ -22,10 +22,9 @@ def __init__(
         wasb_conn_id: str = "wasb_default",
         public_read: bool = False,
     ) -> None:
-        super().__init__()
```
pankajkoti, line 25: Why are we removing `super().__init__()` ?

phanikumv: If you look at parent class, you see that you already do everything it does in this __init__

pankajkoti (approved):

PR #781, opened 2022-11-25
Fix `SnowflakeOperatorAsync` XCom value to return

bharanidharan14:
Return an XCom of results in SnowflakeOperatorAsync as same as the sync version SnowflakeOperator

closes: https://github.com/astronomer/astronomer-providers/issues/714

### DIFF
--- a/astronomer/providers/snowflake/operators/snowflake.py
+++ b/astronomer/providers/snowflake/operators/snowflake.py
@@ -198,7 +198,7 @@ class SnowflakeOperatorAsync(SnowflakeOperator):
                 self.log.info("%s completed successfully.", self.task_id)
                 if self.do_xcom_push:
                     return results
-                return None  # pragma: no cover
+                return None
         else:
             raise AirflowException("Did not receive valid event from the trigerrer")
 
--- a/astronomer/providers/snowflake/operators/snowflake.py
+++ b/astronomer/providers/snowflake/operators/snowflake.py
@@ -198,7 +198,6 @@ class SnowflakeOperatorAsync(SnowflakeOperator):
                 self.log.info("%s completed successfully.", self.task_id)
                 if self.do_xcom_push:
                     return results
-                return None
         else:
             raise AirflowException("Did not receive valid event from the trigerrer")
 
--- a/astronomer/providers/snowflake/operators/snowflake.py
+++ b/astronomer/providers/snowflake/operators/snowflake.py
@@ -1,6 +1,8 @@
+from __future__ import annotations
+
 import typing
 from datetime import timedelta
-from typing import Any, Dict, List, Optional, Union
+from typing import Any, Callable, Dict, List, Optional, Union
 
 from airflow.exceptions import AirflowException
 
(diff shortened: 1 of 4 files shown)


bharanidharan14:
Closed PR https://github.com/astronomer/astronomer-providers/pull/752 created separate PR to return an XCom of results in SnowflakeOperatorAsync as same as the sync version SnowflakeOperator.

bharanidharan14:
@rajaths010494 @pankajkoti @phanikumv can you review this PR

astronomer/providers/snowflake/operators/snowflake.py
```diff
@@ -169,7 +180,7 @@ def execute(self, context: Context) -> None:
 
     def execute_complete(
         self, context: Context, event: Optional[Dict[str, Union[str, List[str]]]] = None
-    ) -> None:
+    ) -> Any:
```
rajaths010494, line 183: Do we need `Any `? can we  not specify the exact ones?

bharanidharan14: It can be a list of Dict, None, tuples, or list of tuples, I initially assigned it as the union of list, tuple still it was complaining

astronomer/providers/snowflake/operators/snowflake.py
```diff
@@ -183,12 +194,13 @@ def execute_complete(
             elif "status" in event and event["status"] == "success":
                 hook = self.get_db_hook()
                 qids = typing.cast(List[str], event["query_ids"])
-                hook.check_query_output(qids)
+                results = hook.check_query_output(qids, self.handler)
                 self.log.info("%s completed successfully.", self.task_id)
-                return None
+                if self.do_xcom_push:
+                    return results
+                return None  # pragma: no cover
```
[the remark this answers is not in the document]
bharanidharan14: Fixed it now

rajaths010494 (approved):

astronomer/providers/snowflake/operators/snowflake.py
```diff
@@ -183,12 +194,13 @@ def execute_complete(
             elif "status" in event and event["status"] == "success":
                 hook = self.get_db_hook()
                 qids = typing.cast(List[str], event["query_ids"])
-                hook.check_query_output(qids)
+                results = hook.check_query_output(qids, self.handler)
                 self.log.info("%s completed successfully.", self.task_id)
+                if self.do_xcom_push:
+                    return results
                 return None
```
pankajkoti, line 201: Why do we need to explicitly write `return None`. What does it return if we don't specify this?

phanikumv: good point. we shouldnt return None

bharanidharan14: Yeah initially it was nothing it was just 
```
if self.do_xcom_push:
                    return results
```
mypy was complaining it expecting a return  it tried fixing with optional return for the function, still it is complaining

pankajkoti: can you paste the error you were getting from mypy? maybe remove the return and get the output from local run of mypy (no need to push)

bharanidharan14: Screenshot 2022-11-29 at 7 56 34 PM

bharanidharan14: if the return type of the execute_complete method is set to Any it is not complaining, if I set that to Optional list of dict or tuple it is complaining

pankajkoti: Okay 👍🏽

Issue #750, opened 2022-11-10
Add poke_interval to S3KeySensor/Trigger


tseruga:
**Is your feature request related to a problem? Please describe.**
Unlike the majority of triggers, the `S3KeyTrigger` (and by extension, the sensor itself) does not sleep for any period of time when looking for a file in s3. This causes the trigger to fire as fast as s3 can return a response regarding the existence of a key.

This can be seen here:
https://github.com/astronomer/astronomer-providers/blob/942a6aa263f765fe14b27d508517be2b3accb0f5/astronomer/providers/amazon/aws/triggers/s3.py#L62-L70

and contrasted with another trigger that performs a sleep to throttle requests in the `while` loop:
https://github.com/astronomer/astronomer-providers/blob/942a6aa263f765fe14b27d508517be2b3accb0f5/astronomer/providers/amazon/aws/triggers/batch.py#L157-L166

The lack of a form of throttling can cause the triggerer process to become overwhelmed when sensing for a large number of files in s3.

**Describe the solution you'd like**
Ideally the `S3KeyTrigger` would accept `poke_interval` as a kwarg to allow for throttling between calls to s3. This is already the standard used by other triggers, so it makes sense for this to be applied to the `S3KeyTrigger` as well.

PR #782, opened 2022-11-28
Added poke interval for the S3KeyTrigger

bharanidharan14:
Added `poke_interval` for the S3KeyTrigger. Adding poke_interval makes the trigger sleep for seconds to avoid the trigger process becoming overwhelmed when sensing a large number of files in s3.

closes: #750


rajaths010494 (approved):
LGTM

pankajkoti (approved):

PR #793, opened 2022-12-01
Fix return value of snowflakeOperatorAsync

bharanidharan14:
Passed the `return_last` parameter to hook `check_query_output` which is now the operator params, this param helps the user to return the result of only the last statement output or all statement output. By default, it is set to `True`, which will return the last value in case of multiple query execution.


phanikumv (commented):
could you please mention why we are doing this change in the description.

bharanidharan14:
> could you please mention why we are doing this change in the description.

@phanikumv Updated the description, please take a look

phanikumv (approved):

rajaths010494 (approved):

PR #797, opened 2022-12-02
Bump version to 1.13.0.dev1

bharanidharan14:
Since 1.12.0 is released, so bumped up this version to 1.13.0.dev1


rajaths010494:
We need to fix mypy which @pankajkoti  has already pushed in commit https://github.com/astronomer/astronomer-providers/commit/fcf7ffa300baec70813f1d9f40613ebd0c056a99

pankajkoti (approved):

rajaths010494 (approved):

Issue #568, opened 2022-08-02
class based Unit Test


pankajastro:
- Convert function-based test to class-based test example: 
- Move any third-party mock object to the utils module if the object makes the test crowded 
- Move constant to utils module if getting used at multiple places
- Remove context  fixture from the test file and use the one which is available in conftest.py

Need to fix the below provider 

- [x] amazon
- [x] apache
- [x] cncf
- [x] core
- [x] databricks
- [x] google
- [x] http
- [x] microsoft
- [x] snowflake

pankajastro:
We have converted functional tests to class-based.

PR #837, opened 2023-01-10
Databricks class based tests

pankajastro:
closes: https://github.com/astronomer/astronomer-providers/issues/693
related-to: https://github.com/astronomer/astronomer-providers/issues/568

- convert the function-based test to a class-based
- remove context fixture definition from operator/sensor/extractor and reuse from conftest
- remove function create_context from a couple of test files and reuse it from the util
- remove duplicate code for initializing op/senor/trigger/hook at possible places


rajaths010494 (approved):
Overall LGTM

Issue #433, opened 2022-06-11
Fix error when SnowflakeHookAsync take empty list in sql param


pankajastro:
OSS PR: https://github.com/apache/airflow/pull/23767

PR #810, opened 2022-12-13
Sync SnowflakeHook return type and add SnowflakeSensorAsync

park-peter:
- Sync SnowflakeHook return type to conform to upstream OSS SnowflakeHook from provider 4.0.2 (https://github.com/apache/airflow/pull/28006)
- Add SnowflakeSensorAsync (need to add test)

Fixes https://github.com/astronomer/astronomer-providers/issues/729

PR #838, opened 2023-01-10
Snowflake handle empty sql list

pankajastro:
closes: https://github.com/astronomer/astronomer-providers/issues/433


park-peter:
this will conflict with #810 as it changes the way SnowflakeHook returns the result to be in line with the non-deferrable hook. Please wait until #810 is merged.

rajaths010494 (approved):

Issue #842, opened 2023-01-16
Fix GKE Operator auth failures


pankajkoti:
`example_google_kubernetes_engine` has been failing on Integration tests deploymnet on Astro Cloud since 10th January with the below error.

```
[2023-01-16, 00:22:52 UTC] {process_utils.py:187} INFO - CRITICAL: ACTION REQUIRED: gke-gcloud-auth-plugin, which is needed for continued use of kubectl, was not found or is not executable. Install gke-gcloud-auth-plugin for use with kubectl by following https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke
[2023-01-16, 00:22:52 UTC] {process_utils.py:187} INFO - kubeconfig entry generated for provider-team-gke-cluster.
[2023-01-16, 00:22:52 UTC] {kubernetes_pod.py:621} INFO - Building pod astro-k8s-gke-test-pod-41fa02f539cb459395594f025ae98bd8 with labels: {'dag_id': 'example_google_kubernetes_engine', 'task_id': 'start_pod', 'run_id': 'kubernetes_engine_dag_example_google_kubernetes_engin-3a22d825f', 'kubernetes_pod_operator': 'True', 'try_number': '1'}
[2023-01-16, 00:22:52 UTC] {kube_config.py:517} ERROR - [Errno 2] No such file or directory: 'gke-gcloud-auth-plugin'
[2023-01-16, 00:22:52 UTC] {taskinstance.py:1772} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/astronomer/providers/google/cloud/operators/kubernetes_engine.py", line 112, in execute
    self._get_or_create_pod(context)
  File "/usr/local/lib/python3.9/site-packages/astronomer/providers/google/cloud/operators/kubernetes_engine.py", line 106, in _get_or_create_pod
    self.pod: k8s.V1Pod = self.get_or_create_pod(self.pod_request_obj, context)
  File "/usr/local/lib/python3.9/site-packages/airflow/providers/cncf/kubernetes/operators/kubernetes_pod.py", line 411, in get_or_create_pod
    pod = self.find_pod(self.namespace or pod_request_obj.metadata.namespace, context=context)
  File "/usr/local/lib/python3.9/site-packages/airflow/providers/cncf/kubernetes/operators/kubernetes_pod.py", line 393, in find_pod
    pod_list = self.client.list_namespaced_pod(
[...]

PR #843, opened 2023-01-16
Install gke-gcloud-auth-plugin to fix auth failures

pankajkoti:
With kubectl v1.26, the auth mechanism has been altered and now it needs the gke-gcloud-auth-plugin for authentication to succeed. Install the same in the docker image getting built to be run in CI.

closes: #842


rajaths010494:
we need to update also in dev/Dockerfile.google_cloud

phanikumv (approved):

PR #848, opened 2023-01-20
Replace flake8, isort with ruff

pankajastro:
see details: https://github.com/astronomer/astro-sdk/pull/1578#issue-1530786861
> https://github.com/charliermarsh/ruff/ is a faster replacement of most of the linting tools we use, and is starting to picked up by several other projects. Even project like Pandas have adopted Ruff (https://github.com/pandas-dev/pandas/pull/50160)


ruff.toml
```diff
@@ -0,0 +1,35 @@
...
+    "C",    # flake8-comprehensions
+#    "ANN",  # flake8-comprehensions
+    "ISC",  # flake8-implicit-str-concat
+    "T10",  # flake8-debugger
+    "A",    # flake8-builtins
+    "UP",   # pyupgrade
+]
+
+# Remove this in future
+extend-ignore = ["A002", "C901", "E402", "E501", "ISC001", "ISC003", "UP030"]
```
pankajastro, line 18: I'll enable these check separately

phanikumv (commented):
There are lot of changes here, can we do a deployment to astro cloud and see whether all our  integrations DAGs are running fine?

pankajastro:
> There are lot of changes here, can we do a deployment to astro cloud and see whether all our integrations DAGs are running fine?

We can do, I'll start. most of the changes are static.

astronomer/providers/google/cloud/hooks/bigquery.py
```diff
@@ -4,10 +4,10 @@
 from airflow.exceptions import AirflowException
 from airflow.providers.google.cloud.hooks.bigquery import BigQueryHook
 from gcloud.aio.bigquery import Job, Table
-from google.cloud.bigquery import CopyJob, ExtractJob, LoadJob, QueryJob
 from requests import Session
 
 from astronomer.providers.google.common.hooks.base_google import GoogleBaseHookAsync
+from google.cloud.bigquery import CopyJob, ExtractJob, LoadJob, QueryJob
```
kaxil, line 10: this should stay at the top -- it is unfortunately detecting this is a "internal" module hence it is at the bottom -- check how we did it on Airflow https://github.com/apache/airflow/blob/d03b9a76914babef23b84acdb6061b0b93bdc2e3/pyproject.toml#L75-L97 -- we should do the same here

astronomer/providers/snowflake/extractors/snowflake.py
```diff
@@ -15,9 +15,9 @@ class SnowflakeAsyncExtractor(BaseExtractor):
     source_type = "SNOWFLAKE"
     default_schema = "PUBLIC"
 
-    def __init__(self, operator):  # type: ignore[no-untyped-def]
+    def __init__(self, operator: BaseOperator) -> None:
         super().__init__(operator)
-        self.conn: "Connection" = None
+        self.conn: Connection = Connection()
```
kaxil, line 20: No,  we shouldn't need to do this

ruff.toml
```diff
@@ -0,0 +1,35 @@
+line-length = 120
```
kaxil, line 1: Since we already have a top-level `pyproject.toml` file, we could move all these settings over there instead of `ruff.toml` like 

https://github.com/apache/airflow/blob/d03b9a76914babef23b84acdb6061b0b93bdc2e3/pyproject.toml#L29

kaxil: We didn't do it in Astro SDK as it is a mono-repo

astronomer/providers/snowflake/extractors/snowflake.py
pankajastro: mypy fails otherwise
```
astronomer/providers/snowflake/extractors/snowflake.py: note: In member "__init__" of class "SnowflakeAsyncExtractor":
astronomer/providers/snowflake/extractors/snowflake.py:20: error: Incompatible
types in assignment (expression has type "None", variable has type "Connection")
 [assignment]
            self.conn: "Connection" = None
```

astronomer/providers/google/cloud/hooks/bigquery.py
pankajastro: Ah, didn't notice. thanks for pointing out

ruff.toml
pankajastro: fixed it

astronomer/providers/snowflake/extractors/snowflake.py
kaxil: So you should change it to: `self.conn: Connection | None = None`

kaxil: as we don't want to instantiate it unncessarily

astronomer/providers/snowflake/extractors/snowflake.py
```diff
@@ -15,27 +17,27 @@ class SnowflakeAsyncExtractor(BaseExtractor):
     source_type = "SNOWFLAKE"
     default_schema = "PUBLIC"
 
-    def __init__(self, operator):  # type: ignore[no-untyped-def]
+    def __init__(self, operator: BaseOperator) -> None:
         super().__init__(operator)
-        self.conn: "Connection" = None
+        self.conn: Connection | None = None
```
kaxil, line 22: If this throws an error, use `Optional[Connection]`

kaxil, line 22:
```suggestion
        self.conn: Connection
```

kaxil:
```suggestion
        self.conn: Connection
```

kaxil (approved):

PR #849, opened 2023-01-20
Bump runtime image to 7.2.0

pankajastro:
update runtime image.
Look like yesterday while testing rc I upgraded the version

Bump runtime image to 7.2.0

### DIFF
--- a/.circleci/integration-tests/Dockerfile.astro_cloud
+++ b/.circleci/integration-tests/Dockerfile.astro_cloud
@@ -1,5 +1,5 @@
 # Deploy from astro runtime image
-FROM quay.io/astronomer/astro-runtime-dev:7.2.0-alpha1-base
+FROM quay.io/astronomer/astro-runtime:7.2.0-base
 
 ENV AIRFLOW__CORE__DAGS_ARE_PAUSED_AT_CREATION=False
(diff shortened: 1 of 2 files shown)


pankajastro:
PR https://github.com/astronomer/astronomer-providers/pull/848 will fix the CI

phanikumv (commented):
should we update Makefile as well?

pankajkoti (commented):
Should we create new deployments while testing RCs? 
What should be the right way as deployments most often cannot be downgraded back to stable released versions once we deploy RC images?

cc: @phanikumv

phanikumv:
> Should we create new deployments while testing RCs? What should be the right way as deployments most often cannot be downgraded back to stable released versions once we deploy RC images?
> 
> cc: @phanikumv

yeah that's a good idea, and let's improve test-rc-deps to do that @pankajkoti

pankajastro:
7.2.0 has been release to lets bump up the version

pankajastro:
> I think we will need a new make target as this is for testing Runtime image changes and not provider RCs. I have created a ticket here: #851

Added my comments on above ticket

phanikumv (approved):

PR #855, opened 2023-01-23
Use entire tenant domain name in dbt Cloud connection

josh-fell:
Related: https://github.com/apache/airflow/issues/27146 https://github.com/apache/airflow/pull/28890

In OSS Airflow, the dbt Cloud connection has replaced `schema` for `host` to provide the "tenant" name. This fix was included in the latest version of the dbt Cloud provider (2.3.1) and should also be reflected in this provider.

### DIFF
--- a/astronomer/providers/dbt/cloud/hooks/dbt.py
+++ b/astronomer/providers/dbt/cloud/hooks/dbt.py
@@ -1,3 +1,4 @@
+import warnings
 from functools import wraps
 from inspect import signature
 from typing import Any, Dict, List, Optional, Tuple, TypeVar, cast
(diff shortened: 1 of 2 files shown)


astronomer/providers/dbt/cloud/hooks/dbt.py
```diff
@@ -74,19 +87,16 @@ def get_request_url_params(
...
         """
         data: Dict[str, Any] = {}
-        base_url = f"https://{tenant}.getdbt.com/api/v2/accounts/"
+        base_url = f"https://{tenant}/api/v2/accounts/"
         if include_related:
             data = {"include_related": include_related}
-        if base_url and not base_url.endswith("/") and endpoint and not endpoint.startswith("/"):
-            url = base_url + "/" + endpoint
-        else:
-            url = (base_url or "") + (endpoint or "")
```
josh-fell, line 89: Removed since this line would never be executed.

josh-fell:
If the fix here is acceptable, I can add it to OSS Airflow as well. Although there is probably some other refactoring that could be done (i.e. a shared function between the Async and sync versions).

rajaths010494 (approved):

pankajastro (approved):

PR #866, opened 2023-02-07
865 the hook needs to contain similar kwargs as the google providers

erdos2n:
I have added the `location` parameter in the hook. I do think this needs to get upgraded to integrate the `use_legacy_sql` parameter, as well, but that change may not be so simple given it could change the requirements.txt.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,6 +25,13 @@ workflows:
           <<: *all_branches_and_version_tag
       - mypy:
           <<: *all_branches_and_version_tag
+      - generate-constraints:
+          name: constraints-<< matrix.python_version >>-<< matrix.airflow_version >>
+          matrix:
+            parameters:
+              python_version: ["3.8", "3.9", "3.10"]
+              airflow_version: ["2.3.4", "2.4.2", "2.5.1"]
+          <<: *all_branches_and_version_tag
       - build-docs:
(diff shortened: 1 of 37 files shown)


pankajastro (approved):
LGTM

rajaths010494:
rebase once the PR gets merged https://github.com/astronomer/astronomer-providers/pull/867
and mypy issue should be resolved

pankajastro:
FYI: Look like google async lib have some issue with no us location https://github.com/apache/airflow/issues/29307

erdos2n:
One of the errors, currently, is a permission issue. It would make sense that it's a region issue, but I haven't verified for sure, since it's not my service-account. 

We should include this and just update the docs to reflect a possible issue with non-US locations.

rajaths010494:
Can we fix the failing tests as well.

erdos2n:
I've gone ahead and removed the `use_legacy_sql` kwarg from this PR since it was failing the tests. I think it's best to open another issue for this, as the current issue with the `location` kwarg is breaking a DAG for a customer. I will open a new, much more broad issue for update the `bigquery.py` file to meet the current kwargs of the provider package.

rajaths010494:
Can you rebase the branch? There is mypy issues also

rajaths010494 (approved):

PR #876, opened 2023-02-15
Fix mypy issues

rajaths010494:
https://app.circleci.com/pipelines/github/astronomer/astronomer-providers/4007/workflows/66592839-4f18-4ca7-8ea4-a64e46f8fd36/jobs/17639

image

### DIFF
--- a/astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py
+++ b/astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py
@@ -81,7 +81,7 @@ class KubernetesPodOperatorAsync(KubernetesPodOperator):
         self.pod: k8s.V1Pod = self.get_or_create_pod(self.pod_request_obj, context)
         self.defer()
 
-    def execute_complete(self, context: Context, event: Dict[str, Any]) -> Any:
+    def execute_complete(self, context: Context, event: Dict[str, Any]) -> Any:  # type: ignore[override]
         """Deprecated; replaced by trigger_reentry."""
         warnings.warn(
             "Method `execute_complete` is deprecated and replaced with method `trigger_reentry`.",
--- a/astronomer/providers/google/cloud/operators/dataproc.py
+++ b/astronomer/providers/google/cloud/operators/dataproc.py
@@ -379,7 +379,7 @@ class DataprocUpdateClusterOperatorAsync(DataprocUpdateClusterOperator):
             method_name="execute_complete",
         )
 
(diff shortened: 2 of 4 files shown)


astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py
```diff
@@ -81,7 +81,7 @@ def execute(self, context: Context) -> None:  # noqa: D102
         self.pod: k8s.V1Pod = self.get_or_create_pod(self.pod_request_obj, context)
         self.defer()
 
-    def execute_complete(self, context: Context, event: Dict[str, Any]) -> Any:
+    def execute_complete(self, context: Context, event: Dict[str, Any]) -> Any:  # type: ignore[override]
```
pankajastro, line 84:
```suggestion
    def execute_complete(self, context: Context, event: Dict[str, Any]) -> Any
```

pankajastro: can you please try above ^^

astronomer/providers/google/cloud/operators/dataproc.py
```diff
@@ -379,7 +379,7 @@ def execute(self, context: "Context") -> None:
             method_name="execute_complete",
         )
 
-    def execute_complete(self, context: Dict[str, Any], event: Optional[Dict[str, Any]] = None) -> Any:
+    def execute_complete(self, context: Dict[str, Any], event: Optional[Dict[str, Any]] = None) -> Any:  # type: ignore[override]
```
pankajastro, line 382:
```suggestion
    def execute_complete(self, context: Context, event: Dict[str, Any]) -> Any
```

pankajastro: can you please try above ^^

pankajastro:
We should match the signature of `execute_complete` with the parent class `execute_complete ` method instead of ignoring mypy, WDYT?

rajaths010494:
> We should match the signature of `execute_complete` with the parent class `execute_complete ` method instead of ignoring mypy, WDYT?

Agreed where ever we can we should match to parent class. But in OSS they are certain classes use 
`def execute_complete(self, context: Context, event: dict, **kwargs):`  if we use this signature mypy will still throw error because dict should be defined as `Dict[str, Any]`. Let me check if the changes can be made same as parent class.

pankajkoti (approved):

pankajastro (approved):

PR #889, opened 2023-02-23
Run dbt example dag as part of  master dag

rajaths010494:
DBT example dag was not being triggered as part of master dag.


pankajkoti (approved):

pankajastro (approved):
LGTM. we should test it once on Astro

phanikumv (approved):

Issue #895, opened 2023-02-27
HttpSensorAsync always uses default connection ID


marc-fouquet:
**Describe the bug**
HttpSensorAsync always uses the default connection ID.

**To Reproduce**
1. Use astronomer.providers.http.sensors.http import HttpSensorAsync
2. Set a connection ID other than default as parameter
```python
sense = HttpSensorAsync(
    task_id="sense",
    http_conn_id="some_connection_id",
    endpoint=endpoint,
    poke_interval=60, # 1 Minute
    timeout=60 * 60 * 24 * 7,
)
```
3. Look at the triggerer's logfile:
```
{triggerer_job.py:359} INFO - Trigger <astronomer.providers.http.triggers.http.HttpTrigger endpoint=kv/LOCATION_statistics_call_logs_2022-11-14.csv, data={}, headers={}, extra_options={}, http_conn_id=http_default, poke_interval=60> (ID 481) starting
```

**Expected behavior**
```
http_conn_id=some_connection_id
```

**Desktop (please complete the following information):**
 - OS: Linux
 - astronomer-providers==1.15.0
 - airflow: 2.5.1
 - Python 3.10

**Additional context**
This appears to fix it for me:

```python
    def execute(self, context: Context) -> None:
        """
        Logic that the sensor uses to correctly identify which trigger to
        execute, and defer execution as expected.
        """
        # TODO: We can't currently serialize arbitrary function
        # Maybe we set method_name as users function??? to run it again
        # and evaluate the response.
        if self.response_check:
            self.log.warning("Since response_check param is passed, using the sync version of the sensor.")
            super().execute(context=context)
        else:
            self.defer(
                timeout=timedelta(seconds=self.timeout),
                trigger=HttpTrigger(
                    method=self.hook.method,  # TODO: Fix this to directly get method from ctor
                    http_conn_id=self.http_conn_id,
                    endpoint=self.endpoint,
                    data=self.request_params,
                    headers=self.headers,
                    extra_options=self.extra_options,
                    poke_interval=self.poke_interval,
                ),
                method_name="execute_complete",
            )
```

pankajastro:
correct, thanks for raising it.

PR #896, opened 2023-02-27
Pass connection_id in http trigger

pankajastro:
closes: https://github.com/astronomer/astronomer-providers/issues/895


rajaths010494:
We may need to address the mypy issue on a separate PR. It seems to be coming from Microsoft wasb hook

rajaths010494 (approved):

Issue #897, opened 2023-02-27
Fix Microsoft Azure Provider example DAG to run ADF pipeline


pankajkoti:
On the new Astronomer Validation org deployment, the ADF DAG fails with the below error:
```
[2023-02-22, 07:56:55 UTC] {standard_task_runner.py:83} INFO - Job 5912: Subtask run_pipeline_wait
[2023-02-22, 07:56:55 UTC] {task_command.py:388} INFO - Running <TaskInstance: example_async_adf_run_pipeline.run_pipeline_wait manual__2023-02-22T07:54:24.179097+00:00 [running]> on host 172.20.21.95
[2023-02-22, 07:56:55 UTC] {taskinstance.py:1768} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/astronomer/providers/microsoft/azure/operators/data_factory.py", line 86, in execute_complete
    raise AirflowException(event["message"])
airflow.exceptions.AirflowException: A Subscription ID is required to connect to Azure Data Factory.
[2023-02-22, 07:56:55 UTC] {taskinstance.py:1318} INFO - Marking task as FAILED. dag_id=example_async_adf_run_pipeline, task_id=run_pipeline_wait, execution_date=20230222T075424, start_date=20230222T075650, end_date=20230222T075655
[2023-02-22, 07:56:55 UTC] {manager.py:61} WARNING - Unable to find an extractor. task_type=AzureDataFactoryRunPipelineOperatorAsync airflow_dag_id=example_async_adf_run_pipeline task_id=run_pipeline_wait airflow_run_id=manual__2023-02-22T07:54:24.179097+00:00
```

pankajkoti:
We had created a Zendesk ticket on Wednesday 22/02/2023 to get help from support for this issue: https://astronomer.zendesk.com/agent/tickets/16624

PR #899, opened 2023-02-27
Support retriveing extra fields which do not have extra prefix set - Azure ADF DAG failure

pankajkoti:
With the 5.0.0 release of the Microsoft Azure PR, the PR https://github.com/apache/airflow/pull/27047 introduced a breaking change where while creating new connections, the ``extra__`` prefix is no longer set for extra fields in the conneciton. This issue was not identified with testing the 5.0.0 RC because, it only happens for new connections that are created. The existing connections still contain the extra fiels with the ``extra__`` prefix. Hence, the existing code which looks for the connection field with the prefix ``extra__azure_data_factory__subscriptionId`` works on the older deployment with the new provider release as the connection was created before the release, but it fails on new deployments when a fresh connection is created.

To fix this, we're removing the prefix while retrieving the connection field and at the same time we're supporting previously created connections by using the same ``get_field`` method from Airflow OSS introduced in the same PR above to allow backward compatibility.

closes: #897

### DIFF
--- a/astronomer/providers/microsoft/azure/hooks/data_factory.py
+++ b/astronomer/providers/microsoft/azure/hooks/data_factory.py
@@ -1,9 +1,10 @@
+"""This module contains the Azure Data Factory hook's asynchronous implementation."""
 import inspect
 from functools import wraps
 from typing import Any, Optional, TypeVar, Union, cast
 
 from airflow.exceptions import AirflowException
-from airflow.providers.microsoft.azure.hooks.data_factory import AzureDataFactoryHook
+from airflow.providers.microsoft.azure.hooks.data_factory import AzureDataFactoryHook, get_field
 from asgiref.sync import sync_to_async
 from azure.identity.aio import ClientSecretCredential, DefaultAzureCredential
 from azure.mgmt.datafactory.aio import DataFactoryManagementClient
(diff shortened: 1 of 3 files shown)


rajaths010494 (approved):

rajaths010494:
@pankajkoti good that we got the root cause of the issue and fixed it.

pankajkoti:
I am merging this PR to test deployment from CircleCI main branch to new Astronomer Validation Org deployment. I will address comments if any that come later in a separate PR.

cc: @phanikumv @pankajastro

astronomer/providers/microsoft/azure/hooks/data_factory.py
```diff
@@ -1,9 +1,10 @@
+"""This module contains the Azure Data Factory hook's asynchronous implementation."""
 import inspect
 from functools import wraps
 from typing import Any, Optional, TypeVar, Union, cast
 
 from airflow.exceptions import AirflowException
-from airflow.providers.microsoft.azure.hooks.data_factory import AzureDataFactoryHook
+from airflow.providers.microsoft.azure.hooks.data_factory import AzureDataFactoryHook, get_field
```
ashb, line 7: Importing this will fail on v4 of the provider I think - is that a problem? Is this meant to only work with v5+?

pankajkoti: It's meant to support both versions. Thanks, @ashb , I will create another PR to fix this.

pankajkoti: Created a PR https://github.com/astronomer/astronomer-providers/pull/900 to fix this.

PR #900, opened 2023-02-27
Fix get_field import availability - PR #899 followup

pankajkoti:
PR #899 introduced an import for method ``get_field`` and started its usage. However, as reviewed and suggested by @ashb this will fail on providers with version < 5.0.0 as it was only the release 5.0.0 which included the ``get_field`` method. To solve this, we remove the dependency on the imported method and introduce the same method in local module.

### DIFF
--- a/astronomer/providers/microsoft/azure/hooks/data_factory.py
+++ b/astronomer/providers/microsoft/azure/hooks/data_factory.py
@@ -15,7 +15,7 @@ Credentials = Union[ClientSecretCredential, DefaultAzureCredential]
 T = TypeVar("T", bound=Any)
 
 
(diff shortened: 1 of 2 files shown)


astronomer/providers/microsoft/azure/hooks/data_factory.py
```diff
@@ -15,6 +15,23 @@
 T = TypeVar("T", bound=Any)
 
 
+def get_field(extras: dict, field_name: str, strict: bool = False):
```
pankajkoti, line 18:
```suggestion
def get_field(extras: dict[str, Any], field_name: str, strict: bool = False):
```

rajaths010494 (approved):

PR #905, opened 2023-03-01
SFTPSensorAsync using newer_than doesn't work well with file_pattern

### LINKED ISSUE #903: SFTPSensorAsync not working exactly in the same way as SFTPSensor
**Describe the bug**
If a user authenticates to SFTP with login and password, `SFTPSensorAsync` throws an error:
```
  File "/usr/local/lib/python3.9/site-packages/astronomer/providers/sftp/sensors/sftp.py", line 67, in execute_complete
    raise AirflowException(event["message"])
airflow.exceptions.AirflowException: [Errno 2] No such file or directory: '"~/.ssh/known_hosts"'
```

**To Reproduce**
1. Create a local SFTP server (like shown [here](https://itnext.io/using-a-local-sftp-server-for-testing-659daaa0290a))
2. Create an Airflow connection:
  - conn id: `sftp_default`
  - conn type: `SFTP`
  - host: `host.docker.internal`
  - username: `foo`
  - password: `pass`
  - port: `2222`
3. Upload a test file to SFTP `/upload/test.txt`
4. Create a DAG:
```python
import pendulum
from airflow import DAG
from astronomer.providers.sftp.sensors.sftp import SFTPSensorAsync

with DAG(
    dag_id="sftp_test",
    start_date=pendulum.datetime(2023, 2, 1, tz="UTC"),
    schedule="@once",
):

    wait_for_file = SFTPSensorAsync(
        task_id="wait_for_file",
        sftp_conn_id="sftp_default",
        path="/upload/test.txt",
        poke_interval=15,
        timeout=60 * 10,
    )
```
5. See error
image

**Expected behavior**
`SFTPSensorAsync` should work in the same way as `SFTPSensor`. Now it only works after adding `{"known_hosts": "none"}` to the `Extra` field.

### LINKED ISSUE #904: SFTPSensorAsync using newer_than doesn't work well with file_pattern
**Describe the bug**
Please see the description [here](https://github.com/apache/airflow/issues/29781).

**Expected behavior**
There is a fix for this issue which needs to be implemented in the async operator as well, see [here](https://github.com/apache/airflow/pull/29794).

rajaths010494:
When we use file_pattern and newer_than in SFTPSensorAsync to find only the files that landed in SFTP after the data interval of the prior successful DAG run ({{ prev_data_interval_end_success  }}).
I have four text files (file.txt, file1.txt, file2.txt and file3.txt) but only file3.txt has the last modification date after the data interval of the prior successful DAG run. I use the following file pattern: "*.txt".
The moment the first file (file.txt) was matched and the modification date did not meet the requirement.

The other files matching the pattern should be checked as well.

closes #903 
closes #904

### DIFF
--- a/astronomer/providers/sftp/hooks/sftp.py
+++ b/astronomer/providers/sftp/hooks/sftp.py
@@ -28,7 +28,7 @@ class SFTPHookAsync(BaseHook):
     default_conn_name = "sftp_default"
     conn_type = "sftp"
     hook_name = "SFTP"
-    default_known_hosts = '"~/.ssh/known_hosts"'
+    default_known_hosts = "none"
 
     def __init__(  # nosec: B107
         self,
@@ -118,23 +118,20 @@ class SFTPHookAsync(BaseHook):
         except asyncssh.SFTPNoSuchFile:
             return None
 
-    async def get_file_by_pattern(self, path: str = "", fnmatch_pattern: str = "") -> str:
+    async def get_files_by_pattern(self, path: str = "", fnmatch_pattern: str = "") -> List[str]:
         """
         Returns the name of a file matching the file pattern at the provided path, if one exists
         Otherwise, raises an AirflowException to be handled upstream for deferring
         """
         files_list = await self.list_directory(path)
-
+        matched_files = []
         if files_list is None:
             raise AirflowException(f"No files at path {path} found...")
 
         for file in files_list:
-            if not fnmatch(file, fnmatch_pattern):
-                pass
-            else:
-                return file
-
-        raise AirflowException(f"No files matching file pattern were found at {path} — Deferring")
+            if fnmatch(file, fnmatch_pattern):
+                matched_files.append(file)
+        return matched_files
 
     async def get_mod_time(self, path: str) -> str:
         """
--- a/astronomer/providers/sftp/hooks/sftp.py
+++ b/astronomer/providers/sftp/hooks/sftp.py
@@ -124,13 +124,9 @@ class SFTPHookAsync(BaseHook):
         Otherwise, raises an AirflowException to be handled upstream for deferring
         """
         files_list = await self.list_directory(path)
-        matched_files = []
         if files_list is None:
             raise AirflowException(f"No files at path {path} found...")
-
-        for file in files_list:
-            if fnmatch(file, fnmatch_pattern):
-                matched_files.append(file)
+        matched_files = [file for file in files_list if fnmatch(file, fnmatch_pattern)]
         return matched_files
 
     async def get_mod_time(self, path: str) -> str:
--- a/tests/sftp/triggers/test_sftp.py
+++ b/tests/sftp/triggers/test_sftp.py
@@ -138,11 +138,11 @@ class TestSFTPTrigger:
 
     @pytest.mark.asyncio
     @mock.patch("astronomer.providers.sftp.hooks.sftp.SFTPHookAsync.get_files_by_pattern")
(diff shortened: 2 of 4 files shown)


tests/sftp/triggers/test_sftp.py
```diff
@@ -128,12 +137,12 @@ async def test_sftp_with_newer_then_date_greater(self, mock_mod_time):
         asyncio.get_event_loop().stop()
 
     @pytest.mark.asyncio
-    @mock.patch("astronomer.providers.sftp.hooks.sftp.SFTPHookAsync.get_file_by_pattern")
-    async def test_sftp_trigger_run_trigger_failure_state(self, mock_get_file_by_pattern):
+    @mock.patch("astronomer.providers.sftp.hooks.sftp.SFTPHookAsync.get_files_by_pattern")
+    async def test_sftp_trigger_run_trigger_failure_state(self, get_files_by_pattern):
```
Lee-W, line 141: May I know why do we change it to `get_files_by_pattern` here. According to other test cases in this test file, it seems adding a `mock_` before the mocked function is the convention.

tests/sftp/triggers/test_sftp.py
```diff
@@ -144,3 +153,21 @@ async def test_sftp_trigger_run_trigger_failure_state(self, mock_get_file_by_pat
...
             assert TriggerEvent(expected_event) == actual_event
+
+    @pytest.mark.asyncio
+    @mock.patch("astronomer.providers.sftp.hooks.sftp.SFTPHookAsync.get_files_by_pattern")
+    async def test_sftp_trigger_run_airflow_exception(self, mock_get_files_by_pattern):
+        """
+        Assert that a the task does not complete if the hook raises an AirflowException,
+        indicating that the task needs to be deferred
+        """
+        mock_get_files_by_pattern.side_effect = AirflowException("No files at path found...")
```
Lee-W, line 164: *nitpick*

it seems this might not be a valid message for 

```python
            raise AirflowException(f"No files at path {path} found...")
```

Maybe something like `No files at path swh path found...` would be better, but it won't affect the test.

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -118,23 +118,20 @@ async def list_directory(self, path: str = "") -> Optional[List[str]]:
...
         Returns the name of a file matching the file pattern at the provided path, if one exists
         Otherwise, raises an AirflowException to be handled upstream for deferring
         """
         files_list = await self.list_directory(path)
-
+        matched_files = []
         if files_list is None:
             raise AirflowException(f"No files at path {path} found...")
 
         for file in files_list:
```
Lee-W, line 131: The variable `matched_files` seems to be a bit far from where it's used to me.

I might try something like 

```python
matched_files = [
    file
    for file in files_list
    if fnmatch(file, fnmatch_pattern)
]
```

or `filter` might work as well

tests/sftp/triggers/test_sftp.py
rajaths010494: added `mock_` .

tests/sftp/triggers/test_sftp.py
rajaths010494: `AirflowException("No files at path /test/path/ found...")`

astronomer/providers/sftp/hooks/sftp.py
rajaths010494: `matched_files = [file for file in files_list if fnmatch(file, fnmatch_pattern)]` changed

Lee-W (approved):

PR #907, opened 2023-03-02
Fix mypy and add snowflake sensor dag to master dag

rajaths010494:
Fix mypy issues on main and  currently, the `example_snowflake_sensor` dag was not included as part of the master dag and this dag was not running on astro cloud. 
This PR adds the `example_snowflake_sensor` into master_dag.

### DIFF
--- a/astronomer/providers/snowflake/hooks/snowflake.py
+++ b/astronomer/providers/snowflake/hooks/snowflake.py
@@ -16,7 +16,7 @@ from snowflake.connector.util_text import split_statements
 
 def fetch_all_snowflake_handler(
     cursor: SnowflakeCursor,
-) -> list[dict[str, Any] | tuple[Any, ...] | None]: 
+) -> list[dict[str, Any] | tuple[Any, ...] | None]:
     """Handler for SnowflakeCursor to return results"""
     return cursor.fetchall()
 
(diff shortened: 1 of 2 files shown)


astronomer/providers/snowflake/hooks/snowflake.py
```diff
@@ -88,7 +88,7 @@ def run(  # type: ignore[override]
                         cur.execute_async(sql_statement)
                     query_id = cur.sfqid
                     self.log.info("Snowflake query id: %s", query_id)
-                    self.query_ids.append(query_id)
+                    self.query_ids.append(query_id)  # type: ignore[arg-type]
```
pankajkoti, line 91: I think we should avoid ignoring the type incompatibility error. In general, those errors help us to identify potential issues in our code, no?

I am looking at the parent hook and it defines self.query_ids type to be list[str]. But query_id = cur.sfqid could be None too. Hence, I believe we should not append None query_id values to self.query_ids, because we would not be able to check the statuses of such queries anyway with the None ID. 

```suggestion
                    if query_id:
                        self.query_ids.append(query_id)
```

pankajastro (approved):

pankajkoti (approved):

PR #908, opened 2023-03-02
Update ExternalTaskSensorAsync and SnowflakeSensorAsync to check on worker first

park-peter:
Based on internal discussion to mitigate race condition on deferrable sensor immediately getting successful, made the changes to the sensors to check on the worker first, then defer if the condition is not met.
- SnowflakeSensorAsync refactored to inherit from `SqlSensor` and rewrote the test
- ExternalTaskSensorAsync
- ExternalDeploymentTaskSensorAsync


Lee-W (commented):
Except for the failing CI, everything looks good to me

rajaths010494:
the mypy issues are solved as part of this PR https://github.com/astronomer/astronomer-providers/pull/907 if you rebase it should be gone.

rajaths010494 (approved):

Lee-W (approved):

PR #912, opened 2023-03-06
Make amazon sagemaker async operators compatible with new amazon provider

rajaths010494:
With the new RC release, the test cases were failing and dag was failing due to https://app.circleci.com/pipelines/github/astronomer/astronomer-providers/4127/workflows/1c8507c3-d802-46fc-9c1f-49c7fd9cc37e/jobs/19074,
This PR makes sagemaker operator compatible with an older and newer versions also.

### DIFF
--- a/astronomer/providers/amazon/aws/operators/sagemaker.py
+++ b/astronomer/providers/amazon/aws/operators/sagemaker.py
@@ -157,7 +157,7 @@ class SageMakerTransformOperatorAsync(SageMakerTransformOperator):
                 self._check_if_transform_job_exists()
             except AttributeError:
                 # for apache-airflow-providers-amazon>=7.3.0
-                transform_config["TransformJobName"] = self._get_unique_job_name(
+                transform_config["TransformJobName"] = self._get_unique_job_name(  # type: ignore[attr-defined]
                     transform_config["TransformJobName"],
                     self.action_if_job_exists == "fail",
                     self.hook.describe_transform_job,
@@ -245,7 +245,7 @@ class SageMakerTrainingOperatorAsync(SageMakerTrainingOperator):
                 self._check_if_job_exists()
             except TypeError:
                 # for apache-airflow-providers-amazon>=7.3.0
-                self.config["TrainingJobName"] = self._get_unique_job_name(
+                self.config["TrainingJobName"] = self._get_unique_job_name(  # type: ignore[attr-defined]
                     self.config["TrainingJobName"],
                     self.action_if_job_exists == "fail",
                     self.hook.describe_training_job,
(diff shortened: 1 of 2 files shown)


astronomer/providers/amazon/aws/operators/sagemaker.py
```diff
@@ -151,8 +151,17 @@ def execute(self, context: Context) -> None:  # type: ignore[override]
...
+            try:
+                # for apache-airflow-providers-amazon<=7.2.1
+                self._check_if_transform_job_exists()
+            except AttributeError:
+                # for apache-airflow-providers-amazon>=7.3.0
+                transform_config["TransformJobName"] = self._get_unique_job_name(  # type: ignore[attr-defined]
+                    transform_config["TransformJobName"],
+                    self.action_if_job_exists == "fail",
+                    self.hook.describe_transform_job,
+                )
```
pankajkoti, line 164: In my opinion, the code for latest provider version should be tried first and if it fails we should handle old versions in the except block. WDYT?

astronomer/providers/amazon/aws/operators/sagemaker.py
```diff
@@ -230,8 +239,17 @@ def execute(self, context: Context) -> None:  # type: ignore[override]
...
+            try:
+                # for apache-airflow-providers-amazon<=7.2.1
+                self._check_if_job_exists()
+            except TypeError:
+                # for apache-airflow-providers-amazon>=7.3.0
+                self.config["TrainingJobName"] = self._get_unique_job_name(  # type: ignore[attr-defined]
+                    self.config["TrainingJobName"],
+                    self.action_if_job_exists == "fail",
+                    self.hook.describe_training_job,
+                )
```
pankajkoti, line 252: same suggestion as above.

astronomer/providers/amazon/aws/operators/sagemaker.py
rajaths010494: till now we have used same convention https://github.com/astronomer/astronomer-providers/blob/main/astronomer/providers/amazon/aws/operators/emr.py#L46-L51, https://github.com/astronomer/astronomer-providers/blob/main/astronomer/providers/databricks/operators/databricks.py#L346-L351
so i maintained same way here also.

rajaths010494: have created a ticket to make these changes in other places also. https://github.com/astronomer/astronomer-providers/issues/913

rajaths010494: yeah added the ticket https://github.com/astronomer/astronomer-providers/issues/913

PR #918, opened 2023-03-10
Bump version to 1.16.0-dev

rajaths010494:
Since 1.15.1 is released, so bumped up the version to 1.16.0-dev

### DIFF
--- a/astronomer/providers/package.py
+++ b/astronomer/providers/package.py
@@ -8,7 +8,7 @@ def get_provider_info() -> Dict[str, Any]:
         "package-name": "astronomer-providers",
         "name": "Astronomer Providers",
         "description": "Apache Airflow Providers containing Deferrable Operators & Sensors from Astronomer",
-        "versions": "1.15.1",
+        "versions": "1.16.0.dev1",
         # Optional.
         "hook-class-names": [],
         "extra-links": [],
--- a/astronomer/providers/package.py
+++ b/astronomer/providers/package.py
@@ -8,7 +8,7 @@ def get_provider_info() -> Dict[str, Any]:
         "package-name": "astronomer-providers",
         "name": "Astronomer Providers",
         "description": "Apache Airflow Providers containing Deferrable Operators & Sensors from Astronomer",
-        "versions": "1.16.0.dev1",
+        "versions": "1.16.0-dev",
         # Optional.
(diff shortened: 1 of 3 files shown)


astronomer/providers/package.py
```diff
@@ -8,7 +8,7 @@ def get_provider_info() -> Dict[str, Any]:
         "package-name": "astronomer-providers",
         "name": "Astronomer Providers",
         "description": "Apache Airflow Providers containing Deferrable Operators & Sensors from Astronomer",
-        "versions": "1.15.1",
+        "versions": "1.16.0.dev1",
```
pankajkoti, line 11:
```suggestion
        "versions": "1.16.0dev",
```

docs/conf.py
```diff
@@ -27,7 +27,7 @@
 author = "Astronomer Inc."
 
 # The full version, including alpha/beta/rc tags
-release = "1.15.1"
+release = "1.16.0.dev1"
```
pankajkoti, line 30:
```suggestion
release = "1.16.0dev"
```

setup.cfg
```diff
@@ -1,6 +1,6 @@
 [metadata]
 name = astronomer-providers
-version = 1.15.1
+version = 1.16.0.dev1
```
pankajkoti, line 3:
```suggestion
version = 1.16.0dev
```

pankajkoti (commented):
I have not observed `.` between the patch version and suffixes. 
And since we won't have `dev1`, `dev2`, `dev3` out, just `dev` is sufficient. If we want to create RCs, they can be `rc1`, `rc2`, `rc3`, etc. e.g. `1.16.0rc1`

So I believe we can start following this convention unless we have other opinions and examples of projects with semantic versioning following such a pattern.

astronomer/providers/package.py
rajaths010494: Changed

docs/conf.py
rajaths010494: Changed

setup.cfg
rajaths010494: Changed

pankajkoti (approved):
Please also update the description of the PR as per the latest commit. Thanks for the bump!

PR #935, opened 2023-03-28
Add max attempts to Redshift Cluster dag to wait

rajaths010494:
Currently, when the dag was running by default max attempt was 5 and the interval period is 60 default. it will check for 5 mins and error out if the cluster is not brought up by then with the below error.
image

but in the subsequent retry when it tries to bring up the cluster and throws an error saying ClusterAlreadyExists error.
image

So this PR adds a max attempt of 30 which waits for a maximum of 30 mins to cluster to bring up.

### DIFF
--- a/astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
@@ -58,6 +58,7 @@ with DAG(
             {"Key": "Purpose", "Value": "ProviderTest"},
         ],
         wait_for_completion=True,
+        max_attempt=30,
         aws_conn_id=AWS_CONN_ID,
     )
 


astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
```diff
@@ -58,6 +58,7 @@
             {"Key": "Purpose", "Value": "ProviderTest"},
         ],
         wait_for_completion=True,
+        max_attempt=30,
```
phanikumv, line 61: is there any other way of waiting for the cluster (using boto3 waiters concept) than harcoding it?

rajaths010494: https://github.com/apache/airflow/blob/main/airflow/providers/amazon/aws/operators/redshift_cluster.py#L328-L334
 the OSS library uses waiters and we are sending value to the operator.

PR #945, opened 2023-04-07
Removed protobuf dependency

utkarsharma2:
Removed protobuf dependency

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -136,7 +136,7 @@ all =
     impyla
     openlineage-airflow>=0.12.0
     paramiko
-    protobuf<=3.20.0  # Bigquery provider isn't compatible with it. Details in https://github.com/apache/airflow/commit/25a9ae3b2eec85dfd500b0a921045fc95ab8ffd6
+    protobuf
 
 [options.packages.find]
 include =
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -64,7 +64,7 @@ google =
     apache-airflow-providers-google>=8.1.0
     gcloud-aio-storage
     gcloud-aio-bigquery
-    protobuf<=3.20.0  # Bigquery provider isn't compatible with it. Details in https://github.com/apache/airflow/commit/25a9ae3b2eec85dfd500b0a921045fc95ab8ffd6
+    protobuf  # Bigquery provider isn't compatible with it. Details in https://github.com/apache/airflow/commit/25a9ae3b2eec85dfd500b0a921045fc95ab8ffd6
```
pankajkoti, line 67:
```suggestion
    protobuf 
```

pankajkoti (commented):
Can we update the Title that we're not actually removing the dependency but removing the pin to an upper bound.
Also, would be nice to update the PR description on why we're removing the pin now and whether something has been fixed that was an issue earlier.

astronomer/providers/databricks/operators/new.py
Lee-W, line 1: May I know why we need this file in this PR? Thanks!

utkarsharma2:
@pankajkoti @Lee-W This PR was raised for debugging. We don't need this. Closing it.

Issue #949, opened 2023-04-10
Set TriggerDagRunOperator's  trigger_rule to all_done in master_dag


pankajkoti:
The master_dag has implemented chaining of DAGs by grouping them based on providers. However, if one of the DAG fails in the chain, subsequent DAGs are not run and hence we do not test them. As such, the DAGs do not need to depend on each other and hence we can set the `trigger_rule` in the `TriggerDagRunOperator` to `all_done` so that we test all the DAGs irrespective of whether the previous DAG in the chain succeeded.

PR #950, opened 2023-04-10
Set TriggerDagRunOperator's trigger_rule to all_done in master_dag

pankajkoti:
The `master_dag` has implemented chaining of DAGs by grouping them based on providers. However, if one of the DAG fails in the chain, subsequent DAGs are not run and hence we do not test them. As such, the DAGs do not need to depend on each other and hence we can set the `trigger_rule` in the `TriggerDagRunOperator` to `all_done` so that we test all the DAGs irrespective of whether the previous DAG in the chain succeeded.

closes: #949

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -96,6 +96,7 @@ repos:
         # Excluding Example DAGs temporarily until they can be fixed
         exclude: ^tests/.*\.py$|.*example_dags/.*
         args: ["-c", "./pyproject.toml"]
+        additional_dependencies: [".[toml]"]
 
   - repo: local
     hooks:
--- a/.circleci/integration-tests/master_dag.py
+++ b/.circleci/integration-tests/master_dag.py
@@ -78,6 +78,7 @@ def prepare_dag_dependency(task_info, execution_time):
                 reset_dag_run=True,
                 execution_date=execution_time,
                 allowed_states=["success", "failed", "skipped"],
+                trigger_rule="all_done",
             )
         )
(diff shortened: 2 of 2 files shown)


pankajkoti:
The `mypy` failures will be resolved once PR https://github.com/astronomer/astronomer-providers/pull/951 is merged and this branch is then rebased on the main.

.pre-commit-config.yaml
```diff
@@ -96,6 +96,7 @@ repos:
         # Excluding Example DAGs temporarily until they can be fixed
         exclude: ^tests/.*\.py$|.*example_dags/.*
         args: ["-c", "./pyproject.toml"]
+        additional_dependencies: [".[toml]"]
```
phanikumv, line 99: why this change?

pankajkoti: Somehow, I could no longer commit without this. I was getting this error:
```
pyproject.toml : toml parser not available, reinstall with toml extra
```

I followed the Stackoverflow answer https://stackoverflow.com/questions/75303927/pyproject-toml-toml-parser-not-available-reinstall-with-toml-extra

pankajastro: hmm, docs suggest something like https://bandit.readthedocs.io/en/latest/config.html?highlight=additional_dependencies#bandit-settings
```
additional_dependencies: ["bandit[toml]"]
```
not sure if it will work in this case

pankajastro (approved):

phanikumv (approved):

PR #954, opened 2023-04-11
fix(google): add back removed utility function get_gke_config_file

### LINKED ISSUE #947: Support breaking changes in Google provider release 9.0.0 for GKEStartPodOperatorAsync
PR https://github.com/apache/airflow/pull/29266/files has deprecated the method `get_gke_config_file()` and is out in the major release RC 9.0.0rc1. We need to support the change for provider version` >=9.0.0` (at the same time making sure that our operator works for provider versions `<9.0.0` too) as we're seeing the below error in `example_google_kubernetes_engine` DAG 

```
  File "/usr/local/lib/python3.9/site-packages/astronomer/providers/google/cloud/operators/kubernetes_engine.py", line 112, in execute
    self._get_or_create_pod(context)
  File "/usr/local/lib/python3.9/site-packages/astronomer/providers/google/cloud/operators/kubernetes_engine.py", line 95, in _get_or_create_pod
    with GKEStartPodOperator.get_gke_config_file(
TypeError: get_gke_config_file() got an unexpected keyword argument 'gcp_conn_id'
```

Lee-W:
Contradicted to the previous discussion with @pankajkoti . I dig into the code on both sides a bit. It seems to me that adding the removed function back makes more sense.

1. GKEStartPodOperatorAsync is not inherited from GKEStartPodOperator. They’re different operators from the beginning. If we’re to copy the logic from GKEStartPodOperator, it might make more sense for us to remove  GKEStartPodOperatorAsync  as GKEStartPodOperator now supports deferrable as well.
2. get_gke_config_file can be used alone and doesn’t necessary need to tide to a class which might make sense to make it a utility function

closes #947

### DIFF
--- a/astronomer/providers/google/cloud/__init__.py
+++ b/astronomer/providers/google/cloud/__init__.py
@@ -1,0 +1,81 @@
+from __future__ import annotations
+
+import os
+import tempfile
+from contextlib import contextmanager
+from typing import Generator, Sequence
+
+from airflow.exceptions import AirflowException
+from airflow.providers.google.common.hooks.base_google import GoogleBaseHook
+from airflow.utils.process_utils import execute_in_subprocess, patch_environ
+
+KUBE_CONFIG_ENV_VAR = "KUBECONFIG"
+
+
+@contextmanager
+def _get_gke_config_file(
+    gcp_conn_id: str,
+    project_id: str | None,
+    cluster_name: str,
+    impersonation_chain: str | Sequence[str] | None,
+    regional: bool,
+    location: str,
+    use_internal_ip: bool,
+) -> Generator[str, None, None]:
+    hook = GoogleBaseHook(gcp_conn_id=gcp_conn_id)
+    project_id = project_id or hook.project_id
+
+    if not project_id:
+        raise AirflowException(
+            "The project id must be passed either as "
+            "keyword project_id parameter or as project_id extra "
+            "in Google Cloud connection definition. Both are not set!"
+        )
+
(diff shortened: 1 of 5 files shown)


pankajastro:
> I am agnostic to this approach of copying the deleted method from Airflow OSS in the new release to our repository.

I agree with this.

pankajastro:
if someone will run `pip install astronomer-providers` then it would install the latest version of apache-airflow-providers-google. The problem will come if someone runs `pip install astronomer-providers` and then `pip install apache-airflow-providers-google<9.0.0` because pip will replace the new apache-airflow-providers-google with old.

IMO, it would be great if we can make it work with apache-airflow-providers-google==9.0.0 first. after that maybe we can the old version as a fallback like we are doing at a place 

I was never happy with `get_gke_config_file` since it depends on gcloud cli so would be great if we can move it out

rajaths010494:
> get_gke_config_file

I tried to get rid of the get_gke_config_file but had issues connecting to Google Kubernetes pods, which was unable to find or connect to run. So created a ticket here https://github.com/astronomer/astronomer-providers/issues/956 to spend more time and develop a proper solution.

pankajastro (approved):
LGTM. But create an issue to remove _get_gke_config_file ultimately

rajaths010494:
> LGTM. But create an issue to remove _get_gke_config_file ultimately



> LGTM. But create an issue to remove _get_gke_config_file ultimately

Yes i have created the issue https://github.com/astronomer/astronomer-providers/issues/956

PR #960, opened 2023-04-13
Increase `RedshiftCreateClusterSnapshotOperator` max attempt to wait for snapshot creation.

rajaths010494:
Currently, the RedshiftCreateClusterSnapshotOperator takes on avg of 4 mins to create the snapshot but sometimes it crosses 5 mins.  the default values in operator is for 15secs poll interval and 20 max_attempts and it fails with 
`botocore.exceptions.WaiterError: Waiter SnapshotAvailable failed: Max attempts exceeded`.

This PR increases the max attempt for the waiting of this operator to create the snapshot


pankajastro (approved):
LGTM, do we need to increase for other redshift operators too like RedshiftDeleteClusterSnapshotOperator?

Issue #958, opened 2023-04-12
SnowflakeOperatorAsync fails with % symbol


erdos2n:
**Describe the bug**
SnowflakeOperatorAsync is throwing a parsing error

**To Reproduce**
Use the SnowflakeOperatorAsync
Create a query using the `%b%` logic

will throw

```python
[2023-04-04T23:46:19.636+0000] {taskinstance.py:1902} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/astronomer/providers/snowflake/operators/snowflake.py", line 109, in execute
    hook.run(self.sql, autocommit=self.autocommit, parameters=self.parameters)  # type: ignore[arg-type]
  File "/usr/local/lib/python3.9/site-packages/astronomer/providers/snowflake/hooks/snowflake.py", line 71, in run
    cur.execute_async(sql_statement, parameters)
  File "/usr/local/lib/python3.9/site-packages/snowflake/connector/cursor.py", line 813, in execute_async
    return self.execute(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/snowflake/connector/cursor.py", line 680, in execute
    query = command % processed_params
ValueError: unsupported format character 'b' (0x62) at index 10952
```

**Expected behavior**
Query is parsed successfully

**Desktop (please complete the following information):**
 - OS: MacOS, Debian
 - Browser Chrome

**Additional context**
User has stated that this does not happen with their current custom operator that builds off of the airflow snowflakeoperator

rajaths010494:
Can i get an example query?

phanikumv:
Fixed in https://github.com/astronomer/astronomer-providers/pull/962

erdos2n:
User still saw this error in their task logs 

```
[2023-04-21, 02:44:03 UTC] {standard_task_runner.py:100} ERROR - Failed to execute job 2082653 for task 01_TEST_PERCENTAGE_OPERATOR (unsupported format character 'b' (0x62) at index 34; 43)
```

here are the libraries 

```
astronomer-providers                     1.15.3
apache-airflow-providers-snowflake       4.0.3
snowflake-connector-python               2.9.0
snowflake-sqlalchemy                     1.4.4
```

phanikumv:
The issue is on snowflake-sqlalchemy side.

I have a another github issue for the same
https://github.com/snowflakedb/snowflake-sqlalchemy/issues/408

[...]

PR #962, opened 2023-04-17
build(setup.cfg): pinning snowflake-sqlalchemy to greater than or eqaul to 1.4.4

Lee-W:
As discussed in #958, we might need to pin snowflake-sqlalchemy >= 1.4.4 to resolved the issue. For the next step, I'm thinking of sending a PR to pin it to >= 1.4.4 on `apache-airflow-providers-snowflake` and then pin the `apache-airflow-providers-snowflake` in this repo after it's released

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -126,7 +126,6 @@ all =
     apache-airflow-providers-google>=8.1.0
     apache-airflow-providers-http
     apache-airflow-providers-snowflake
-    snowflake-sqlalchemy>=1.4.4  # Temporary solution for https://github.com/astronomer/astronomer-providers/issues/958, we should pin apache-airflow-providers-snowflake version after it pins this package to great than or equal to 1.4.4.
     apache-airflow-providers-sftp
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -125,6 +126,7 @@ all =
     apache-airflow-providers-google>=8.1.0
     apache-airflow-providers-http
     apache-airflow-providers-snowflake
+    snowflake-sqlalchemy>=1.4.4  # Temporary solution for https://github.com/astronomer/astronomer-providers/issues/958, we should pin apache-airflow-providers-snowflake version after it pins this package to great than or equal to 1.4.4.
```
pankajkoti, line 129: Can we move this line beneath the `protobuf` dependency as the order in this `all` section looks like to be alphabetically sorted?

pankajkoti (approved):
Looks good, have a minor comment if you could address before merging, please.

setup.cfg
Lee-W: sure. just update it!

Issue #906, opened 2023-03-01
SFTPSensorAsync to start supporting host key


magdagultekin:
**Is your feature request related to a problem? Please describe.**
`SFTPSensorAsync` currently does not support host keys (whereas `SFTPSensor` does) which prevent against man in the middle attacks.

**Describe the solution you'd like**
I would like the same support for all types of authorization like in the traditional operator.

pankajkoti:
I reached out to @magdagultekin and @rajaths010494 earlier today to understand the issue more. Looks like Magda created this issue post a discussion with Rajath during a customer interaction. We do not have an example DAG showing exactly what works with the Sync version and what does not work with Async version. 

I started understanding more about host keys wrt SFTP. Currently, I am trying to use the Sync sensor using host key of an EC2 instance which I got by turning on an instance and running `ssh-keyscan <ec2-public-ip>`. However, setting the got host keys in the sftp connection do not seem to help as of now and the DAG is failing. I am continuing to get the DAG working by trying various combinations of the host keys. Will update here more if I could progress further.

PR #963, opened 2023-04-17
Support host key verification for SFTPSensorAsync

pankajkoti:
Since, the AsyncSSH client does not support host keys other than the ones in known_hosts file, we create and use a paramiko client to verify the provided host key. If no host key is given, we issue a warning in the log that it is susceptible to man in the middle attacks.

closes: #906

### DIFF
--- a/astronomer/providers/sftp/hooks/sftp.py
+++ b/astronomer/providers/sftp/hooks/sftp.py
@@ -1,11 +1,17 @@
+from __future__ import annotations
+
+from base64 import decodebytes
 from datetime import datetime
 from fnmatch import fnmatch
-from typing import List, Optional
+from typing import Any
 
 import asyncssh
+import paramiko
 from airflow.exceptions import AirflowException
 from airflow.hooks.base import BaseHook
 from asgiref.sync import sync_to_async
+from paramiko.config import SSH_PORT
+from tenacity import Retrying, stop_after_attempt, wait_fixed, wait_random
 
 
 class SFTPHookAsync(BaseHook):
@@ -30,6 +36,13 @@ class SFTPHookAsync(BaseHook):
     hook_name = "SFTP"
     default_known_hosts = "none"
 
+    _host_key_mappings = {
+        "rsa": paramiko.RSAKey,
+        "dss": paramiko.DSSKey,
+        "ecdsa": paramiko.ECDSAKey,
+        "ed25519": paramiko.Ed25519Key,
+    }
+
     def __init__(  # nosec: B107
         self,
         sftp_conn_id: str = default_conn_name,
@@ -51,6 +64,8 @@ class SFTPHookAsync(BaseHook):
         self.key_file = key_file
         self.passphrase = passphrase
         self.private_key = private_key
+        self.no_host_key_check = True
+        self.host_key = None
 
     async def _get_conn(self) -> asyncssh.SSHClientConnection:
         """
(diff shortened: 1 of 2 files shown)


astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -1,11 +1,17 @@
...
-from typing import List, Optional
+from typing import Any
 
 import asyncssh
+import paramiko
 from airflow.exceptions import AirflowException
 from airflow.hooks.base import BaseHook
 from asgiref.sync import sync_to_async
+from paramiko.config import SSH_PORT
+from tenacity import Retrying, stop_after_attempt, wait_fixed, wait_random
```
Lee-W, line 14: Do we need to add it as a dependency in `setup.cfg`?

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -81,6 +96,69 @@ async def _get_conn(self) -> asyncssh.SSHClientConnection:
                 if "private_key" in extra_options:
                     self.private_key = extra_options.get("private_key")
 
+                host_key = extra_options.get("host_key")
+                no_host_key_check = extra_options.get("no_host_key_check")
+
+                if no_host_key_check is not None:
```
Lee-W, line 102: Should we use `if not no_host_key_check`? I was suggested to write in this way by TP when I sent a PR to OSS airflow

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -81,6 +96,69 @@ async def _get_conn(self) -> asyncssh.SSHClientConnection:
...
+                    no_host_key_check = str(no_host_key_check).lower() == "true"
+                    if host_key is not None and no_host_key_check:
```
Lee-W, line 104: nitpick: Should we use `if host_key and no_host_key_check`? Or is there case such check won't work

astronomer/providers/sftp/hooks/sftp.py
pankajkoti: I have the same opinion in general, but I guess here the idea is to process the value in case if it set to anything in the extras. I have copied the logic from the SSH hook https://github.com/apache/airflow/blob/88f9c0c41effef13c1014afeef4caa0481c91d9f/airflow/providers/ssh/hooks/ssh.py#L195

astronomer/providers/sftp/hooks/sftp.py
pankajkoti: same comment as above

astronomer/providers/sftp/hooks/sftp.py
Lee-W: got it 🙂

astronomer/providers/sftp/hooks/sftp.py
Lee-W: got it 🙂

Lee-W (commented):
CI failed. Do we need to address that?

pankajkoti:
> CI failed. Do we need to address that?

yes, have added tests to cover the patch. Thanks

pankajkoti:
Logs when host key has incorrect padding
Screenshot 2023-04-19 at 7 52 12 PM

Logs when host key does not match as expected by the server host configured
Screenshot 2023-04-19 at 7 51 55 PM

Successful run with correct host_key
Screenshot 2023-04-19 at 8 03 33 PM

Async sensor triggerer logs showing warning while sensing in deferral when no host key configured
Screenshot 2023-04-19 at 8 04 46 PM

Lee-W (approved):

Issue #966, opened 2023-04-19
'BigQueryInsertJobOperatorAsync' object has no attribute 'poll_interval'


m1racoli:
**Describe the bug**
When running BigQueryInsertJobOperatorAsync with astronomer-providers 1.15.3 we get the following error:

```
[2023-04-19, 14:08:36 UTC] {taskinstance.py:1090} INFO - Dependencies all met for dep_context=non-requeueable deps ti=<TaskInstance: test_bigquery.test_query_async manual__2023-04-19T14:08:34.235897+00:00 [queued]>
[2023-04-19, 14:08:36 UTC] {taskinstance.py:1090} INFO - Dependencies all met for dep_context=requeueable deps ti=<TaskInstance: test_bigquery.test_query_async manual__2023-04-19T14:08:34.235897+00:00 [queued]>
[2023-04-19, 14:08:36 UTC] {taskinstance.py:1288} INFO - 
--------------------------------------------------------------------------------
[2023-04-19, 14:08:36 UTC] {taskinstance.py:1289} INFO - Starting attempt 1 of 1
[2023-04-19, 14:08:36 UTC] {taskinstance.py:1290} INFO - 
--------------------------------------------------------------------------------
[2023-04-19, 14:08:36 UTC] {taskinstance.py:1309} INFO - Executing <Task(BigQueryInsertJobOperatorAsync): test_query_async> on 2023-04-19 14:08:34.235897+00:00
[2023-04-19, 14:08:36 UTC] {standard_task_runner.py:55} INFO - Started process 276 to run task
[2023-04-19, 14:08:36 UTC] {standard_task_runner.py:82} INFO - Running: ['airflow', 'tasks', 'run', 'test_bigquery', 'test_query_async', 'manual__2023-04-19T14:08:34.235897+00:00', '--job-id', '8', '--raw', '--subdir', 'DAGS_FOLDER/test_bigquery.py', '--cfg-path', '/tmp/tmpzwbdj_mc']
[2023-04-19, 14:08:36 UTC] {standard_task_runner.py:83} INFO - Job 8: Subtask test_query_async
[2023-04-19, 14:08:36 UTC] {task_command.py:389} INFO - Running <TaskInstance: test_bigquery.test_query_async manual__2023-04-19T14:08:34.235897+00:00 [running]> on host 3c2825c2be74
[2023-04-19, 14:08:36 UTC] {taskinstance.py:1516} INFO - Exporting the following env vars:
[...]

PR #967, opened 2023-04-19
Bigquery operator backward compatibility

pankajastro:
Bigquery operator backward compatibility
currently, the task fails if we have astronomer provider 1.15.3 and apache-airflow-providers-google < 8.11.0

Add init in affected operators and if `poll_interval` is pass by user then use it otherwise use default trigger poll_interval

with new google provider
Screenshot 2023-04-19 at 10 13 13 PM

with a old google provider

Screenshot 2023-04-19 at 10 14 07 PM

closes: https://github.com/astronomer/astronomer-providers/issues/966

### DIFF
--- a/astronomer/providers/google/cloud/operators/bigquery.py
+++ b/astronomer/providers/google/cloud/operators/bigquery.py
@@ -72,7 +72,7 @@ class BigQueryInsertJobOperatorAsync(BigQueryInsertJobOperator, BaseOperator):
     :param poll_interval: polling period in seconds to check for the status of job. Defaults to 4 seconds.
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.poll_interval = kwargs.pop("poll_interval", None)
         super().__init__(*args, **kwargs)
 
@@ -152,7 +152,7 @@ class BigQueryCheckOperatorAsync(BigQueryCheckOperator):
     for the status in async mode by using the job id
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.poll_interval = kwargs.pop("poll_interval", None)
         super().__init__(*args, **kwargs)
 
@@ -258,7 +258,7 @@ class BigQueryGetDataOperatorAsync(BigQueryGetDataOperator):
     :param poll_interval: polling period in seconds to check for the status of job. Defaults to 4 seconds
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.poll_interval = kwargs.pop("poll_interval", None)
         super().__init__(*args, **kwargs)
 
@@ -365,7 +365,7 @@ class BigQueryIntervalCheckOperatorAsync(BigQueryIntervalCheckOperator):
     :param poll_interval: polling period in seconds to check for the status of job. Defaults to 4 seconds.
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.poll_interval = kwargs.pop("poll_interval", None)
         super().__init__(*args, **kwargs)
 
@@ -432,7 +432,7 @@ class BigQueryIntervalCheckOperatorAsync(BigQueryIntervalCheckOperator):
 
 
 class BigQueryValueCheckOperatorAsync(BigQueryValueCheckOperator):  # noqa: D101
-    def __init__(self, *args, **kwargs):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
         self.poll_interval = kwargs.pop("poll_interval", None)
         super().__init__(*args, **kwargs)
 
--- a/astronomer/providers/google/cloud/operators/bigquery.py
+++ b/astronomer/providers/google/cloud/operators/bigquery.py
@@ -73,7 +73,7 @@ class BigQueryInsertJobOperatorAsync(BigQueryInsertJobOperator, BaseOperator):
     """
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
-        self.poll_interval = kwargs.pop("poll_interval", None)
+        self.poll_interval: float = kwargs.pop("poll_interval", 4.0)
         super().__init__(*args, **kwargs)
 
     def execute(self, context: Context) -> None:  # noqa: D102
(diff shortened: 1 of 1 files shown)


astronomer/providers/google/cloud/operators/bigquery.py
```diff
@@ -72,6 +72,10 @@ class BigQueryInsertJobOperatorAsync(BigQueryInsertJobOperator, BaseOperator):
     :param poll_interval: polling period in seconds to check for the status of job. Defaults to 4 seconds.
     """
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
```
utkarsharma2, line 75: @pankajastro All of these operator uses the same trigger?

pankajastro: No, but we introduced dynamic poll_interval in these all operator/trigger

utkarsharma2: I think we should add tests for these operators, just to make there is not regression. WDYT?

astronomer/providers/google/cloud/operators/bigquery.py
```diff
@@ -72,6 +72,10 @@ class BigQueryInsertJobOperatorAsync(BigQueryInsertJobOperator, BaseOperator):
...
+        self.poll_interval = kwargs.pop("poll_interval", None)
```
pankajkoti, line 76:
```suggestion
        self.poll_interval: float = kwargs.pop("poll_interval", 4.0)
```
The default value seems to be 4.0 in the OSS provider. I guess we should have the default value here too and across other places in the PR.

astronomer/providers/google/cloud/operators/bigquery.py
```diff
@@ -148,6 +152,10 @@ class BigQueryCheckOperatorAsync(BigQueryCheckOperator):
     for the status in async mode by using the job id
     """
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        self.poll_interval = kwargs.pop("poll_interval", None)
```
pankajkoti, line 156:
```suggestion
        self.poll_interval: float = kwargs.pop("poll_interval", 4.0)
```

astronomer/providers/google/cloud/operators/bigquery.py
```diff
@@ -250,6 +258,10 @@ class BigQueryGetDataOperatorAsync(BigQueryGetDataOperator):
     :param poll_interval: polling period in seconds to check for the status of job. Defaults to 4 seconds
     """
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        self.poll_interval = kwargs.pop("poll_interval", None)
```
pankajkoti, line 262:
```suggestion
        self.poll_interval: float = kwargs.pop("poll_interval", 4.0)
```

astronomer/providers/google/cloud/operators/bigquery.py
```diff
@@ -353,6 +365,10 @@ class BigQueryIntervalCheckOperatorAsync(BigQueryIntervalCheckOperator):
     :param poll_interval: polling period in seconds to check for the status of job. Defaults to 4 seconds.
     """
 
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        self.poll_interval = kwargs.pop("poll_interval", None)
```
pankajkoti, line 369:
```suggestion
        self.poll_interval: float = kwargs.pop("poll_interval", 4.0)
```

astronomer/providers/google/cloud/operators/bigquery.py
```diff
@@ -416,6 +432,10 @@ def execute_complete(self, context: Context, event: Dict[str, Any]) -> None:
 
 
 class BigQueryValueCheckOperatorAsync(BigQueryValueCheckOperator):  # noqa: D101
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        self.poll_interval = kwargs.pop("poll_interval", None)
```
pankajkoti, line 436:
```suggestion
        self.poll_interval: float = kwargs.pop("poll_interval", 4.0)
```

phanikumv (commented):
can you share the screenshot of the tested scenarios on this operator? for both latest release and earlier release

astronomer/providers/google/cloud/operators/bigquery.py
phanikumv: we need to test for old version as well

pankajkoti (commented):
Also, I guess whatever value the user gives for the `poll_interval`, it will be replaced by the super class `init` with the default value since we are calling init afterwards, no? 

Instead of 
```
self.poll_interval = kwargs.pop("poll_interval", None)
super().__init__(*args, **kwargs)
```

Do we need to do something like this?
```
poll_interval: float = kwargs.pop("poll_interval", 4.0)
super().__init__(*args, **kwargs)
self.poll_interval = poll_interval

```

Attaching my trial in IPython
Screenshot 2023-04-19 at 10 17 58 PM

pankajastro:
> can you share the screenshot of the tested scenarios on this operator? for both latest release and earlier release

added in description

astronomer/providers/google/cloud/operators/bigquery.py
phanikumv: what about the test result for `apache-airflow-providers-google` < 8.11.0

astronomer/providers/google/cloud/operators/bigquery.py
utkarsharma2: > Also, I guess whatever value the user gives for the `poll_interval`, it will be replaced by the super class `init` with the default value since we are calling init afterwards, no?
> 
> Instead of
> 
> ```
> self.poll_interval = kwargs.pop("poll_interval", None)
> super().__init__(*args, **kwargs)
> ```
> 
> Do we need to do something like this?
> 
> ```
> poll_interval: float = kwargs.pop("poll_interval", 4.0)
> super().__init__(*args, **kwargs)
> self.poll_interval = poll_interval
> ```
> 
> Attaching my trial in IPython Screenshot 2023-04-19 at 10 17 58 PM

@pankajkoti 

```
super().__init__(*args, **kwargs)
self.poll_interval = poll_interval
```
^^ This will fail for the old provider version since it doesn't expect the `poll_interval` No?

astronomer/providers/google/cloud/operators/bigquery.py
pankajastro: unfortunately, we run tests against only one version of oss provider maybe @phanikumv if you want we can try run for more version in a separate PR

phanikumv: yeah just want to make sure that we don't break things this time

pankajkoti: locally you could just pin the provider to 8.11.0 and give the DAG a run, right? 

Also, In the description, I see 8.12.0 and 9.0.0, but anything >=8.12 will work with 1.15.3, we need to test one for <8.12. Can we update the description snapshot for the old release to reflect 8.11.0 instead of 9.0.0

pankajkoti (approved):
Looks good to me. Would be nice to update the description screenshot after testing for Google provider <8.12

PR #971, opened 2023-04-19
Update release docs updating version section

pankajastro:
We also need to update version in ``astronomer/providers/package.py`` so lets document that

add `commitizen` in pyproject.toml for bumping version

### DIFF
--- a/RELEASING.rst
+++ b/RELEASING.rst
@@ -21,7 +21,9 @@ You will need to update it in:
 
 * ``setup.cfg`` - in ``version`` key of ``[metadata]`` section
 * ``docs/conf.py`` - ``release`` variable
-* ``astronomer/providers/package.py`` - ``get_provider_info`` method return value ``version``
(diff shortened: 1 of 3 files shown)


RELEASING.rst
```diff
@@ -21,6 +21,7 @@ You will need to update it in:
 
 * ``setup.cfg`` - in ``version`` key of ``[metadata]`` section
 * ``docs/conf.py`` - ``release`` variable
+* ``astronomer/providers/package.py`` - ``get_provider_info`` method return value ``version``
```
pankajkoti, line 24:
```suggestion
* ``astronomer/providers/package.py`` - ``get_provider_info`` method return value ``versions``
```

Lee-W (commented):
If we have multiple places that we need to update our versions, not sure whether it makes more sense to use tools to update them all at once. https://github.com/commitizen-tools/commitizen is the tools I used to do such thing.

phanikumv (approved):

PR #976, opened 2023-04-24
SnowflakeOperatorAsync: check whether all queries finish before deferring the task

Lee-W:
Closes: #982

### DIFF
--- a/astronomer/providers/snowflake/operators/snowflake.py
+++ b/astronomer/providers/snowflake/operators/snowflake.py
@@ -1,10 +1,15 @@
 from __future__ import annotations
 
+import logging
 import typing
+from contextlib import closing
 from datetime import timedelta
 from typing import Any, Callable, List
 
 from airflow.exceptions import AirflowException
+
+from snowflake.connector import SnowflakeConnection
+from snowflake.connector.constants import QueryStatus
 
 try:
     from airflow.providers.snowflake.operators.snowflake import SnowflakeOperator
@@ -29,6 +34,14 @@ from astronomer.providers.snowflake.triggers.snowflake_trigger import (
     get_db_hook,
 )
 from astronomer.providers.utils.typing_compat import Context
+
+
+def check_queries_success(conn: SnowflakeConnection, query_ids: list[str]) -> bool:
+    with closing(conn) as conn:
+        succeeded_before_defer = all(
+            [conn.get_query_status(query_id) == QueryStatus.SUCCESS for query_id in query_ids]
+        )
+    return succeeded_before_defer
 
 
 class SnowflakeOperatorAsync(SnowflakeOperator):
@@ -169,16 +182,20 @@ class SnowflakeOperatorAsync(SnowflakeOperator):
         if self.do_xcom_push:
             context["ti"].xcom_push(key="query_ids", value=self.query_ids)
 
-        self.defer(
-            timeout=self.execution_timeout,
-            trigger=SnowflakeTrigger(
-                task_id=self.task_id,
-                poll_interval=self.poll_interval,
-                query_ids=self.query_ids,
-                snowflake_conn_id=self.snowflake_conn_id,
-            ),
-            method_name="execute_complete",
-        )
+        if not check_queries_success(hook.get_conn(), self.query_ids):
+            logging.info("Task deferred")
+            self.defer(
+                timeout=self.execution_timeout,
+                trigger=SnowflakeTrigger(
+                    task_id=self.task_id,
+                    poll_interval=self.poll_interval,
+                    query_ids=self.query_ids,
+                    snowflake_conn_id=self.snowflake_conn_id,
+                ),
+                method_name="execute_complete",
+            )
+        else:
+            logging.info("Queries finish before deferred")
 
     def execute_complete(self, context: Context, event: dict[str, str | list[str]] | None = None) -> Any:
         """
--- a/astronomer/providers/snowflake/operators/snowflake.py
+++ b/astronomer/providers/snowflake/operators/snowflake.py
@@ -39,6 +39,7 @@ from astronomer.providers.utils.typing_compat import Context
 
 
 def _check_queries_finish(conn: SnowflakeConnection, query_ids: list[str]) -> bool:
+    """Check whether snowflake queries finish (in aborting, failed_with_error or success)"""
     with closing(conn) as conn:
         for query_id in query_ids:
             status = conn.get_query_status(query_id)
--- a/astronomer/providers/snowflake/operators/snowflake.py
+++ b/astronomer/providers/snowflake/operators/snowflake.py
@@ -49,8 +49,17 @@ def _check_queries_finish(conn: SnowflakeConnection, query_ids: list[str]) -> bo
                 raise AirflowException(f"FAILED_WITH_ERROR {FAILED_WITH_ERROR_MESSAGE}")
             elif status == QueryStatus.SUCCESS:
                 pass
-            else:
+            elif status in (
+                QueryStatus.RUNNING,
+                QueryStatus.QUEUED,
+                QueryStatus.DISCONNECTED,
+                QueryStatus.RESUMING_WAREHOUSE,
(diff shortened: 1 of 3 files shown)


astronomer/providers/snowflake/operators/snowflake.py
```diff
@@ -31,6 +38,23 @@
 from astronomer.providers.utils.typing_compat import Context
 
 
+def _check_queries_finish(conn: SnowflakeConnection, query_ids: list[str]) -> bool:
```
sunank200, line 41: doc-string in this class would be useful

astronomer/providers/snowflake/operators/snowflake.py
```diff
@@ -31,6 +38,23 @@
...
+        for query_id in query_ids:
+            status = conn.get_query_status(query_id)
+
+            if status == QueryStatus.ABORTING:
+                raise AirflowException(f"ABORTING {ABORTING_MESSAGE}")
+            elif status == QueryStatus.FAILED_WITH_ERROR:
+                raise AirflowException(f"FAILED_WITH_ERROR {FAILED_WITH_ERROR_MESSAGE}")
+            elif status == QueryStatus.SUCCESS:
+                pass
+            else:
```
sunank200, line 52: What is other query statuses? it would be good to log in with warning before returning False. WDYT?

pankajkoti:
@Lee-W can you please create a ticket for this and link in the description. Also, please tag the milestone 1.16.0 to the the issue.

astronomer/providers/snowflake/operators/snowflake.py
Lee-W: Sure!

astronomer/providers/snowflake/operators/snowflake.py
Lee-W: I think `warning` might not be the best solution here. Because even if it returns `false`, it just means we're going to defer the task. But for better readability, I'll make the following changes 

```suggestion
            elif (
                    QueryStatus.RUNNING,
                    QueryStatus.QUEUED,
                    QueryStatus.DISCONNECTED,
                    QueryStatus.RESUMING_WAREHOUSE,
                    QueryStatus.BLOCKED,
                    QueryStatus.NO_DATA,
            ):
                return False
            else:
                raise ValueErorr(f"Unexpected QueryStatus: str(status)")
```

phanikumv (commented):
Can you test this on kubernetes executor..?

Lee-W:
> @Lee-W can you please create a ticket for this and link in the description. Also, please tag the milestone 1.16.0 to the the issue.

@pankajkoti  Yep, sure! I already create one #982 

> Can you test this on kubernetes executor..?

@phanikumv Sure.

> Hi @Lee-W , did we get a chance this PR for both Celery and Kubernetes executor and do we also need to test any backward compatibility scenario? The reason I'm asking is because we want to check if we can include this in today's release of Providers for the version 1.15.5

Yes, I'll test in on both celery and Kubernetes executors. Currently, I can't think of any backward compatibility issue on this change

Lee-W:
## Successful DAG run with celery executor

image

Working on KE

Lee-W:
@phanikumv Sure. I've tested with your DAG


image

and the original example DAG

image

astronomer/providers/snowflake/operators/snowflake.py
```diff
@@ -31,6 +38,33 @@
...
 
 
+def _check_queries_finish(conn: SnowflakeConnection, query_ids: list[str]) -> bool:
+    """Check whether snowflake queries finish (in aborting, failed_with_error or success)"""
+    with closing(conn) as conn:
+        for query_id in query_ids:
+            status = conn.get_query_status(query_id)
+
+            if status == QueryStatus.ABORTING:
+                raise AirflowException(f"ABORTING {ABORTING_MESSAGE}")
```
pankajkoti, line 48: It might help the user if we also log the query ID in the exception.

astronomer/providers/snowflake/operators/snowflake.py
```diff
@@ -31,6 +38,33 @@
...
+            elif status == QueryStatus.FAILED_WITH_ERROR:
+                raise AirflowException(f"FAILED_WITH_ERROR {FAILED_WITH_ERROR_MESSAGE}")
```
pankajkoti, line 50: Same suggestion for logging the query ID

astronomer/providers/snowflake/operators/snowflake.py
```diff
@@ -31,6 +38,33 @@
...
+            elif status == QueryStatus.SUCCESS:
+                pass
+            elif status in (
+                QueryStatus.RUNNING,
+                QueryStatus.QUEUED,
+                QueryStatus.DISCONNECTED,
```
pankajkoti, line 56: Should QueryStatus.DISCONNECTED also qualify for an exception? Is Snowflake automatically able to resolve disconnected queries? If not, we might keep on polling for it until triggerer times out.

astronomer/providers/snowflake/operators/snowflake.py
```diff
@@ -31,6 +38,33 @@
...
+                QueryStatus.RESUMING_WAREHOUSE,
+                QueryStatus.BLOCKED,
+                QueryStatus.NO_DATA,
+            ):
+                return False
+            else:
+                raise ValueError(f"Unexpected QueryStatus: {str(status)}")
```
pankajkoti, line 63: There is another status FAILED_WITH_INCIDENT, I guess we could also raise an exception for it.

pankajkoti (commented):
Looks good to me mostly, have a minor suggestions on handling the query statuses

Lee-W:
Thanks for @pankajastro 's help! It works fine on astro cloud
image

astronomer/providers/snowflake/operators/snowflake.py
Lee-W: It doesn't seem to be supported by [snowflake-python-connector](https://docs.snowflake.com/developer-guide/python-connector/python-connector-api#enums)

astronomer/providers/snowflake/operators/snowflake.py
Lee-W: I kinda hesitate on `QueryStatus.DISCONNECTED`. According to [doc](https://docs.snowflake.com/developer-guide/python-connector/python-connector-api#enums), `The session’s connection is broken. The query’s state will change to “FAILED_WITH_ERROR” soon.`, it's going to failed but just not yet failed.

astronomer/providers/snowflake/operators/snowflake.py
Lee-W: Agree. I was meant to match the original message. I just updated both places. Thanks!

astronomer/providers/snowflake/operators/snowflake.py
Lee-W: Updated. Thanks!

astronomer/providers/snowflake/operators/snowflake.py
pankajkoti: Makes sense, but if it is anyway going to change to Failed with error then can we not raise an early exception? 

I am happy with whatever you feel is the best. Just had this thought.

astronomer/providers/snowflake/operators/snowflake.py
pankajkoti: Looks like that document is missing it. But it does seem to have it 
Screenshot 2023-04-25 at 6 45 34 PM

Lee-W: You're right. I should have looked into the source code. Doc could be misleading. Thanks! Just added it

astronomer/providers/snowflake/operators/snowflake.py
Lee-W: After a second thought, I think your idea is better. After I applied your suggestion, I added a comment there cause it might still be a bit confusing.

phanikumv (approved):

Issue #977, opened 2023-04-24
Test Airflow Providers prepared on April 21, 2023


pankajkoti:
Test the following RCs

1. https://pypi.org/project/apache-airflow-providers-amazon/8.0.0rc1/
2. https://pypi.org/project/apache-airflow-providers-cncf-kubernetes/6.1.0rc1/
3. https://pypi.org/project/apache-airflow-providers-databricks/4.1.0rc1/
4. https://pypi.org/project/apache-airflow-providers-dbt-cloud/3.1.1rc1/
5. https://pypi.org/project/apache-airflow-providers-google/10.0.0rc1/
6. https://pypi.org/project/apache-airflow-providers-microsoft-azure/6.0.0rc1/
7. https://pypi.org/project/apache-airflow-providers-snowflake/4.0.5rc1/

Test that mypy checks, unit tests and example DAGs run fine and make necessary implementation changes to make the operators/sensors future ready for the providers release and ensuring backward compatibility for the older versions.

pankajastro:
@pankajkoti can we close this?

pankajkoti:
yes, thanks for the check @pankajastro, the corresponding PR was marked closed and we have tested the RCs. Closing it now.

PR #978, opened 2023-04-24
Fix Batch operator async for Amazon provider release 8.0.0

pankajkoti:
The `overrides` param has been removed and should instead `container_overrides` parameter be used with the release of provider version 8.0.0. The commit adds this fix ensuring backward compatibility for older versions of the provider to continue using the `overrrides` param.

related to: #977


astronomer/providers/amazon/aws/operators/batch.py
```diff
@@ -27,12 +27,13 @@ class BatchOperatorAsync(BatchOperator):
     :param job_name: the name for the job that will run on AWS Batch (templated)
     :param job_definition: the job definition name on AWS Batch
     :param job_queue: the queue name on AWS Batch
-    :param overrides: the `containerOverrides` parameter for boto3 (templated)
+    :param overrides: DEPRECATED and removed in 8.0.0, use container_overrides instead with the same value.
```
pankajastro, line 30: `DEPRECATED and removed in 8.0.0 ` might be confusing for the reader here because the user may or may not know that 8.0.0 we are referring for oss version

pankajastro (approved):

astronomer/providers/amazon/aws/operators/batch.py
pankajkoti, line 30:
```suggestion
    :param overrides: Removed in apache-airflow-providers-amazon release 8.0.0, use container_overrides instead with the same value.
```

pankajkoti:
**Successful DAG runs for older versions of the providers**
Screenshot 2023-04-24 at 6 21 04 PM


**Successful DAG runs for RC versions of the providers**
Screenshot 2023-04-24 at 6 21 24 PM

phanikumv (approved):

PR #986, opened 2023-04-25
Reorder master DAG for quicker report

pankajkoti:
The master DAG run currently takes about ~1 hour to execute all the DAGS.
I have observed that some of the long-running DAGs like Redshift cluster management, Sagemaker, 
EMR EKS PI job, etc. are run later in their groups. This PR re-order the DAG groups to run
the long-running DAGs first so that we reduce the overall execution time for the master DAG.
The duration was down to **37 mins** from **1-hour** with this minimum re-ordering.


pankajkoti:
Previous Gannt chart
Screenshot 2023-04-25 at 1 59 43 PM


New Gannt chart
Screenshot 2023-04-25 at 1 59 59 PM

phanikumv (approved):

PR #991, opened 2023-04-26
ci(github-actions): make deploy workflow load input from the github secrets

Lee-W:
The current solution cannot load variable from inputs and make it secret. It only makes it empty [1]. It seems there's no valid solution to mask the input. (There's some workarounds but can be easily misused). Thus, I think we should load the secrets from github secrets. If we agree on this approach, as I don't have the permission, I'd kindly ask repo owner/maintainer to help me setup the following secrets.

* DOCKER_REGISTRY
* ORGANIZATION_ID
* PROVIDER_INTEGRATION_TESTS_DEPLOYMENT_ID
* PROVIDER_INTEGRATION_TESTS_ASTRONOMER_KEY_ID
* PROVIDER_INTEGRATION_TESTS_ASTRONOMER_KEY_SECRET
* PROVIDER_INTEGRATION_TESTS_ON_KE_DEPLOYMENT_ID
* PROVIDER_INTEGRATION_TESTS_ON_KE_ASTRONOMER_KEY_ID
* PROVIDER_INTEGRATION_TESTS_ON_KE_ASTRONOMER_KEY_SECRET

We can also make `DOCKER_REGISTRY` and `ORGANIZATION_ID` separate values for both environments if that's desired. (i.e., `PROVIDER_INTEGRATION_TESTS_DOCKER_REGISTRY` and `PROVIDER_INTEGRATION_TESTS_ON_KE_DOCKER_REGISTRY` )

Another thing I was thinking of is adding another set of custom values (e.g., `CUSTOM_DEPLOYMENT_ID`), but only who has the permission to set the secrets can use this feature. I'm not sure whether we want it

[1] https://github.com/astronomer/astronomer-providers/actions/runs/4803682040


pankajkoti (commented):
Makes sense. Unfortunately, I do not have permissions to set secrets too. @phanikumv might be able to help here.

Additionally, do you think if it would make sense to have the runtime image as an arg in the Dockerfile.astro_cloud (https://github.com/astronomer/astronomer-providers/blob/main/.circleci/integration-tests/Dockerfile.astro_cloud#L2) and pass it as a variable from the workflow_dispatch?

Lee-W:
In the latest version, I add an option "both" to deploy integration tests to both environment

image

pankajkoti (approved):
Looks amazing to me to merge when we have tested this on Astro. Thanks a lot @Lee-W for the quick efforts here 🎉

Lee-W:
image
It's now deploying.  🎉

https://github.com/astronomer/astronomer-providers/actions/runs/4807498142/jobs/8556340083
https://github.com/astronomer/astronomer-providers/actions/runs/4807477983/jobs/8556295791

Lee-W:
image
It succeeded 🎉.

pankajkoti:
Kudos to you @Lee-W 👏🏽 This is great automation!!

PR #992, opened 2023-04-26
test(snowflake): extend snowflake example dag to run multiple quries in one run

Lee-W:
as title

### DIFF
--- a/astronomer/providers/snowflake/example_dags/example_snowflake.py
+++ b/astronomer/providers/snowflake/example_dags/example_snowflake.py
@@ -24,6 +24,12 @@ SNOWFLAKE_SLACK_MESSAGE = (
     "Results in an ASCII table:\n```{{ results_df | tabulate(tablefmt='pretty', headers='keys') }}```"
 )
 
+MULTIPLE_QUERY_IN_ONE_RUN = (
+    "CREATE OR REPLACE TRANSIENT TABLE sample_table (name VARCHAR(250), id INT);"
+    "INSERT INTO sample_table VALUES ('name', 1);"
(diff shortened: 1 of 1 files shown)


astronomer/providers/snowflake/example_dags/example_snowflake.py
```diff
@@ -24,6 +24,12 @@
     "Results in an ASCII table:\n```{{ results_df | tabulate(tablefmt='pretty', headers='keys') }}```"
 )
 
+MULTIPLE_QUERY_IN_ONE_RUN = (
```
phanikumv, line 27: The table name needs to change as "sample_table" is already being created in task above

Lee-W: That's why I put it as the last task. But sure, we can change it to another name

phanikumv (approved):

astronomer/providers/snowflake/example_dags/example_snowflake.py
```diff
@@ -66,8 +74,13 @@
         task_id="snowflake_op_sql_select_stmts", sql=SNOWFLAKE_SLACK_SQL, return_last=False
     )
 
+    snowflake_op_multiple_query_in_one_run = SnowflakeOperatorAsync(
```
pankajkoti, line 77: Is the objective of this task something different than the above `snowflake_op_sql_multiple_stmts` task?

phanikumv: yes we basically want create, insert and drop in the same batch

PR #994, opened 2023-04-26
feat(livy): check state before deferring LivyOperatorAsync

Lee-W:
Similar to [SnowflakeOperatorAsync](https://github.com/astronomer/astronomer-providers/blob/b2dade827ad8425952b2f5313b6a2863cb0c3e5e/astronomer/providers/snowflake/operators/snowflake.py#L217), we want to check whether the task has finished before it's deferred which can avoid unnecessarily deferring the task to the trigger. For most sensors, we have a poke method in its sync counterpart which can help us do such check

Closes: #995

### DIFF
--- a/astronomer/providers/apache/livy/operators/livy.py
+++ b/astronomer/providers/apache/livy/operators/livy.py
@@ -91,4 +91,5 @@ class LivyOperatorAsync(LivyOperator):
             self.task_id,
             event["response"],
         )
+        context["ti"].xcom_push(key="app_id", value=self.get_hook().get_batch(event["batch_id"])["appId"])
         return event["batch_id"]
--- a/astronomer/providers/apache/livy/operators/livy.py
+++ b/astronomer/providers/apache/livy/operators/livy.py
@@ -2,7 +2,7 @@
 from typing import Any, Dict
 
 from airflow.exceptions import AirflowException
-from airflow.providers.apache.livy.operators.livy import LivyOperator
+from airflow.providers.apache.livy.operators.livy import BatchState, LivyOperator
 
 from astronomer.providers.apache.livy.triggers.livy import LivyTrigger
 from astronomer.providers.utils.typing_compat import Context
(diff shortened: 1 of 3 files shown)


astronomer/providers/apache/livy/operators/livy.py
```diff
@@ -40,26 +40,38 @@ class LivyOperatorAsync(LivyOperator):
...
+                ),
+                method_name="execute_complete",
+            )
+        else:
+            self.log.info("Batch with id %s terminated with state: %s", self._batch_id, state.value)
+            hook.dump_batch_logs(self._batch_id)
+            if state != BatchState.SUCCESS:
+                raise AirflowException(f"Batch {self._batch_id} did not succeed")
+
+            context["ti"].xcom_push(key="app_id", value=self.get_hook().get_batch(self._batch_id)["appId"])
```
pankajastro, line 73: this app_id will not get pushed in xcom if the operator defer, right? shall we push this before defer or in execute_complete wdyt?

Lee-W: Sounds great! Just updated

astronomer/providers/apache/livy/operators/livy.py
```diff
@@ -79,4 +91,5 @@ def execute_complete(self, context: Context, event: Dict[str, Any]) -> Any:
             self.task_id,
             event["response"],
         )
+        context["ti"].xcom_push(key="app_id", value=self.get_hook().get_batch(event["batch_id"])["appId"])
```
pankajastro, line 94: I'm not sure, but I feel `self.get_hook()` may be none here wdyt? would it be possible to run DAG once?

pankajkoti (commented):
Some more description to the PR is nicer to take a quick look in the commit history and get a quick sneak on what & why we did in the commit.

pankajkoti (approved):

Issue #999, opened 2023-04-27
Explore project starship to clone connections from one Astro Cloud deployment to another


pankajkoti:
https://pypi.org/project/astronomer-starship-provider/

The above lists an example DAG that could help us transfer connections from one Astro Cloud deployment to another. 
Deploy such a DAG in our integration tests and see if we can transfer the connections and other meta to a new deployment in Astro Cloud.

PR #1002, opened 2023-04-27
Use Astronomer Starship Provider to transfer metadata between Astro Cloud deployments

pankajkoti:
The `https://pypi.org/project/astronomer-starship-provider/` helps transfer Airflow metadata from source deployment to Astro Cloud deployment. The PR adds a migration dag that facilitates transfer of Airflow metadata from source Airflow deployment (including Astro cloud deployments) to target Astro Cloud deployment. With this, using the Trigger DAG with config we can easily transfer metadata instead of creating those manually for the new deployments that we launch needing similar connections, variables and other metadata.

closes: #999

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -52,8 +52,6 @@ apache.hive =
 apache.livy =
     apache-airflow-providers-apache-livy
     paramiko
-astronomer.starship =
-    astronomer-starship-provider
 cncf.kubernetes =
     apache-airflow-providers-cncf-kubernetes>=4
     kubernetes_asyncio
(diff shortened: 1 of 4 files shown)


pankajkoti:
The DAG transferred **Airflow connections and variables** in just **8 seconds** from our existing providers-integration-tests deployment to a new deployment I created for testing this.

Screenshot 2023-04-27 at 12 23 35 PM

Screenshot 2023-04-27 at 12 23 51 PM

Screenshot 2023-04-27 at 12 26 37 PM

setup.cfg
```diff
@@ -128,6 +130,7 @@ all =
     apache-airflow-providers-snowflake
     apache-airflow-providers-sftp
     apache-airflow-providers-microsoft-azure
+    astronomer-starship-provider
```
phanikumv, line 133: I dont think this should go into all section, as this is related to migration

Lee-W (approved):
Looks great 👍

PR #1006, opened 2023-04-27
feat(sftp): check state before deferring SFTPSensorAsync

Lee-W:


Lee-W:
## providers-integration-test

image

Lee-W:
Currently, the example failed due to `PythonOperator` before `SFTPSensorAsync`. Will need to take a deeper look

image

pankajastro (approved):
LGTM

astronomer/providers/sftp/sensors/sftp.py
```diff
@@ -43,17 +43,18 @@ def execute(self, context: Context) -> None:
...
-            trigger=SFTPTrigger(
-                path=self.path,
-                file_pattern=self.file_pattern,
-                sftp_conn_id=self.sftp_conn_id,
-                poke_interval=self.poke_interval,
-                newer_than=self.newer_than,
-            ),
-            method_name="execute_complete",
-        )
+        if not self.poke(context=context):
```
pankajkoti, line 46: Can we add a description regarding this in the PR? I went to check the code what poke actually does. But, would be nice for a reviewer to quickly check the description to understand what exactly we're doing. Also, PRs with descriptions are better to go and check the history in the future. The commit message will easily give more details than having to go and check what's happening in the PR actually :)

pankajkoti (commented):
Looks great 👍🏽 . Have added a comment emphasising on having PR descriptions :)

astronomer/providers/sftp/sensors/sftp.py
Lee-W: Yep, sure!

Issue #1008, opened 2023-04-28
Accept DAG schedule time as Astro Cloud Deployment environment variable for master DAG in integration tests


pankajkoti:
Now that we have two master DAGs running one on Celery Executor and another on Kubernetes Executor, we observed quite a few failures today in our #provider-alert Slack channel and the reason I believe is that both the DAGs contend for the same resources and they fail.
We should have an Astro Cloud deployment environment variable so that it is set to different times for master DAGs to run without contention.

PR #1009, opened 2023-04-28
test(integration-tests): make master dag schedule configurable from env var

Lee-W:
Closes: #1008

### DIFF
--- a/.circleci/integration-tests/master_dag.py
+++ b/.circleci/integration-tests/master_dag.py
@@ -18,6 +18,7 @@ from airflow.utils.session import create_session
 SLACK_CHANNEL = os.getenv("SLACK_CHANNEL", "#provider-alert")
 SLACK_WEBHOOK_CONN = os.getenv("SLACK_WEBHOOK_CONN", "http_slack")
 SLACK_USERNAME = os.getenv("SLACK_USERNAME", "airflow_app")
+MASTER_DAG_SCHEDULE = os.getenv("MASTER_DAG_SCHEDULE", "@daily")
 
 
 def get_report(dag_run_ids: List[str], **context: Any) -> None:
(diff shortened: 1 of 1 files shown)


.circleci/integration-tests/master_dag.py
```diff
@@ -89,7 +90,7 @@ def prepare_dag_dependency(task_info, execution_time):
 
 with DAG(
     dag_id="example_master_dag",
-    schedule_interval="@daily",
+    schedule_interval=MASTER_DAG_SCHEDULE,
```
Lee-W, line 93: Should we replace `schedule_interval` with `schedule`?

pankajastro: +1 for this. maybe in this PR or in separate PR both ok to me

pankajastro: not only here but also other places if we have this param

pankajastro: a separate PR is ok to me

Lee-W: https://github.com/astronomer/astronomer-providers/pull/1010

pankajastro: shall we make the default value a cron expression? That way I feel both the env value and default value would be consistent, otherwise  `@daily` in 2 env we can't do?

Lee-W: I don't have a strong opinion on this one. Either way works for me. Just updated it

pankajastro (approved):
LGTM

pankajkoti (commented):
Were we able to test this in Astro Cloud? If not could we do a workflow_dispatch and test it once?

Lee-W:
Sure! Let me test it now!

Lee-W:
## providers-integration-tests
image


## providers-integration-tests-on-KE
image

pankajkoti:
looks great, thank you 👏🏽

PR #1012, opened 2023-04-28
feat(dbt): check state before deferring DbtCloudJobRunSensorAsync

Lee-W:
Similar to SnowflakeOperatorAsync, we want to check whether the task has finished before it's deferred. For most sensors, we have a poke method in its sync counter part which can help us do such check.


Lee-W:
## providers-integration-tests
image

## providers-integration-tests-on-ke

image

pankajkoti (approved):

PR #1014, opened 2023-04-28
Add a make target to bump version

pankajastro:
Add a make target to bump the version locally.
 example
```
make ASTRO_PROVIDER_VERSION=dev bump-version  // this will set the version to next minor-version-dev1 example 1.1.0-dev1
make ASTRO_PROVIDER_VERSION=1.1.1 bump-version
make ASTRO_PROVIDER_VERSION=1.1.0 bump-version 
make ASTRO_PROVIDER_VERSION=1.0.0 bump-version
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,8 @@
 .PHONY: dev logs stop clean build build-emr_eks_container_example_dag-image build-aws build-google-cloud build-run docs
 .PHONY: restart restart-all run-tests run-static-checks run-mypy run-local-lineage-server test-rc-deps shell help
 
+ASTRO_PROVIDER_VERSION ?= "dev"
+
 # If the first argument is "run"...
 ifeq (run-mypy,$(firstword $(MAKECMDGOALS)))
   # use the rest as arguments for "run"
@@ -97,6 +99,13 @@ test-rc-deps: ## Test providers RC by building an image with given dependencies 
 
 shell:  ## Runs a shell within a container (Allows interactive session)
 	docker compose -f dev/docker-compose.yaml run --rm airflow-scheduler bash
+
+bump-version:  ## Bump versions in files locally. By default bump to DEV. set env ASTRO_PROVIDER_VERSION to change default value (dev | patch | major | minor).
+	@if [ $(ASTRO_PROVIDER_VERSION) = "dev" ]; then\
+		cz bump --version-type semver --increment minor --devrelease 1 --files-only;\
+	else \
+	  cz bump --version-type semver --files-only --increment $(ASTRO_PROVIDER_VERSION);\
+    fi
 
 help: ## Prints this message
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-41s\033[0m %s\n", $$1, $$2}'
--- a/Makefile
+++ b/Makefile
@@ -103,8 +103,8 @@ shell:  ## Runs a shell within a container (Allows interactive session)
 bump-version:  ## Bump versions in files locally. By default bump to DEV. set env ASTRO_PROVIDER_VERSION to change default version. see docs https://github.com/astronomer/astronomer-providers/blob/main/RELEASING.rst#update-the-version-number
 	@if [ $(ASTRO_PROVIDER_VERSION) = "dev" ]; then\
 		cz bump --version-type semver --increment minor --devrelease 1 --files-only;\
-	else \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -98,5 +100,12 @@ test-rc-deps: ## Test providers RC by building an image with given dependencies
 shell:  ## Runs a shell within a container (Allows interactive session)
 	docker compose -f dev/docker-compose.yaml run --rm airflow-scheduler bash
 
+bump-version:  ## Bump versions in files locally. By default bump to DEV. set env ASTRO_PROVIDER_VERSION to change default value (dev | patch | major | minor).
+	@if [ $(ASTRO_PROVIDER_VERSION) = "dev" ]; then\
+		cz bump --version-type semver --increment minor --devrelease 1 --files-only;\
+	else \
+	  cz bump --version-type semver --files-only --increment $(ASTRO_PROVIDER_VERSION);\
```
pankajkoti, line 107: What happens when the previous release was 1.15.6, next we bump it to 1.16.0-dev1 and then want to create a patch release. Will it be able to be set to 1.15.7 from 1.16.0-dev1?

pankajastro: good point, I think `commitizen` don't know the last released version so should be 1.16.1.

pankajastro: but let me check if there is way to fix it

pankajastro: I don't think they can move backwards, but one can pass the manual version for example `cz bump 1.15.6` maybe I'll few more condition for this, thanks for pointing out

Lee-W: In essence, commitizen handles project versioning in accordance with [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) and [SemVer](https://semver.org/). Thus, downgrading project version might not be something that has ever considered by commitizen

RELEASING.rst
```diff
@@ -25,6 +25,17 @@ You will need to update it in:
...
+Install commitizen on mac
+
+```
+brew install commitizen
+```
+
+Bump versions locally
+
+```
+make ASTRO_PROVIDER_VERSION=(1.0.0 | 1.1.0 | 1.1.1) bump-version
```
Lee-W, line 37: Sorry, I'm a bit confused by `(1.0.0 | 1.1.0 | 1.1.1)`. What does that mean?

pankajastro: yep, it's confusing. I'll clean up it. But I mean you can pass major or minor or patch version

pankajastro: yes, we can do that. in fact, I did that first https://github.com/astronomer/astronomer-providers/pull/1014/commits/f09b613968fdf057f6b27f810d8ce63753c39f1e but let's say my dev version is 1.7.0-dev1 and if I'll run `cz bump --increment PATCH` it set version to 1.7.1 which is problematic if my last release was 1.6.5 and I want to release 1.6.6, is this correct I'm missing something?

pankajastro: I have tried to update a bit, PTAL

Makefile
```diff
@@ -98,5 +100,12 @@ test-rc-deps: ## Test providers RC by building an image with given dependencies
 shell:  ## Runs a shell within a container (Allows interactive session)
 	docker compose -f dev/docker-compose.yaml run --rm airflow-scheduler bash
 
+bump-version:  ## Bump versions in files locally. By default bump to DEV. set env ASTRO_PROVIDER_VERSION to change default version (1.0.0 | 1.1.0 | 1.1.1).
```
Lee-W, line 103: maybe we'll need to change here as well?

pankajastro: yep, thanks for pointing it out.

Makefile
```diff
@@ -98,5 +100,12 @@ test-rc-deps: ## Test providers RC by building an image with given dependencies
 shell:  ## Runs a shell within a container (Allows interactive session)
 	docker compose -f dev/docker-compose.yaml run --rm airflow-scheduler bash
 
+bump-version:  ## Bump versions in files locally. By default bump to DEV. set env ASTRO_PROVIDER_VERSION to change default version. see docs https://github.com/astronomer/astronomer-providers/blob/main/RELEASING.rst#update-the-version-number
+	@if [ $(ASTRO_PROVIDER_VERSION) = "dev" ]; then\
```
Lee-W, line 104: The indentation here looks a bit unusual here. Is it something expected?

pankajastro: looks ok to me even if you see L101. any specific thing you like to point out?
Screenshot 2023-05-02 at 4 50 07 PM

Lee-W: 圖片

Ah, what I meant is 104-109

pankajastro: Nice catch, fixed it.

PR #1019, opened 2023-05-02
Include AWS nuke DAG in integration test master DAG

pankajastro:
since now we run tests for both KE/CE 
and both use the same resource 
so it makes sense to include nuke DAG in master DAG
 so that cleanup starts as soon as test execution is completed.

### DIFF
--- a/astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
@@ -1,3 +1,4 @@
+"""DAG to nuke AWS resources."""
 import os
 from datetime import datetime, timedelta
 
@@ -20,7 +21,7 @@ default_args = {
 with DAG(
     dag_id="example_aws_nuke",
     start_date=datetime(2022, 1, 1),
-    schedule="30 20 * * *",
+    schedule=None,
     catchup=False,
     default_args=default_args,
     tags=["example", "aws-nuke"],


.circleci/integration-tests/master_dag.py
```diff
@@ -242,6 +242,18 @@ def prepare_dag_dependency(task_info, execution_time):
         provide_context=True,
     )
 
+    end_test = DummyOperator(
+        task_id="end_test",
+        trigger_rule="all_success",
+    )
+
+    aws_nuke_task_info = [
+        {"aws_nuke_dag": "example_aws_nuke"},
```
phanikumv, line 251: The example_aws_nuke has its schedule set to 8:30 PM, that might need to be corrected. WDYT?

pankajastro: shall I set to None?

.circleci/integration-tests/master_dag.py
```diff
@@ -242,6 +242,18 @@ def prepare_dag_dependency(task_info, execution_time):
         provide_context=True,
     )
 
+    end_test = DummyOperator(
+        task_id="end_test",
+        trigger_rule="all_success",
```
pankajkoti, line 247: The nuke DAG may not run as it is a dependency of this task and this task will only be triggered when all previous DAGs are successful

pankajkoti (commented):
We need to re-check the trigger rule as the nuke DAG should run irrespective of the previous DAGs status and after all the DAGs have finished processing.

.circleci/integration-tests/master_dag.py
pankajastro: we have added a trigger rule in`TriggerDagRunOperator` isn't that sufficient?

pankajkoti: No, it will not be sufficient. end_test will not be triggered if any of the DAGs fail. And in that case, since end_test will not be triggered, it will never be done and hence nuke DAG won't be triggered.

pankajastro: if the previous DAG will fail the status of `end_test` will be upstream fail, isn't that the case?
and if a task has trigger_rule "all_done" it always runs (which is the case of the nuke task)? or I'm missing something

pankajkoti: can you please paste a screenshot of the Graph when one of the providers DAGs fail? If `upstream_failed` tasks do not hinder `all_done` then it I guess you're right and we're safe. But will the final `end` task report an `upstream_failed` when one of the providers' DAGs fails? Or will it be marked as a success if only the nuke DAG runs successfully? My questions is because we should mark the master DAG status as failed if any of the DAGs fail.

pankajastro: ok, I'll put a screenshot here but it should be `[provider-DAG]->report->end_test->nuke->end`

pankajkoti: yes and try to fail the provider DAG on purpose to observe the behaviour

pankajastro: Screenshot 2023-05-02 at 4 26 17 PM

pankajkoti: Why did the aws_nuke_dag fail?

what happens if provider DAGs fail but `aws_nuke_dag` succeeds? Does the `end` task report upstream_failed? Because if `end` is marked as success in that case then the `master_dag` will also be marked as successful, no?

pankajastro: if we think that way we have more problems. what happens if all tasks succeed and bigquery fail? the task will be marked a success. the trigger rule applies to just the previous task and DAG status is the status of the last task.  but I'm ok to remove the end task but that will not solve the problem you are pointing out. One thing  we could do is make all task previous to the end task but that looks out of the scope of this PR

pankajkoti: The problem I see is the change in this PR might alter the behaviour we have currently in the `main` branch.

Currently, if any provider DAGs fail then the master DAG is also marked as `failed`. But with this change, I think if only the nuke dag runs fine which could & should happen even if any provider DAG fails, then the `end` will be marked `success` and also the `master_dag` will be marked `success` **even though provider DAGs might have failed**.

Attaching screenshots of the current DAG graph and how we mark the `master_dag` as `failed` even if one provider DAG fails.
Screenshot 2023-05-02 at 4 52 18 PM

Screenshot 2023-05-02 at 4 52 50 PM

pankajastro: I see the end task is success in the graph view but upstream fail in the grid view, is that expected?

pankajkoti: yes, that is another issue but that fix can come in a separate PR in my opinion. But, at least we should try to keep the existing behaviour as it is in this PR. WDYT?

pankajastro: I have updated the DAG a little bit but still, it would not solve the problem that you are pointing out for all cases.

Screenshot 2023-05-02 at 5 38 27 PM

pankajkoti: this looks like a good idea to me for now. The other issue is an existing issue and we can address that in another PR later

pankajkoti (approved):
Looks great to me apart from the thing that we might need to set AWS nuke DAG schedule to None as suggested by Phani.

pankajkoti:
> Look like circle CI is not stable today :(

It's not CircleCI, but one of our tests tries to reach out to an endpoint, https://www.httpbin.org/test-endpoint.

This endpoint is returning a bad gateway intermittently and hence the tests are failing. This is happening since last week.

pankajastro:
> apart from the thing that we might need to set AWS nuke DAG schedule to None as suggested by Phani

thought about this and I'm ok to make that change. but if we run it also once at the scheduled time so that if any manual resource is created it will nuke that would be not a bad idea.

pankajastro:
> > Look like circle CI is not stable today :(
> 
> It's not CircleCI, but one of our tests tries to reach out to an endpoint, https://www.httpbin.org/test-endpoint.
> 
> This endpoint is returning a bad gateway intermittently and hence the tests are failing. This is happening since last week.
> 
> Screenshot 2023-05-02 at 5 45 55 PM

maybe we should mock that

pankajastro:
> maybe we should mock that
PR https://github.com/astronomer/astronomer-providers/pull/1020

.circleci/integration-tests/master_dag.py
phanikumv: yeah please set it to None

phanikumv:
> > apart from the thing that we might need to set AWS nuke DAG schedule to None as suggested by Phani
> 
> thought about this and I'm ok to make that change. but if we run it also once at the scheduled time so that if any manual resource is created it will nuke that would be not a bad idea.

hmm, prefer to run nuke after the CE and KE runs in the master DAG to preserve uniformity. Lets not run it again at 8:30 PM. I believe it will be an overkill

.circleci/integration-tests/master_dag.py
pankajastro: ok, set this the schedule None for AWS nuke DAG

PR #1020, opened 2023-05-02
Mock external http call

pankajastro:
Fix transient failure for tests. Mocking external call. this is happening look like after sensor optimization.

### DIFF
--- a/tests/http/sensors/test_http.py
+++ b/tests/http/sensors/test_http.py
@@ -53,8 +53,9 @@ class TestHttpSensorAsync:
 
         assert isinstance(exc.value.trigger, HttpTrigger), "Trigger is not a HttpTrigger"
 
+    @mock.patch("astronomer.providers.http.sensors.http.HttpSensorAsync.defer")
     @mock.patch("airflow.sensors.base.BaseSensorOperator.execute")
-    def test_sensor_not_defer(self, mock_execute):
+    def test_sensor_not_defer(self, mock_execute, mock_defer):
         task = HttpSensorAsync(
             task_id="run_now",
             endpoint="test-endpoint",
@@ -62,6 +63,7 @@ class TestHttpSensorAsync:
         )
         task.execute({})
         mock_execute.assert_called_once()
+        mock_defer.assert_not_called()
 
     @mock.patch("airflow.providers.http.sensors.http.HttpSensor.poke")
     def test_sensor_defer(self, mock_poke):
--- a/tests/http/sensors/test_http.py
+++ b/tests/http/sensors/test_http.py
@@ -1,4 +1,3 @@
-import datetime
 from unittest import mock
 
 import pytest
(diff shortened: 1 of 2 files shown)


pankajastro:
Ran test 3 times to make sure the result is consitent

tests/http/sensors/test_http.py
```diff
@@ -54,29 +53,23 @@ def test_http_run_now_sensor_async(self, mock_poke):
 
         assert isinstance(exc.value.trigger, HttpTrigger), "Trigger is not a HttpTrigger"
 
-    def test_http_response_check_does_not_run_async(self):
-        """Asserts that a task is not deferred when response_check arg is passed to HttpSensorAsync."""
-
+    @mock.patch("airflow.sensors.base.BaseSensorOperator.execute")
+    def test_sensor_not_defer(self, mock_execute):
```
pankajkoti, line 57: Do we not need to check here that defer is not called?

pankajastro: yes, since we do not always defer so it makes sense to check that if defer is not expected it should not defer.

pankajastro: maybe I can add that too

pankajastro: Added, let see what CI think

tests/http/sensors/test_http.py
```diff
@@ -54,29 +53,25 @@ def test_http_run_now_sensor_async(self, mock_poke):
...
             task_id="run_now",
             endpoint="test-endpoint",
             response_check=lambda response: "httpbin" in response.text,
         )
-
-        with mock.patch(
-            "astronomer.providers.http.sensors.http.HttpSensorAsync.defer"
-        ) as mock_defer, mock.patch("airflow.sensors.base.BaseSensorOperator.execute"):
-            task.execute({})
-
```
pankajkoti, line 70: Can we not use mocking it in the same fashion?

pankajkoti (commented):
Also, we could test the test cases locally too, no? Or we could only these tests on CI?

tests/http/sensors/test_http.py
pankajastro: I prefer to have have small test and one test testing one path here

tests/http/sensors/test_http.py
```diff
@@ -54,29 +53,25 @@ def test_http_run_now_sensor_async(self, mock_poke):
...
+    def test_sensor_defer(self, mock_poke):
+        mock_poke.return_value = False
         task = HttpSensorAsync(task_id="run_now", endpoint="test-endpoint")
-
-        with mock.patch("astronomer.providers.http.sensors.http.HttpSensorAsync.defer") as mock_defer:
+        with pytest.raises(TaskDeferred) as exc:
             task.execute({})
-            mock_defer.assert_called_once_with(
-                timeout=datetime.timedelta(days=7), trigger=mock.ANY, method_name="execute_complete"
-            )
```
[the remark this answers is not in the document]
pankajastro: we are testing by checking TaskDeferred exception

pankajastro:
> Also, we could test the test cases locally too, no? Or we could only these tests on CI?

you can run locally

pankajkoti:
> > Also, we could test the test cases locally too, no? Or we could only these tests on CI?
> 
> you can run locally. I think we have make target and also pytest should work

okay, I felt so too. My question was in response to https://github.com/astronomer/astronomer-providers/pull/1020#discussion_r1182593144 :)

pankajkoti (approved):
Thank you for fixing the test 👏🏽

PR #1028, opened 2023-05-03
Check whether the task finishes before deferring the task for WasbBlobSensorAsync

Lee-W:

### DIFF
--- a/tests/microsoft/azure/sensors/test_wasb.py
+++ b/tests/microsoft/azure/sensors/test_wasb.py
@@ -17,6 +17,8 @@ TEST_DATA_STORAGE_BLOB_NAME = "test_blob_providers_team.txt"
 TEST_DATA_STORAGE_CONTAINER_NAME = "test-container-providers-team"
 TEST_DATA_STORAGE_BLOB_PREFIX = TEST_DATA_STORAGE_BLOB_NAME[:10]
 
+MODULE = "astronomer.providers.microsoft.azure.sensors.wasb"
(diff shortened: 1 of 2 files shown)


Lee-W:
圖片

tests/microsoft/azure/sensors/test_wasb.py
```diff
@@ -25,7 +27,15 @@ class TestWasbBlobSensorAsync:
         blob_name=TEST_DATA_STORAGE_BLOB_NAME,
     )
 
-    def test_wasb_blob_sensor_async(self):
+    @mock.patch(f"{MODULE}.WasbBlobSensorAsync.defer")
+    @mock.patch(f"{MODULE}.WasbBlobSensorAsync.poke", return_value=True)
+    def test_WasbBlobSensorAsync_async_finish_before_deferred(self, mock_poke, mock_defer, context):
```
pankajkoti, line 32: Can we use a snake case here for the test name? also async is repeated twice in the name.

pankajkoti (approved):
Looks good, minor suggestion

tests/microsoft/azure/sensors/test_wasb.py
Lee-W: ah yes, I forget to update it. let me fix it for all my branches. Thanks!

Lee-W: Just fixed it

PR #1037, opened 2023-05-04
Check whether the task finishes before deferring the task for S3KeySensorAsync

Lee-W:

### DIFF
--- a/astronomer/providers/amazon/aws/sensors/s3.py
+++ b/astronomer/providers/amazon/aws/sensors/s3.py
@@ -69,9 +69,10 @@ class S3KeySensorAsync(S3KeySensor):
         verify: str | bool | None = None,
         **kwargs: Any,
     ):
+        self.bucket_key: list[str] = [bucket_key] if isinstance(bucket_key, str) else bucket_key
         super().__init__(
             bucket_name=bucket_name,
-            bucket_key=[bucket_key] if isinstance(bucket_key, str) else bucket_key,
+            bucket_key=self.bucket_key,
             wildcard_match=wildcard_match,
             check_fn=check_fn,
             aws_conn_id=aws_conn_id,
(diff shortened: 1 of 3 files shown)


Lee-W:
```
  File "/usr/local/lib/python3.10/site-packages/astronomer/providers/amazon/aws/sensors/s3.py", line 83, in execute
    if not self.poke(context):
  File "/usr/local/lib/python3.10/site-packages/airflow/providers/amazon/aws/sensors/s3.py", line 132, in poke
    return all(self._check_key(key) for key in self.bucket_key)
  File "/usr/local/lib/python3.10/site-packages/airflow/providers/amazon/aws/sensors/s3.py", line 132, in <genexpr>
    return all(self._check_key(key) for key in self.bucket_key)
  File "/usr/local/lib/python3.10/site-packages/airflow/providers/amazon/aws/sensors/s3.py", line 118, in _check_key
    obj = self.hook.head_object(key, bucket_name)
AttributeError: 'NoneType' object has no attribute 'head_object'
```

Lee-W:
image

astronomer/providers/amazon/aws/sensors/s3.py
```diff
@@ -59,40 +61,43 @@ def check_fn(files: List) -> bool:
...
-                wildcard_match=self.wildcard_match,
-                check_fn=self.check_fn,
-                aws_conn_id=self.aws_conn_id,
-                verify=self.verify,
-                poke_interval=self.poke_interval,
-            ),
-            method_name="execute_complete",
-        )
+        if not self.poke(context):
+            self.hook: S3Hook | None = None  # type: ignore[assignment]
```
pankajkoti, line 85: Do we need this line anymore? If not, can we remove it?

astronomer/providers/amazon/aws/sensors/s3.py
```diff
@@ -59,40 +61,43 @@ def check_fn(files: List) -> bool:
...
+            self.defer(
+                timeout=timedelta(seconds=self.timeout),
+                trigger=S3KeyTrigger(
+                    bucket_name=cast(str, self.bucket_name),
+                    bucket_key=self.bucket_key,  # type: ignore[arg-type]
```
pankajkoti, line 90: Can we get away with this type ignore by correcting the type in the S3KeyTrigger class?

pankajkoti (approved):

astronomer/providers/amazon/aws/sensors/s3.py
pankajastro: We have a `hook` property in the base class `S3KeySensor` but that gets antialiased using connection id so I feel should be ok to remove

pankajastro (approved):

Lee-W:
image

I've addressed both comments and tested on astro cloud. As this is approved, I'll merge it now. Thanks

astronomer/providers/amazon/aws/sensors/s3.py
Lee-W: I removed it by using different logic in `__init__` instead of changing `S3KeyTrigger` (I don't think we should change the trigger)

astronomer/providers/amazon/aws/sensors/s3.py
Lee-W: Yep, just removed

PR #1038, opened 2023-05-04
Check whether the task finishes before deferring the task for S3KeysUnchangedSensorAsync

Lee-W:

### DIFF
--- a/astronomer/providers/amazon/aws/sensors/s3.py
+++ b/astronomer/providers/amazon/aws/sensors/s3.py
@@ -185,7 +185,7 @@ class S3KeysUnchangedSensorAsync(S3KeysUnchangedSensor):
                     allow_delete=self.allow_delete,
                     aws_conn_id=self.aws_conn_id,
                     verify=self.verify,
-                    last_activity_time=None,
+                    last_activity_time=self.last_activity_time,
                 ),
                 method_name="execute_complete",
             )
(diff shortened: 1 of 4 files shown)


Lee-W:
圖片

astronomer/providers/amazon/aws/sensors/s3.py
```diff
@@ -172,22 +172,23 @@ def __init__(
...
+                    bucket_name=self.bucket_name,
+                    prefix=self.prefix,
+                    inactivity_period=self.inactivity_period,
+                    min_objects=self.min_objects,
+                    previous_objects=self.previous_objects,
+                    inactivity_seconds=self.inactivity_seconds,
+                    allow_delete=self.allow_delete,
+                    aws_conn_id=self.aws_conn_id,
+                    verify=self.verify,
+                    last_activity_time=None,
```
pankajkoti, line 188: Why are we setting this to None as opposed to self.last_activity_time previously?

Lee-W: `poke` might change the value of `self.last_activity_time`. Thus, I change it to the original value of `self.last_activity_time`. Another option is to store the original value and assign it here. But after tracing the code, I discovered that users cannot change this value. Thus I think using its default value makes sense.

pankajkoti: even if poke changes it, it's still valid, right? We did a check in poke and that was the last activity. WDYT?

Lee-W: I've addressed this. Thanks!

pankajastro (approved):

pankajkoti (approved):

PR #1041, opened 2023-05-05
Check whether the job finishes before deferring the task for BigQueryInsertJobOperatorAsync

Lee-W:

### DIFF
--- a/astronomer/providers/google/cloud/operators/bigquery.py
+++ b/astronomer/providers/google/cloud/operators/bigquery.py
@@ -79,7 +79,7 @@ class BigQueryInsertJobOperatorAsync(BigQueryInsertJobOperator, BaseOperator):
         super().__init__(*args, **kwargs)
         self.poll_interval = poll_interval
 
-    def execute(self, context: Context) -> None:  # noqa: D102
+    def execute(self, context: Context) -> Any:  # noqa: D102
         kwargs: dict[Any, Any] = {}
         if hasattr(self, "delegate_to"):
             kwargs["delegate_to"] = self.delegate_to
@@ -124,18 +124,23 @@ class BigQueryInsertJobOperatorAsync(BigQueryInsertJobOperator, BaseOperator):
 
         self.job_id = job.job_id
         context["ti"].xcom_push(key="job_id", value=self.job_id)
-        self.defer(
-            timeout=self.execution_timeout,
-            trigger=BigQueryInsertJobTrigger(
-                conn_id=self.gcp_conn_id,
-                job_id=self.job_id,
-                project_id=self.project_id,
-                impersonation_chain=self.impersonation_chain,
-                poll_interval=self.poll_interval,
-                **kwargs,
-            ),
-            method_name="execute_complete",
-        )
+        if job.running():
+            self.defer(
+                timeout=self.execution_timeout,
+                trigger=BigQueryInsertJobTrigger(
+                    conn_id=self.gcp_conn_id,
+                    job_id=self.job_id,
+                    project_id=self.project_id,
+                    impersonation_chain=self.impersonation_chain,
+                    poll_interval=self.poll_interval,
+                    **kwargs,
+                ),
+                method_name="execute_complete",
+            )
+        else:
+            job.result()
+            self._handle_job_error(job)
+            return self.job_id
 
     def execute_complete(self, context: Context, event: dict[str, Any]) -> None:
         """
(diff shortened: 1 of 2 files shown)


Lee-W:
圖片

astronomer/providers/google/cloud/operators/bigquery.py
```diff
@@ -124,18 +124,25 @@ def execute(self, context: Context) -> None:  # noqa: D102
...
-                job_id=self.job_id,
-                project_id=self.project_id,
-                impersonation_chain=self.impersonation_chain,
-                poll_interval=self.poll_interval,
-                **kwargs,
-            ),
-            method_name="execute_complete",
-        )
+
+        # breakpoint()
```
pankajastro, line 128: can we remove this comment?

Lee-W: Just removed. Thanks!

astronomer/providers/google/cloud/operators/bigquery.py
```diff
@@ -124,18 +124,23 @@ def execute(self, context: Context) -> None:  # noqa: D102
...
+                    job_id=self.job_id,
+                    project_id=self.project_id,
+                    impersonation_chain=self.impersonation_chain,
+                    poll_interval=self.poll_interval,
+                    **kwargs,
+                ),
+                method_name="execute_complete",
+            )
+        else:
+            job.result()
```
pankajastro, line 141: I was looking at `BigQueryInsertJobOperator` implementation for the sync version they do `job.result()`  so here if somehow else block gets executed it would be the same as the sync operator??

astronomer/providers/google/cloud/operators/bigquery.py
```diff
@@ -124,18 +124,23 @@ def execute(self, context: Context) -> None:  # noqa: D102
...
-                conn_id=self.gcp_conn_id,
-                job_id=self.job_id,
-                project_id=self.project_id,
-                impersonation_chain=self.impersonation_chain,
-                poll_interval=self.poll_interval,
-                **kwargs,
-            ),
-            method_name="execute_complete",
-        )
+        if job.running():
```
pankajkoti, line 127: Can you please point me to the implementation of the running method?

astronomer/providers/google/cloud/operators/bigquery.py
Lee-W: Yes, we're doing the same thing without waiting and retrying, as the job is not in running status. If it is, we'll defer it

pankajkoti: yes, but that should be okay. We know here that the job is not running anymore, has reached a terminal state and would like to proceed ahead without having an overhead of the deferral.

astronomer/providers/google/cloud/operators/bigquery.py
```diff
@@ -124,18 +124,23 @@ def execute(self, context: Context) -> None:  # noqa: D102
...
+                    impersonation_chain=self.impersonation_chain,
+                    poll_interval=self.poll_interval,
+                    **kwargs,
+                ),
+                method_name="execute_complete",
+            )
+        else:
+            job.result()
+            self._handle_job_error(job)
+            return self.job_id
```
pankajkoti, line 143: if we are processing the job result here and returning job_id, we should also do the same in execute_complete to having consistent results in the XCOM for the task.

astronomer/providers/google/cloud/operators/bigquery.py
Lee-W: https://github.com/googleapis/python-api-core/blob/6acb3a95462f644037d4cf5f3c4d6a8a6149caa8/google/api_core/future/polling.py#L122

Lee-W: which calls https://github.com/googleapis/python-api-core/blob/6acb3a95462f644037d4cf5f3c4d6a8a6149caa8/google/api_core/future/polling.py#L122

astronomer/providers/google/cloud/operators/bigquery.py
Lee-W: Yep, that makes sense. Just update it. Thanks!

astronomer/providers/google/cloud/operators/bigquery.py
pankajkoti: yes, I was able to track it until there but where is https://github.com/googleapis/python-api-core/blob/6acb3a95462f644037d4cf5f3c4d6a8a6149caa8/google/api_core/future/polling.py#LL124C20-L124C31 (self.done() ) implemented

Lee-W: https://github.com/googleapis/python-bigquery/blob/main/google/cloud/bigquery/job/base.py#L855

Sorry, I paste the wrong link in the previous comment

PR #1042, opened 2023-05-05
Check whether the task finishes before deferring the task for BigQueryCheckOperatorAsync

Lee-W:

### DIFF
--- a/astronomer/providers/google/cloud/operators/bigquery.py
+++ b/astronomer/providers/google/cloud/operators/bigquery.py
@@ -186,17 +186,21 @@ class BigQueryCheckOperatorAsync(BigQueryCheckOperator):
         )
         job = self._submit_job(hook, job_id="")
         context["ti"].xcom_push(key="job_id", value=job.job_id)
-        self.defer(
-            timeout=self.execution_timeout,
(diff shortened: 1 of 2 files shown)


Lee-W:
圖片

astronomer/providers/google/cloud/operators/bigquery.py
```diff
@@ -186,17 +186,21 @@ def execute(self, context: Context) -> None:  # noqa: D102
...
+                    conn_id=self.gcp_conn_id,
+                    job_id=job.job_id,
+                    project_id=hook.project_id,
+                    impersonation_chain=self.impersonation_chain,
+                    poll_interval=self.poll_interval,
+                ),
+                method_name="execute_complete",
+            )
+        else:
+            super().execute(context=context)
```
pankajastro, line 203: same question: https://github.com/astronomer/astronomer-providers/pull/1041#discussion_r1187218915

Lee-W: Yes, we're doing the same thing without waiting and retrying, as the job is not in running status. If it is, we'll defer it

Issue #1051, opened 2023-05-07
S3 sensor example DAG marked as successful although few tasks failed


pankajkoti:
<img width="1721" alt="Screenshot 2023-05-07 at 3 45 32 PM" src="https://user-images.githubusercontent.com/10206082/236671548-e3be29a5-c324-4eb9-9058-40288b7145d8.png">

The screenshot above shows that the final DAG status was marked as successful although a few tasks failed. Even if one task fails, we should mark the DAG status as false. As a result, the Master DAG reports the DAG as successful in the Grid view, but the tasks that failed are only seen when we go to the Triggered DAG's grid view.
We should fail the DAG even if one task fails in the DAG

PR #1052, opened 2023-05-07
Mark S3 Sensor DAG failed if any of the tasks fail

pankajkoti:
fail.

closes: #1051

### DIFF
--- a/astronomer/providers/amazon/aws/example_dags/example_s3.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_s3.py
@@ -3,11 +3,14 @@ from datetime import datetime, timedelta
 from typing import Any, List
 
 from airflow.models.dag import DAG
+from airflow.operators.python import PythonOperator
 from airflow.providers.amazon.aws.operators.s3 import (
     S3CreateBucketOperator,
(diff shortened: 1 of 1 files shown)


pankajkoti:
Screenshot 2023-05-07 at 3 51 46 PM

astronomer/providers/amazon/aws/example_dags/example_s3.py
```diff
@@ -39,6 +42,17 @@
     "retry_delay": timedelta(seconds=int(os.getenv("DEFAULT_RETRY_DELAY_SECONDS", 60))),
 }
 
+
+def check_dag_status(**kwargs: Any) -> None:
```
Lee-W, line 46: As this is repeated in multiple DAGs, should we extract it to avoid duplication?

pankajkoti: I had this thought but have following next thoughts:
1. These are example DAGs and need to be self sufficient
2. We may not want to keep this as a util and release it as part of our providers

Owing to above reasons, I guess the duplication should be okay.

Issue #1053, opened 2023-05-07
Sagemaker example DAG marked as successful although few tasks failed


pankajkoti:
<img width="426" alt="Screenshot 2023-05-07 at 3 59 07 PM" src="https://user-images.githubusercontent.com/10206082/236672224-2c68ea16-49b5-4bc8-bbf1-3c3d77d240b3.png">

The screenshot above shows that the final DAG status was marked as successful although a few tasks failed. Even if one task fails, we should mark the DAG status as false. As a result, the Master DAG reports the DAG as successful in the Grid view, but the tasks that failed are only seen when we go to the Triggered DAG's grid view.
We should fail the DAG even if one task fails in the DAG

PR #1054, opened 2023-05-07
Mark SageMaker DAG failed if any of the tasks fail

pankajkoti:
fail.

closes: #1053


pankajkoti:
Screenshot 2023-05-07 at 4 06 28 PM

Lee-W (approved):

Issue #1021, opened 2023-05-02
SFTP provider to support additional requirements


pankajkoti:
WRT to ticket https://astronomer.zendesk.com/agent/tickets/18342 and feedback received from users, the SFTP provider should support the following requirements: 
- If the ssh host key and the SSH command line tool outputs are in the format of `ecdsa-sha2-nistp256`, then ideally the SFTP hook should recognise this format without the need for it to be edited just to get it working with Astronomer’s SFTP library
- If `no_host_key_check`: `False` on the connection, but doesn’t include a host key, it lets the connection through just fine and effectively doesn’t validate the server. This seems wrong. In fact, the whole need to enable host key checking by default seems wrong. It should be enabled by default (so by default users are then forced to use secure/validated connections), and if it is needed to test locally without ssh host key checking, let the user turn on `no_host_key_check`: `True` in the connection details to override the default. It seems a safer approach.
- Users are still having to write out to the docker containers `~/.ssh/config` file as part of their work because one of the servers they connect to is using a much older key exchange algorithm that the ssh libraries don’t check for by default, so they have to tell it which one to use for a specific site. Given they still need to write out this file, it would be convenient for them just to update the `~/.ssh/known_hosts` file too, now they know how to do it as part of the Dockerfile. It would make sense to them for the SFTP provider to default back to using the `known_hosts` file if a host key is not specifically provided on the connection. Developers have the choice then of which approach to take.

Edit: 
- Another new requirement came in 9th May which is available on the same Zendesk ticket on wherein the user would like to pass string values(fetched from `XCOM` results of the previous task) to the `newer_than` field that need to be converted to datetime values in the implementation as currently the `newer_than` field only accepts `datetime` fields.
- One more requirement came in on Friday, May 12th that the current implementation makes multiple roundtrips to the SFTP server for each of the files in the directory to get the file modified time when pattern and newer_than is given; however, this gives intermittent failures on the DAG as the SFTP server throttles the requests when there are large number of files. Need to avoid throttling by checking whether we can make fewer roundtrips.

pankajkoti:
I have opened a PR and have added an internal note on the Zendesk ticket to seek user feedback. Awaiting relay of the message and response from users.

PR #1072, opened 2023-05-09
Enhance SFTP Async sensor addressing user feedback

pankajkoti:
The PR address the following feature requests mentioned in the issue as per user feedback:
- If the ssh host key and the SSH command line tool outputs are in the format of `ecdsa-sha2-nistp256`, then ideally the SFTP hook should recognise this format without the need for it to be edited (by prepending `ssh-`) just to get it working with Astronomer’s SFTP library
- If no_host_key_check: False on the connection, but doesn’t include a host key, it lets the connection through just fine and effectively doesn’t validate the server. This seems wrong. In fact, the whole need to enable host key checking by default seems wrong. Host key check should be enabled by default (so by default users are then forced to use secure/validated connections), and if it is needed to test locally without ssh host key checking, let the user turn on no_host_key_check: True in the connection details to override the default. It seems a safer approach.
- Users are still having to write out to the docker containers `~/.ssh/config` file as part of their work because one of the servers they connect to is using a much older key exchange algorithm that the ssh libraries don’t check for by default, so they have to tell it which one to use for a specific site. Given they still need to write out this file, it would be convenient for them just to update the `~/.ssh/known_hosts` file too, now they know how to do it as part of the Dockerfile. It would make sense to them for the SFTP provider to default back to using the `known_hosts` file if a host key is not specifically provided on the connection. Developers have the choice then of which approach to take.
- Accept `string` values(fetched from XCOM results of the previous task) to the `newer_than` field that need to be converted to `datetime` values in the implementation as currently the `newer_than` field only accepts `datetime` fields.
- Use `readdir` instead of `listdir` to avoid multiple roundtrip for each file to get the modified timestamp of the file.

Based on code review suggestion by @ashb that we could pass `byte string` `known_hosts` to the `asyncssh` client connection for the given connection extras `host_key`, we do not need an additional `paramiko` based host key validation as the asynchssh client usage can validate host keys by itself. This also removes the need for the presence of `known_hosts` files in case of `host_key` provided via connection extras. Thus, we also clean up the `paramiko` based validation added in the previous PR #963 

closes: #1021

### DIFF
--- a/astronomer/providers/sftp/hooks/sftp.py
+++ b/astronomer/providers/sftp/hooks/sftp.py
@@ -20,7 +20,7 @@ class SFTPHookAsync(BaseHook):
     :param host: hostname of the SFTP server
     :param port: port of the SFTP server
     :param username: username used when authenticating to the SFTP server
-    :param password: password used when authenticating to the SFTP server
+    :param password: password used when authenticating to the SFTP server.
         Can be left blank if using a key file
     :param known_hosts: path to the known_hosts file on the local file system. Defaults to ``~/.ssh/known_hosts``.
     :param key_file: path to the client key file used for authentication to SFTP server
--- a/astronomer/providers/sftp/hooks/sftp.py
+++ b/astronomer/providers/sftp/hooks/sftp.py
@@ -69,56 +69,41 @@ class SFTPHookAsync(BaseHook):
         self.no_host_key_check = False
         self.host_key = None
 
-    async def _get_conn(self) -> asyncssh.SSHClientConnection:
-        """
-        Asynchronously connect to the SFTP server as an SSH client
+    def _parse_extras(self, conn: Connection) -> None:
+        """Parse extra fields from the connection into instance fields"""
+        extra_options = conn.extra_dejson
+        if "key_file" in extra_options and self.key_file == "":
+            self.key_file = extra_options.get("key_file")
+        if "known_hosts" in extra_options:
+            self.known_hosts = extra_options.get("known_hosts")
+        if ("passphrase" or "private_key_passphrase") in extra_options:
+            self.passphrase = extra_options.get("passphrase")
+        if "private_key" in extra_options:
+            self.private_key = extra_options.get("private_key")
 
-        The following parameters are provided either in the extra json object in
-        the SFTP connection definition
+        host_key = extra_options.get("host_key")
+        no_host_key_check = extra_options.get("no_host_key_check")
 
-        - key_file
-        - known_hosts
-        - passphrase
-        """
-        conn = await sync_to_async(self.get_connection)(self.sftp_conn_id)
-        if conn.extra is not None:
-            extra_options = conn.extra_dejson
-            if "key_file" in extra_options and self.key_file == "":
-                self.key_file = extra_options.get("key_file")
-            if "known_hosts" in extra_options:
-                self.known_hosts = extra_options.get("known_hosts")
-            if ("passphrase" or "private_key_passphrase") in extra_options:
-                self.passphrase = extra_options.get("passphrase")
-            if "private_key" in extra_options:
-                self.private_key = extra_options.get("private_key")
+        if no_host_key_check is not None:
+            no_host_key_check = str(no_host_key_check).lower() == "true"
+            if host_key is not None and no_host_key_check:
+                raise ValueError("Must check host key when provided.")
+            self.no_host_key_check = no_host_key_check
 
-            host_key = extra_options.get("host_key")
-            no_host_key_check = extra_options.get("no_host_key_check")
+        if host_key is not None:
+            if host_key.startswith("ssh-"):
+                key_type, host_key = host_key.split(None)[:2]
+                key_constructor = self._host_key_mappings[key_type[4:]]
+            elif host_key.startswith("ecdsa-"):
+                _, host_key = host_key.split(None)[:2]
+                key_constructor = paramiko.ECDSAKey
+            else:
+                key_constructor = paramiko.RSAKey
+            decoded_host_key = decodebytes(host_key.encode("utf-8"))
+            self.host_key = key_constructor(data=decoded_host_key)
 
-            if no_host_key_check is not None:
-                no_host_key_check = str(no_host_key_check).lower() == "true"
-                if host_key is not None and no_host_key_check:
-                    raise ValueError("Must check host key when provided.")
-                self.no_host_key_check = no_host_key_check
-
-            if host_key is not None:
-                if host_key.startswith("ssh-"):
-                    key_type, host_key = host_key.split(None)[:2]
-                    key_constructor = self._host_key_mappings[key_type[4:]]
-                elif host_key.startswith("ecdsa-"):
-                    _, host_key = host_key.split(None)[:2]
-                    key_constructor = paramiko.ECDSAKey
-                else:
-                    key_constructor = paramiko.RSAKey
-                decoded_host_key = decodebytes(host_key.encode("utf-8"))
-                self.host_key = key_constructor(data=decoded_host_key)
-
-        if self.no_host_key_check:
-            self.log.warning("No Host Key Verification. This won't protect against Man-In-The-Middle attacks")
-            self.known_hosts = "none"
-        else:
-            self._validate_host_key_using_paramiko(conn)
-
+    async def _get_asyncssh_client_connection(self, conn: Connection) -> asyncssh.SSHClientConnection:
+        """Return an asyncssh client connection object for the given connection"""
         conn_config = {
             "host": conn.host,
             "port": conn.port,
@@ -137,8 +122,32 @@ class SFTPHookAsync(BaseHook):
             conn_config.update(client_keys=[_private_key])
         if self.passphrase:
             conn_config.update(passphrase=self.passphrase)
-        ssh_client = await asyncssh.connect(**conn_config)
-        return ssh_client
+        ssh_client_conn = await asyncssh.connect(**conn_config)
+        return ssh_client_conn
+
+    async def _get_conn(self) -> asyncssh.SSHClientConnection:
+        """
+        Asynchronously connect to the SFTP server as an SSH client
+
+        The following parameters are provided either in the extra json object in
+        the SFTP connection definition
+
+        - key_file
+        - known_hosts
+        - passphrase
+        """
+        conn = await sync_to_async(self.get_connection)(self.sftp_conn_id)
+        if conn.extra is not None:
+            self._parse_extras(conn)
+
+        if self.no_host_key_check:
+            self.log.warning("No Host Key Verification. This won't protect against Man-In-The-Middle attacks")
+            self.known_hosts = "none"
+        else:
+            self._validate_host_key_using_paramiko(conn)
+
+        ssh_client_conn = await self._get_asyncssh_client_connection(conn)
+        return ssh_client_conn
 
     def _validate_host_key_using_paramiko(self, conn: Connection) -> None:
         """
@@ -225,8 +234,8 @@ class SFTPHookAsync(BaseHook):
         self, path: str = "", fnmatch_pattern: str = ""
     ) -> Sequence[asyncssh.sftp.SFTPName]:
         """
-        Returns the files along with their attributes matching the file pattern at the provided path, if one exists
-        Otherwise, raises an AirflowException to be handled upstream for deferring
+        Returns the files along with their attributes matching the file pattern (e.g. *.pdf) at the provided path,
+        if one exists. Otherwise, raises an AirflowException to be handled upstream for deferring
         """
         files_list = await self.read_directory(path)
         if files_list is None:
--- a/astronomer/providers/sftp/hooks/sftp.py
+++ b/astronomer/providers/sftp/hooks/sftp.py
@@ -73,13 +73,13 @@ class SFTPHookAsync(BaseHook):
         """Parse extra fields from the connection into instance fields"""
         extra_options = conn.extra_dejson
         if "key_file" in extra_options and self.key_file == "":
-            self.key_file = extra_options.get("key_file")
+            self.key_file = extra_options["key_file"]
         if "known_hosts" in extra_options:
-            self.known_hosts = extra_options.get("known_hosts")
+            self.known_hosts = extra_options["known_hosts"]
         if ("passphrase" or "private_key_passphrase") in extra_options:
-            self.passphrase = extra_options.get("passphrase")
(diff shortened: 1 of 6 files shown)


astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -34,7 +36,7 @@ class SFTPHookAsync(BaseHook):
     default_conn_name = "sftp_default"
     conn_type = "sftp"
     hook_name = "SFTP"
-    default_known_hosts = "none"
+    default_known_hosts = "~/.ssh/known_hosts"
```
ashb, line 39: Does this cause any problems if this file doesn't exist?

astronomer/providers/sftp/sensors/sftp.py
```diff
@@ -43,18 +42,17 @@ def execute(self, context: Context) -> None:
         Logic that the sensor uses to correctly identify which trigger to
         execute, and defer execution as expected.
         """
-        if not self.poke(context=context):
```
ashb, line 46: Why did you remove the sync check before deferring? Probably worth adding a comment saying why we don't follow that pattern for this Sensor

tests/sftp/hooks/test_sftp.py
```diff
@@ -20,6 +23,12 @@ async def listdir(self, path: str):
         else:
             return ["..", ".", "file"]
 
+    async def readdir(self, path: str):
```
phanikumv, line 26:
```suggestion
    async def read_dir(self, path: str):
```

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -196,15 +211,27 @@ async def list_directory(self, path: str = "") -> list[str] | None:
         except asyncssh.SFTPNoSuchFile:
             return None
 
-    async def get_files_by_pattern(self, path: str = "", fnmatch_pattern: str = "") -> list[str]:
+    async def read_directory(self, path: str = "") -> Sequence[asyncssh.sftp.SFTPName] | None:
+        """Returns a list of files along with their attributes on the SFTP server at the provided path"""
+        ssh_conn = await self._get_conn()
+        sftp_client = await ssh_conn.start_sftp_client()
+        try:
+            files = await sftp_client.readdir(path)
```
phanikumv, line 219:
```suggestion
            files = await sftp_client.read_dir(path)
```

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -78,113 +80,126 @@ async def _get_conn(self) -> asyncssh.SSHClientConnection:
...
-                        before_sleep=log_before_sleep,
-                    ):
-                        with attempt:
-                            paramiko_client.connect(**connect_kwargs)
-
-            conn_config = {
-                "host": conn.host,
-                "port": conn.port,
+        conn = await sync_to_async(self.get_connection)(self.sftp_conn_id)
+        if conn.extra is not None:
```
sunank200, line 84: Can this if and else go to a separate method? WDYT? It would make it cleaner

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -78,113 +80,126 @@ async def _get_conn(self) -> asyncssh.SSHClientConnection:
...
+                    key_constructor = self._host_key_mappings[key_type[4:]]
+                elif host_key.startswith("ecdsa-"):
+                    _, host_key = host_key.split(None)[:2]
+                    key_constructor = paramiko.ECDSAKey
+                else:
+                    key_constructor = paramiko.RSAKey
+                decoded_host_key = decodebytes(host_key.encode("utf-8"))
+                self.host_key = key_constructor(data=decoded_host_key)
+
+        if self.no_host_key_check:
```
sunank200, line 116: I feel rather than having multiple if else at same place we should split it into corresponding methods.

phanikumv (commented):
are there no changes needed to the self sufficient DAG?

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -196,15 +211,27 @@ async def list_directory(self, path: str = "") -> list[str] | None:
...
+            return files
+        except asyncssh.SFTPNoSuchFile:
+            return None
+
+    async def get_files_by_pattern(
+        self, path: str = "", fnmatch_pattern: str = ""
+    ) -> Sequence[asyncssh.sftp.SFTPName]:
         """
-        Returns the name of a file matching the file pattern at the provided path, if one exists
+        Returns the files along with their attributes matching the file pattern at the provided path, if one exists
```
sunank200, line 228: An example of pattern in doc-string would be useful

astronomer/providers/sftp/hooks/sftp.py
pankajkoti: yes, it expects this file to be present on the server. And complains that no known_hosts file found when trying to operate. We're mandating `host_key` check now by default as opposed to the sync version which by allows to proceed if no host key is given by just logging a warning for MITM attacks.

If a host key is not provided in the conn extras, we check for the host key in this file. If the needed host key is not provided in the connection extras or present in this file, we raise an exception. If the user wants to bypass this check, they can set `{no_host_key_check: true}` in the connection extras and in that case if this file is not present, it does not cause an issue.

tests/sftp/hooks/test_sftp.py
pankajkoti: We're mocking the `readdir` method of the `asyncssh.sftp` client. Since this method has been declared this way, we have to mock it that way.

pankajkoti:
> are there no changes needed to the self sufficient DAG?

yes, no changes needed in the self sufficient DAG. I tested the same example DAG that we have for the changes here. The configurations are mostly related to connection extras changes and other changes have been covered by the unit tests.

astronomer/providers/sftp/hooks/sftp.py
ashb: If I'm reading the code right (and I may not be) then we will call `_validate_host_key_using_paramiko` and that will always call `load_host_keys` -- this makes this a breaking change as it requires this file to exist even when host key was specified in the extras.

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -23,9 +26,8 @@ class SFTPHookAsync(BaseHook):
     :param port: port of the SFTP server
     :param username: username used when authenticating to the SFTP server
     :param password: password used when authenticating to the SFTP server
-                     Can be left blank if using a key file
-    :param known_hosts: path to the known_hosts file on the local file system
-                     If known_hosts is set to the literal "none", then no host verification is performed
+        Can be left blank if using a key file
```
utkarsharma2, line 29: is this required for encrypted key files?

astronomer/providers/sftp/hooks/sftp.py
ashb: This file must not be required if a hostkey is provided in extra - this makes it a DAG breaking change.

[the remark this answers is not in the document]
pankajkoti: did not understand the comment. what do we mean by getting used?

pankajkoti: As far as I could check, the asyncssh client that we use in the triggerer can only use host keys given in the known_hosts file unlike the paramiko client instance with which we can add a key to the client on top of the known_hosts file.

So, the approach I am taking is to validate the extras host_key using paramiko and using paramiko add the host key back to the known_hosts file so that the asyncssh client can perform its operations.

Do you have a suggestion for me on how we could tackle this scenario?

astronomer/providers/sftp/sensors/sftp.py
pankajkoti: I have added a comment describing the reason. Please check.

astronomer/providers/sftp/hooks/sftp.py
pankajkoti: hi @ashb , I believe I have addressed the concerns and verified that it does not break anymore asking for the `known_hosts` file to be present when connection `extras` `host_key` is provided in the connection. We have users waiting for these issues to be resolved that are being addressed in this PR and also have an open Zendesk ticket for a long time. So, as discussed in the providers standup today, we're thinking to go ahead with this and create a release. 
If you still foresee some issues, we can have a follow up PR. 

cc: @phanikumv

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -23,9 +21,8 @@ class SFTPHookAsync(BaseHook):
     :param port: port of the SFTP server
     :param username: username used when authenticating to the SFTP server
     :param password: password used when authenticating to the SFTP server
```
ashb, line 23:
```suggestion
    :param password: password used when authenticating to the SFTP server.
```

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -23,9 +21,8 @@ class SFTPHookAsync(BaseHook):
...
-                     Can be left blank if using a key file
-    :param known_hosts: path to the known_hosts file on the local file system
-                     If known_hosts is set to the literal "none", then no host verification is performed
+        Can be left blank if using a key file
+    :param known_hosts: path to the known_hosts file on the local file system. Defaults to ~/.ssh/known_hosts.
```
ashb, line 25:
```suggestion
    :param known_hosts: path to the known_hosts file on the local file system. Defaults to ``~/.ssh/known_hosts``.
```

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -60,12 +50,38 @@ def __init__(  # nosec: B107
         self.port = port
         self.username = username
         self.password = password
-        self.known_hosts = known_hosts
+        self.known_hosts: bytes | str = os.path.expanduser(known_hosts)
```
ashb, line 53: Non-blocking nit: Having this be a bytes or a string doesn't feel "right" -- like what we store on `self` should probably be  either the known hosts path, or the `SSHKnownHosts` object depending on when we want/need/can parse it.

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -60,12 +50,38 @@ def __init__(  # nosec: B107
...
         self.private_key = private_key
-        self.no_host_key_check = True
-        self.host_key = None
+
+    def _parse_extras(self, conn: Connection) -> None:
+        """Parse extra fields from the connection into instance fields"""
+        extra_options = conn.extra_dejson
+        if "key_file" in extra_options and self.key_file == "":
+            self.key_file = extra_options["key_file"]
+        if "known_hosts" in extra_options:
```
ashb, line 63: Only take extra if value not explicitly passed like we do with key file?

```suggestion
        if "known_hosts" in extra_options and self.known_hosts != self.default_known_hosts:
```

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -60,12 +50,38 @@ def __init__(  # nosec: B107
...
+        if "private_key" in extra_options:
+            self.private_key = extra_options["private_key"]
+
+        host_key = extra_options.get("host_key")
+        no_host_key_check = extra_options.get("no_host_key_check")
+
+        if no_host_key_check is not None:
+            no_host_key_check = str(no_host_key_check).lower() == "true"
+            if host_key is not None and no_host_key_check:
+                raise ValueError("Must check host key when provided.")
```
ashb, line 76:
```suggestion
                raise ValueError("Host key check was requested, but no `host_key` value was given")
```

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -60,12 +50,38 @@ def __init__(  # nosec: B107
...
+            self.passphrase = extra_options["passphrase"]
+        if "private_key" in extra_options:
+            self.private_key = extra_options["private_key"]
+
+        host_key = extra_options.get("host_key")
+        no_host_key_check = extra_options.get("no_host_key_check")
+
+        if no_host_key_check is not None:
+            no_host_key_check = str(no_host_key_check).lower() == "true"
+            if host_key is not None and no_host_key_check:
```
ashb, line 75: I'm confused how ~host_key` and `known_hosts` interact -- what is the difference between them? Why do we need two options?

ashb: If someone is being extra careful and sets `no_host_key_check=False`, this will require host_key to be set, by-and not let the connection use the value given in known_hosts.

astronomer/providers/sftp/hooks/sftp.py
```diff
@@ -196,6 +129,29 @@ async def list_directory(self, path: str = "") -> list[str] | None:
...
+    async def get_files_and_attrs_by_pattern(
+        self, path: str = "", fnmatch_pattern: str = ""
+    ) -> Sequence[asyncssh.sftp.SFTPName]:
+        """
+        Returns the files along with their attributes matching the file pattern (e.g. ``*.pdf``) at the provided path,
+        if one exists. Otherwise, raises an AirflowException to be handled upstream for deferring
+        """
+        files_list = await self.read_directory(path)
+        if files_list is None:
+            raise AirflowException(f"No files at path {path} found...")
```
ashb, line 151: Nit: repr to quote it

```suggestion
            raise AirflowException(f"No files at path {path!r} found...")
```

ashb: Wonder if this should be 
```suggestion
            raise NotADirectoryError(f"No files at path {path!r} found...")
```

There is no reason it has to be AirflowException -- any exception is valid

ashb (commented):
A few nits, one observation and a question.

ashb (approved):
Approved - but look at my suggestions before merging.

astronomer/providers/sftp/hooks/sftp.py
pankajkoti: for the host_key provided in the connection extras we're passing it as a byte string as is also accepted by the client and not a SSHKnownHosts objects imported from string.

astronomer/providers/sftp/hooks/sftp.py
pankajkoti: We discussed this over a call and verified the scenario is working well.

astronomer/providers/sftp/hooks/sftp.py
pankajkoti: and os.path.expanduser(known_hosts) results to a string.

Screenshot 2023-05-19 at 2 43 29 PM

PR #1073, opened 2023-05-10
Check whether the task finishes before deferring the task for DataprocCreateClusterOperatorAsync

Lee-W:

### DIFF
--- a/astronomer/providers/google/cloud/operators/dataproc.py
+++ b/astronomer/providers/google/cloud/operators/dataproc.py
@@ -16,6 +16,7 @@ from airflow.providers.google.cloud.operators.dataproc import (
     DataprocUpdateClusterOperator,
 )
 from google.api_core.exceptions import AlreadyExists
+from google.cloud.dataproc_v1 import Cluster
 
 from astronomer.providers.google.cloud.triggers.dataproc import (
     DataprocCreateClusterTrigger,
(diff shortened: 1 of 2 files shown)


Lee-W:
圖片

astronomer/providers/google/cloud/operators/dataproc.py
```diff
@@ -96,24 +97,30 @@ def execute(self, context: Context) -> None:  # type: ignore[override]
...
-                impersonation_chain=self.impersonation_chain,
-                polling_interval=self.polling_interval,
-            ),
-            method_name="execute_complete",
+        cluster = hook.get_cluster(
+            project_id=self.project_id, region=self.region, cluster_name=self.cluster_name
         )
+        if cluster.status.state == cluster.status.State.RUNNING:
+            self.log.info("Cluster created.")
+            return Cluster.to_dict(cluster)
```
pankajkoti, line 105: We do not do the the same in deferrable mode. This would get saved as XCOM result. If we wan't to have it here, we should also add in the the deferrable mode's output and verify that the XCOM results are consistent.

pankajkoti (commented):
Looks good to me. Have a minor comment.

astronomer/providers/google/cloud/operators/dataproc.py
Lee-W: Sure. Let me check what would be the best.

pankajkoti: perfect. Sorry, did not check this.

pankajkoti (approved):

PR #1076, opened 2023-05-11
Check whether the task finishes before deferring the task for DataprocDeleteClusterOperatorAsync

Lee-W:


Lee-W:
圖片

astronomer/providers/google/cloud/operators/dataproc.py
```diff
@@ -190,22 +190,29 @@ def execute(self, context: Context) -> None:
             metadata=self.metadata,
         )
 
-        end_time: float = time.time() + self.timeout
+        try:
+            hook.get_cluster(project_id=self.project_id, region=self.region, cluster_name=self.cluster_name)
+        except NotFound:
+            return
```
pankajkoti, line 196: We should log a message/warning here that the cluster was not found.

astronomer/providers/google/cloud/operators/dataproc.py
```diff
@@ -190,22 +190,29 @@ def execute(self, context: Context) -> None:
...
+        except Exception as e:
+            raise AirflowException(str(e))
+        else:
```
pankajkoti, line 199: Do we need this else block? If any of the exception occurs the flow will be terminated anyway and it won't reach here. My suggestion is to remove this additional nesting for the new lines if we don't need it.

astronomer/providers/google/cloud/operators/dataproc.py
Lee-W: Just added it. Thanks!

astronomer/providers/google/cloud/operators/dataproc.py
Lee-W: I just removed it. Thanks!

astronomer/providers/google/cloud/operators/dataproc.py
```diff
@@ -190,6 +190,14 @@ def execute(self, context: Context) -> None:
             metadata=self.metadata,
         )
 
+        try:
+            hook.get_cluster(project_id=self.project_id, region=self.region, cluster_name=self.cluster_name)
+        except NotFound:
+            self.log.info("Cluster deleted.")
```
pankajkoti, line 196: I would suggest that we log here that `Cluster not found. Skipping deletion.` instead of saying cluster deleted.

pankajkoti (approved):

astronomer/providers/google/cloud/operators/dataproc.py
Lee-W: I'd said this might make the log less accurate as we're still deleting the cluster [here](https://github.com/astronomer/astronomer-providers/blob/24f2b6100ea0ee5a777f849c0aada19adcaf960a/astronomer/providers/google/cloud/operators/dataproc.py#L183). WDYT

Lee-W: I'll merge it for now. But we can send another PR if we decide to change the log 🙂

pankajkoti: ohh, yes, sorry, you're right. In my mind, I thought we're checking for the cluster status before triggering the delete.

PR #1079, opened 2023-05-11
Check whether the task finishes before deferring the task for BigQueryGetDataOperatorAsync

Lee-W:

### DIFF
--- a/astronomer/providers/google/cloud/operators/bigquery.py
+++ b/astronomer/providers/google/cloud/operators/bigquery.py
@@ -325,20 +325,24 @@ class BigQueryGetDataOperatorAsync(BigQueryGetDataOperator):
         job = self._submit_job(hook, job_id="", configuration=configuration)
         self.job_id = job.job_id
         context["ti"].xcom_push(key="job_id", value=self.job_id)
-        self.defer(
-            timeout=self.execution_timeout,
-            trigger=BigQueryGetDataTrigger(
-                conn_id=self.gcp_conn_id,
-                job_id=self.job_id,
-                dataset_id=self.dataset_id,
-                table_id=self.table_id,
-                project_id=hook.project_id,
-                impersonation_chain=self.impersonation_chain,
-                poll_interval=self.poll_interval,
-                **kwargs,
-            ),
-            method_name="execute_complete",
-        )
+
+        if job.running():
+            self.defer(
+                timeout=self.execution_timeout,
+                trigger=BigQueryGetDataTrigger(
+                    conn_id=self.gcp_conn_id,
+                    job_id=self.job_id,
+                    dataset_id=self.dataset_id,
+                    table_id=self.table_id,
+                    project_id=hook.project_id,
+                    impersonation_chain=self.impersonation_chain,
+                    poll_interval=self.poll_interval,
+                    **kwargs,
+                ),
+                method_name="execute_complete",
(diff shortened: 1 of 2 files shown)


Lee-W:
圖片

astronomer/providers/google/cloud/operators/bigquery.py
```diff
@@ -325,20 +325,24 @@ def execute(self, context: Context) -> None:  # noqa: D102
...
+                    table_id=self.table_id,
+                    project_id=hook.project_id,
+                    impersonation_chain=self.impersonation_chain,
+                    poll_interval=self.poll_interval,
+                    **kwargs,
+                ),
+                method_name="execute_complete",
+            )
+        else:
+            super().execute(context=context)
```
pankajkoti, line 345:
```suggestion
        if not job.running():
            super().execute(context=context)
            return
        self.defer(
            timeout=self.execution_timeout,
            trigger=BigQueryGetDataTrigger(
                conn_id=self.gcp_conn_id,
                job_id=self.job_id,
                dataset_id=self.dataset_id,
                table_id=self.table_id,
                project_id=hook.project_id,
                impersonation_chain=self.impersonation_chain,
                poll_interval=self.poll_interval,
                **kwargs,
            ),
            method_name="execute_complete",
        )
```

pankajkoti, line 345:
```suggestion
        if not job.running():
            super().execute(context=context)
            return
        self.defer(
            timeout=self.execution_timeout,
            trigger=BigQueryGetDataTrigger(
                conn_id=self.gcp_conn_id,
                job_id=self.job_id,
                dataset_id=self.dataset_id,
                table_id=self.table_id,
                project_id=hook.project_id,
                impersonation_chain=self.impersonation_chain,
                poll_interval=self.poll_interval,
                **kwargs,
            ),
            method_name="execute_complete",
        )
```

pankajkoti (approved):
Looks good to me. Have added a minor suggestion if you'd like to consider (I feel it cleaner that way :) )

astronomer/providers/google/cloud/operators/bigquery.py
Lee-W: Because the job is likely still running, I'd prefer writing the case in the if block instead of the else block. Or another way to implement might be something like the following. WDYT?


```suggestion
        if job.running():
            self.defer(
                timeout=self.execution_timeout,
                trigger=BigQueryGetDataTrigger(
                    conn_id=self.gcp_conn_id,
                    job_id=self.job_id,
                    dataset_id=self.dataset_id,
                    table_id=self.table_id,
                    project_id=hook.project_id,
                    impersonation_chain=self.impersonation_chain,
                    poll_interval=self.poll_interval,
                    **kwargs,
                ),
                method_name="execute_complete",
            )
            return

        super().execute(context=context)
```

pankajkoti: my idea was based on the principle of having an Early Return. 
https://medium.com/swlh/return-early-pattern-3d18a41bba8  You can look at the second example here and it follows with some explanation on how it could be better.

Regarding your second suggestion those are so many lines in the if block and I feel it is better to have them outside the nesting. It is one level less nesting with the approach I suggested and could appear cleaner.

PR #1080, opened 2023-05-11
Check whether the task finishes before deferring the task for GKEStartPodOperatorAsync

Lee-W:

### DIFF
--- a/astronomer/providers/databricks/operators/databricks.py
+++ b/astronomer/providers/databricks/operators/databricks.py
@@ -222,7 +222,7 @@ class DatabricksSubmitRunOperatorAsync(DatabricksSubmitRunOperator):
             else:
                 _handle_non_successful_terminal_states(run_state, run_info, hook, self.task_id)
 
-    def execute_complete(self, context: Context, event: Any = None) -> None:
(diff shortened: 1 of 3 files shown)


Lee-W:
> image

This is tested with `apache-airflow-providers-google==10.0.0` and `apache-airflow-providers-cncf-kubernetes @https://files.pythonhosted.org/packages/17/32/e3d4cb2c54ed8e61d5be25f505718de08c3b0a7b296f11862146a402305b/apache-airflow-providers-cncf-kubernetes-7.0.0.tar.gz`. 

`apache-airflow-providers-google==10.0.0`  is incompatible with `apache-airflow-providers-cncf-kubernetes==6.*.*`

pankajkoti (approved):
Looks good to me. IMO, we can merge this if we were able to test this on an already SUCCEEDED/FAILED pod that it completed the task without deferring.

Lee-W:
Thanks @pankajkoti for providing extra information. @hkc-8010 It seems the customer ticket has been closed, and we cannot reproduce this issue. Should we close this ticket? If not, could you please provide us with a clearer guide to reproduce it? Thanks!

Lee-W:
@pankajkoti How about I add a while loop before deferring and wait for the pod state to become SUCCEEDED for testing?

PR #1083, opened 2023-05-11
Disable openlineage for local build of providers

pankajkoti:
We do not have extractors for all the operators that we have in this repo.
As a result, when building the setup locally using docker-compose.yml and
running the DAGs we keep on getting Extractor logs not found in the task 
logs which deviates us from finding the right logs.
With this change, we are disabling setting the openlineage specific envrionment
variables, so that we have clean task logs.


pankajkoti:
**Previous task logs:**

Screenshot 2023-05-11 at 4 51 06 PM


**With the change in this PR, task logs are cleaner:**

Screenshot 2023-05-11 at 4 51 50 PM

phanikumv (approved):

PR #1084, opened 2023-05-12
Check whether the task finishes before deferring the task for DatabricksSubmitRunOperatorAsync

Lee-W:

### DIFF
--- a/astronomer/providers/databricks/operators/databricks.py
+++ b/astronomer/providers/databricks/operators/databricks.py
@@ -1,15 +1,47 @@
+from __future__ import annotations
+
 from typing import Any
 
 from airflow.exceptions import AirflowException
+from airflow.providers.databricks.hooks.databricks import DatabricksHook
 from airflow.providers.databricks.operators.databricks import (
     XCOM_RUN_ID_KEY,
     XCOM_RUN_PAGE_URL_KEY,
     DatabricksRunNowOperator,
     DatabricksSubmitRunOperator,
+    RunState,
 )
 
 from astronomer.providers.databricks.triggers.databricks import DatabricksTrigger
 from astronomer.providers.utils.typing_compat import Context
+
+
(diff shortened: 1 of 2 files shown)


Lee-W:
圖片

pankajkoti (commented):
Same comments as in PR https://github.com/astronomer/astronomer-providers/pull/1085

astronomer/providers/databricks/operators/databricks.py
```diff
@@ -1,17 +1,50 @@
...
 )
 
 from astronomer.providers.databricks.triggers.databricks import DatabricksTrigger
 from astronomer.providers.utils.typing_compat import Context
 
 
+def _handle_non_successful_teminal_states(
+    run_state: RunState, run_info: dict[str, Any], hook: DatabricksHook, task_id: str
+) -> None:
+    """Raise AirflowException with detailed error message from run_info"""
```
pankajkoti, line 22: Can we add some more information here please? Regarding what are those parameters and some more details on what all are we trying to do in this method? Since, this is more custom logic and needs understanding on the API response, it would be good to have more information so some other developer revisiting this later would have it easier to understand how the API response is and what the keys in it mean.

Lee-W: Sure, I tried to add more detail. Could you please check again and see whether it's improved? Thanks!

pankajastro (approved):
LGTM

astronomer/providers/databricks/operators/databricks.py
```diff
@@ -12,6 +15,48 @@
 from astronomer.providers.utils.typing_compat import Context
 
 
+def _handle_non_successful_teminal_states(
```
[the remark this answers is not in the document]
Lee-W: Thanks! I just fixed it.

PR #1085, opened 2023-05-12
Check whether the task finishes before deferring the task for DatabricksRunNowOperatorAsync

Lee-W:

### DIFF
--- a/astronomer/providers/databricks/operators/databricks.py
+++ b/astronomer/providers/databricks/operators/databricks.py
@@ -15,7 +15,7 @@ from astronomer.providers.databricks.triggers.databricks import DatabricksTrigge
 from astronomer.providers.utils.typing_compat import Context
 
 
-def _handle_non_successful_teminal_states(
+def _handle_non_successful_terminal_states(
     run_state: RunState, run_info: dict[str, Any], hook: DatabricksHook, task_id: str
 ) -> None:
     """Raise AirflowException with detailed error message from run_info
@@ -426,7 +426,7 @@ class DatabricksRunNowOperatorAsync(DatabricksRunNowOperator):
                 self.log.info("%s completed successfully.", self.task_id)
                 return
             else:
-                _handle_non_successful_teminal_states(run_state, run_info, hook, self.task_id)
+                _handle_non_successful_terminal_states(run_state, run_info, hook, self.task_id)
 
     def execute_complete(
         self, context: Context, event: Any = None
--- a/astronomer/providers/databricks/operators/databricks.py
+++ b/astronomer/providers/databricks/operators/databricks.py
@@ -1,15 +1,47 @@
+from __future__ import annotations
+
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


Lee-W:
圖片

astronomer/providers/databricks/operators/databricks.py
```diff
@@ -1,17 +1,49 @@
...
 )
 
 from astronomer.providers.databricks.triggers.databricks import DatabricksTrigger
 from astronomer.providers.utils.typing_compat import Context
 
 
+def _handle_not_successful_teminated_state(
+    run_state: RunState, run_info: dict[str, Any], hook: DatabricksHook, task_id: str
+) -> None:
+    if run_state.result_state == "FAILED":
```
pankajkoti, line 22: can we please add a docstring here to tell what all we're doing here? Would be easier to understand and review then.

astronomer/providers/databricks/operators/databricks.py
```diff
@@ -1,17 +1,49 @@
...
     DatabricksRunNowOperator,
     DatabricksSubmitRunOperator,
+    RunState,
 )
 
 from astronomer.providers.databricks.triggers.databricks import DatabricksTrigger
 from astronomer.providers.utils.typing_compat import Context
 
 
+def _handle_not_successful_teminated_state(
```
pankajkoti, line 19:
```suggestion
def _handle_non_successful_teminal_states(
```

astronomer/providers/databricks/operators/databricks.py
```diff
@@ -1,17 +1,50 @@
...
 )
 
 from astronomer.providers.databricks.triggers.databricks import DatabricksTrigger
 from astronomer.providers.utils.typing_compat import Context
 
 
+def _handle_non_successful_teminal_states(
+    run_state: RunState, run_info: dict[str, Any], hook: DatabricksHook, task_id: str
+) -> None:
+    """Raise AirflowException with detailed error message from run_info"""
```
pankajkoti, line 22: Same suggestion as in https://github.com/astronomer/astronomer-providers/pull/1084#discussion_r1193499247

astronomer/providers/databricks/operators/databricks.py
```diff
@@ -12,6 +15,48 @@
 from astronomer.providers.utils.typing_compat import Context
 
 
+def _handle_non_successful_teminal_states(
```
phanikumv, line 18:
```suggestion
def _handle_non_successful_terminal_states(
```

phanikumv (changes requested):

pankajastro (approved):
LGTM

astronomer/providers/databricks/operators/databricks.py
pankajastro: does it makes sense to move this took the hook, not sure but maybe if multiple operators need we can reuse it wdyt?

phanikumv (approved):

PR #1087, opened 2023-05-12
Check whether the task finishes before deferring the task for BatchOperatorAsync

Lee-W:


Lee-W:
圖片

astronomer/providers/amazon/aws/operators/batch.py
```diff
@@ -60,18 +60,29 @@ def execute(self, context: Context) -> None:
...
+        job_status = job.get("status")
+
+        if job_status == self.hook.SUCCESS_STATE:
+            self.log.info(f"{job_id} was completed successfully")
+            return
+
+        if job_status == self.hook.FAILURE_STATE:
+            raise AirflowException(f"{job_id} failed")
+
+        if job_status in self.hook.INTERMEDIATE_STATES:
```
pankajkoti, line 74: Do we need this check? We already check success state and failure state. And if we need this check, what will happen if all the 3 if conditions are not met, the operator will silently exit in that case marking the task as success. So, I think we need to do something here.

Lee-W: Yep, let me add an `AirflowException` if unknown states occurs

PR #1089, opened 2023-05-12
Check whether the task finishes before deferring the task for RedshiftDataOperatorAsync

Lee-W:

### DIFF
--- a/astronomer/providers/amazon/aws/operators/redshift_data.py
+++ b/astronomer/providers/amazon/aws/operators/redshift_data.py
@@ -45,6 +45,7 @@ class RedshiftDataOperatorAsync(RedshiftDataOperator):
         context["ti"].xcom_push(key="return_value", value=query_ids)
 
         still_running = False
+        completed_query_ids = []
         for qid in query_ids:
             resp = redshift_data_hook.conn.describe_statement(Id=qid)
             status = resp["Status"]
@@ -59,8 +60,10 @@ class RedshiftDataOperatorAsync(RedshiftDataOperator):
             elif status in ("SUBMITTED", "PICKED", "STARTED"):
                 still_running = True
                 break
+            elif status == "FINISHED":
+                completed_query_ids.append(qid)
 
-        if not still_running:
+        if not still_running and len(completed_query_ids) == len(query_ids):
             self.log.info("%s completed successfully.", self.task_id)
             return
 
--- a/astronomer/providers/amazon/aws/operators/redshift_data.py
+++ b/astronomer/providers/amazon/aws/operators/redshift_data.py
@@ -43,6 +43,27 @@ class RedshiftDataOperatorAsync(RedshiftDataOperator):
         if response.get("status") == "error":
             self.execute_complete(context, event=response)
         context["ti"].xcom_push(key="return_value", value=query_ids)
+
+        still_running = False
+        for qid in query_ids:
+            resp = redshift_data_hook.conn.describe_statement(Id=qid)
+            status = resp["Status"]
+            if status == "FAILED":
+                err_msg = f"Error: {resp['QueryString']} query Failed due to {resp['Error']}"
+                msg = f"context: {context}, error message: {err_msg}"
+                raise AirflowException(msg)
+            elif status == "ABORTED":
+                err_msg = "The query run was stopped by the user."
+                msg = f"context: {context}, error message: {err_msg}"
+                raise AirflowException(msg)
+            elif status in ("SUBMITTED", "PICKED", "STARTED"):
+                still_running = True
+                break
+
+        if not still_running:
+            self.log.info("%s completed successfully.", self.task_id)
+            return
+
         self.defer(
             timeout=self.execution_timeout,
             trigger=RedshiftDataTrigger(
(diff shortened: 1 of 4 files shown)


Lee-W:
圖片

astronomer/providers/amazon/aws/operators/redshift_data.py
```diff
@@ -43,6 +43,25 @@ def execute(self, context: Context) -> None:  # type: ignore[override]
         if response.get("status") == "error":
             self.execute_complete(context, event=response)
         context["ti"].xcom_push(key="return_value", value=query_ids)
+
+        completed_id = []
```
pankajkoti, line 47:
```suggestion
        completed_query_ids = []
```

astronomer/providers/amazon/aws/operators/redshift_data.py
```diff
@@ -43,6 +43,25 @@ def execute(self, context: Context) -> None:  # type: ignore[override]
...
+            if status == "FAILED":
+                err_msg = f"Error: {resp['QueryString']} query Failed due to {resp['Error']}"
+                msg = f"context: {context}, error message: {err_msg}"
+                raise AirflowException(msg)
+            elif status == "ABORTED":
+                err_msg = "The query run was stopped by the user."
+                msg = f"context: {context}, error message: {err_msg}"
+                raise AirflowException(msg)
+            elif status == "FINISHED":
+                completed_id.append(qid)
```
pankajkoti, line 60: Instead of looping over all the queries to check whether they are finished, would it make sense to break the loop and enter deferral as soon as we find a query that is still running(or not in any of the terminal states)?

Lee-W: Indeed. Let me rewrite it. Thanks for the great suggestion!

astronomer/providers/amazon/aws/operators/redshift_data.py
```diff
@@ -43,6 +43,27 @@ def execute(self, context: Context) -> None:  # type: ignore[override]
...
+            status = resp["Status"]
+            if status == "FAILED":
+                err_msg = f"Error: {resp['QueryString']} query Failed due to {resp['Error']}"
+                msg = f"context: {context}, error message: {err_msg}"
+                raise AirflowException(msg)
+            elif status == "ABORTED":
+                err_msg = "The query run was stopped by the user."
+                msg = f"context: {context}, error message: {err_msg}"
+                raise AirflowException(msg)
+            elif status in ("SUBMITTED", "PICKED", "STARTED"):
```
pankajkoti, line 59: I am sorry but I'm rethinking this. It is good that we're breaking early when we see query is still running. But I think your early approach is still needed to check the count of Finished queries. So we would need another elif condition to also add the Finished queries to check later that the count of finished queries is equal to count of query_ids. The reason I am saying we need this is because Redshift might introduce a new intermediate state later that we don't include here and it would give us a false positive because we won't check that state and then `still_running` would remain False for such a new Intermediate state.

Lee-W: yep, that's indeed something we should consider. Thanks for your suggestion! Just updated it

astronomer/providers/amazon/aws/operators/redshift_data.py
```diff
@@ -43,6 +43,30 @@ def execute(self, context: Context) -> None:  # type: ignore[override]
...
             self.execute_complete(context, event=response)
         context["ti"].xcom_push(key="return_value", value=query_ids)
+
+        still_running = False
+        completed_query_ids = []
+        for qid in query_ids:
+            resp = redshift_data_hook.conn.describe_statement(Id=qid)
+            status = resp["Status"]
+            if status == "FAILED":
+                err_msg = f"Error: {resp['QueryString']} query Failed due to {resp['Error']}"
```
phanikumv, line 53: Can we use format specifiers instead of f-strings

phanikumv (changes requested):

astronomer/providers/amazon/aws/operators/redshift_data.py
```diff
@@ -43,6 +43,30 @@ def execute(self, context: Context) -> None:  # type: ignore[override]
         if response.get("status") == "error":
             self.execute_complete(context, event=response)
         context["ti"].xcom_push(key="return_value", value=query_ids)
+
+        still_running = False
+        completed_query_ids = []
+        for qid in query_ids:
```
pankajastro, line 49: I saw a similar change in PR https://github.com/astronomer/astronomer-providers/pull/1090 does it make sense to make this loop a function?

Lee-W: Yep, I think that's something we can do. I just merge #1090  so that I can extract this logic in this PR

astronomer/providers/amazon/aws/operators/redshift_data.py
pankajkoti: I would prefer f-strings in general unless it is used for logging in self.log.....

phanikumv (approved):

astronomer/providers/amazon/aws/operators/redshift_data.py
Lee-W: @pankajastro just finished the refactor. could you please check? Thanks!

pankajastro: looks good. thanks

pankajastro (approved):
LGTM. need some small fixes for ci :)

PR #1091, opened 2023-05-12
Check whether the task finishes before deferring the task for SageMakerProcessingOperatorAsync

Lee-W:

### DIFF
--- a/astronomer/providers/amazon/aws/operators/sagemaker.py
+++ b/astronomer/providers/amazon/aws/operators/sagemaker.py
@@ -56,7 +56,7 @@ class SageMakerProcessingOperatorAsync(SageMakerProcessingOperator):
         (default) and "fail".
     """
 
-    def execute(self, context: Context) -> None:  # type: ignore[override]
+    def execute(self, context: Context) -> dict[str, str] | None:  # type: ignore[override]
         """
         Creates processing job via sync hook `create_processing_job` and pass the
         control to trigger and polls for the status of the processing job in async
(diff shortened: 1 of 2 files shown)


Lee-W:
圖片

Lee-W:
圖片

Lee-W:
圖片

Makefile
```diff
@@ -57,7 +57,7 @@ run-tests: ## Run CI tests
 	docker build --build-arg IMAGE_NAME=$(ASTRO_RUNTIME_IMAGE_NAME) -f dev/Dockerfile . -t astronomer-providers-dev
 	docker run -v `pwd`:/usr/local/airflow/astronomer_providers -v `pwd`/dev/.cache:/home/astro/.cache \
 	 	-w /usr/local/airflow/astronomer_providers \
-		--rm -it astronomer-providers-dev -- pytest tests
+		--rm -it astronomer-providers-dev -- pytest --cov astronomer --cov-report=term-missing tests
```
pankajkoti, line 60: Not much objection, however, would be nice to create a separate tiny PR for this. Because this change is unrelated with the title of this PR and if we want to trace the change in the commit history would be more visible if it has its own Title and maybe would be easy to revert that specific PR also if need be later.

pankajkoti (approved):

astronomer/providers/amazon/aws/operators/sagemaker.py
```diff
@@ -80,22 +82,33 @@ def execute(self, context: Context) -> None:  # type: ignore[override]
...
-            )
 
-    def execute_complete(self, context: Context, event: Any = None) -> Dict[str, Any]:
+        response = self.hook.describe_processing_job(processing_job_name)
+        status = response["ProcessingJobStatus"]
+        if status in self.hook.failed_states:
+            raise AirflowException(f"SageMaker job failed because {response['FailureReason']}")
+        elif status == "Completed":
+            self.log.info(f"{self.task_id} completed successfully.")
+            return {"Training": serialize(response)}
```
pankajkoti, line 92:
```suggestion
            return {"Processing": serialize(response)}
```

Makefile
Lee-W: Yep, this is something I tested locally but forgot to remove 🤦‍♂️ I'll make another PR for it. I think this can be handy for locally testing!

PR #1094, opened 2023-05-12
Check whether the task finishes before deferring the task for EmrContainerOperatorAsync

Lee-W:

### DIFF
--- a/astronomer/providers/amazon/aws/operators/emr.py
+++ b/astronomer/providers/amazon/aws/operators/emr.py
@@ -1,4 +1,6 @@
-from typing import Any, Dict
+from __future__ import annotations
+
+from typing import Any
 
 from airflow.exceptions import AirflowException
 from airflow.providers.amazon.aws.hooks.emr import EmrContainerHook
@@ -30,7 +32,7 @@ class EmrContainerOperatorAsync(EmrContainerOperator):
     :param tags: The tags assigned to job runs. Defaults to None
     """
 
-    def execute(self, context: Context) -> None:
+    def execute(self, context: Context) -> str | None:
         """Deferred and give control to trigger"""
         hook = EmrContainerHook(aws_conn_id=self.aws_conn_id, virtual_cluster_id=self.virtual_cluster_id)
         job_id = hook.submit_job(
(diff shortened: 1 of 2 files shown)


Lee-W:
圖片

astronomer/providers/amazon/aws/operators/emr.py
```diff
@@ -50,6 +52,21 @@ def execute(self, context: Context) -> None:
             # max_tries is deprecated so instead of max_tries using self.max_polling_attempts
             polling_attempts = self.max_polling_attempts
 
+        query_state = hook.check_query_status(job_id)
+        if query_state in hook.SUCCESS_STATES:
+            self.log.info(
+                f"Try : Query execution completed. Final state is {query_state}"
```
pankajkoti, line 58: Isn't it a job that is submitted to EMR? If yes, can we please replace `Query` with `job` in all the logs here? 

This I believe also calls for renaming Airflow OSS upstream hook methods to contain `job` instead of `query` in their name as we're essentially check job status and not query status. 
But, at least let us use `job` instead of `query` in our logs. WDYT?

pankajastro: separate PR should be ok. @pankajkoti ??

pankajastro (approved):

astronomer/providers/amazon/aws/operators/emr.py
sunank200: I think we can include this change as part of this PR. But I am okay if it comes as a separate PR but we should log the ticket.

sunank200 (approved):

astronomer/providers/amazon/aws/operators/emr.py
pankajkoti: I am okay with a separate PR as long as we log a ticket for this cleanup. I would have preferred it to be in this PR  at least for the changes we're making in this PR as we're adding new logs with wrong information. But you feel this leads to inconsistencies in the logs between deferral and non-deferral, then yes, separate PR could make sense, but please create a ticket :) .

Lee-W: I just created a separate issue and assigned it to myself. I'd like to close adding this check logic ASAP and work on the deferrable auto-switch earlier. But I think I will be able to work on this one during building the testing image for deferrable auto-swtich.

https://github.com/astronomer/astronomer-providers/issues/1105

PR #1095, opened 2023-05-13
Upgrade astro runtime image to dev 8.2.0-alpha3-base

pankajkoti:
Test corresponding runtime image for Airflow 2.6.1rc2


Lee-W (approved):
Once tested, I think we're good to merge this one. Otherwise, it might block our scheduled deployment?

PR #1096, opened 2023-05-15
Upgrade astro runtime image to 8.2.0-alpha4-base

pankajkoti:
Test corresponding runtime image for Airflow 2.6.1rc3


pankajkoti:
We tested this on both CeleryExecutor and KubernetesExecutor deployments. All DAGs ran fine except for the `example_async_batch` dag. The failure is unrelated to the RC and the fix is in PR https://github.com/astronomer/astronomer-providers/pull/1098

Lee-W (approved):
LGTM. As #1098 has been merged, I'll merge this one as well

Issue #1097, opened 2023-05-15
Example Async Batch DAG failing 


pankajkoti:
The DAG has started failing with the below error log
```
[2023-05-15, 17:50:14 UTC] {taskinstance.py:1824} ERROR - Task failed with exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/astronomer/providers/amazon/aws/operators/batch.py", line 64, in execute
    job = self.hook.get_job_description(job_id)
  File "/usr/local/lib/python3.10/site-packages/airflow/providers/amazon/aws/hooks/batch_client.py", line 365, in get_job_description
    response = self.get_conn().describe_jobs(jobs=[job_id])
  File "/usr/local/lib/python3.10/site-packages/botocore/client.py", line 530, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/usr/local/lib/python3.10/site-packages/botocore/client.py", line 919, in _make_api_call
    request_dict = self._convert_to_request_dict(
  File "/usr/local/lib/python3.10/site-packages/botocore/client.py", line 990, in _convert_to_request_dict
    request_dict = self._serializer.serialize_to_request(
  File "/usr/local/lib/python3.10/site-packages/botocore/validate.py", line 381, in serialize_to_request
    raise ParamValidationError(report=report.generate_report())
botocore.exceptions.ParamValidationError: Parameter validation failed:
Invalid type for parameter jobs[0], value: None, type: <class 'NoneType'>, valid types: <class 'str'>
```

PR #1098, opened 2023-05-15
Use `self.job_id` to describe Batch job

pankajkoti:
With the change in PR https://github.com/astronomer/astronomer-providers/pull/1087, we called `submit_job` and assigned its output to a local variable `job_id` and then used this local variable `job_id` to describe on the job subsequently by passing it as parameter to the describe method. But, the `submit_job` method implicitly assigns the `job_id` as an instance attribute and it returns `None`. Hence, `self.job_id` actually contains the `job_id` and the local variable `job_id` which was capturing the output of `submit_job` was assigned `None` value and hence the subsequent method to describe job was failing as it was calling the describe on a `None` (job_id) type value.

The commit fixes this issue by referencing `self.job_id` instead of using the `None` valued local variable `job_id`

closes: #1097.

### DIFF
--- a/astronomer/providers/amazon/aws/operators/batch.py
+++ b/astronomer/providers/amazon/aws/operators/batch.py
@@ -61,6 +61,8 @@ class BatchOperatorAsync(BatchOperator):
         Submit the job and get the job_id using which we defer and poll in trigger
         """
         self.submit_job(context)
+        if not self.job_id:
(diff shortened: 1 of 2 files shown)


pankajkoti:
Successful DAG run post this change.

Screenshot 2023-05-16 at 12 14 50 AM

Lee-W (approved):

Issue #1099, opened 2023-05-15
AWS Batch example DAG marked as successful although few tasks failed


pankajkoti:
<img width="665" alt="Screenshot 2023-05-16 at 12 19 57 AM" src="https://github.com/astronomer/astronomer-providers/assets/10206082/dc34c466-579c-4987-94d9-11596298a88d">

The screenshot above shows that the final DAG status was marked as successful although a few tasks failed. Even if one task fails, we should mark the DAG status as false. As a result, the Master DAG reports the DAG as successful in the Grid view, but the tasks that failed are only seen when we go to the Triggered DAG's grid view.
We should fail the DAG even if one task fails in the DAG.

PR #1100, opened 2023-05-15
Mark AWS Batch DAG failed if any of the tasks fail

pankajkoti:
fail.

closes: #1099

### DIFF
--- a/astronomer/providers/amazon/aws/example_dags/example_batch.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_batch.py
@@ -5,11 +5,13 @@ import time
 from datetime import datetime
 from json import loads
 from os import environ
+from typing import Any
 
 from airflow import DAG, AirflowException
 from airflow.operators.bash import BashOperator
-from airflow.operators.dummy import DummyOperator
 from airflow.operators.python import PythonOperator
+from airflow.utils.state import State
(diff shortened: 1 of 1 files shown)


pankajkoti:
Screenshot 2023-05-16 at 12 30 25 AM

astronomer/providers/amazon/aws/example_dags/example_batch.py
```diff
@@ -236,6 +238,16 @@ def delete_job_queue_func() -> None:
         raise error
 
 
+def check_dag_status(**kwargs: Any) -> None:
+    """Raises an exception if any of the DAG's tasks failed and as a result marking the DAG failed."""
+    for task_instance in kwargs["dag_run"].get_task_instances():
+        if (
+            task_instance.current_state() != State.SUCCESS
+            and task_instance.task_id != kwargs["task_instance"].task_id
```
Lee-W, line 246: May I know in what situation would `task_instance.task_id != kwargs["task_instance"].task_id`?

Lee-W (approved):

astronomer/providers/amazon/aws/example_dags/example_batch.py
pankajkoti: For the current task itself that is being executed i.e. dag_final_status task itself.

PR #1104, opened 2023-05-16
Check whether the task finishes before deferring the task for KubernetesPodOperatorAsync

Lee-W:

### DIFF
--- a/astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py
+++ b/astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py
@@ -5,9 +5,9 @@ from airflow.exceptions import AirflowException, TaskDeferred
 from airflow.providers.cncf.kubernetes.operators.kubernetes_pod import (
     KubernetesPodOperator,
 )
-from airflow.providers.cncf.kubernetes.triggers.kubernetes_pod import (
-    ContainerState,
-    KubernetesPodTrigger,
+from airflow.providers.cncf.kubernetes.utils.pod_manager import (
+    PodPhase,
+    container_is_running,
 )
 from kubernetes.client import models as k8s
 from pendulum import DateTime
@@ -54,21 +54,6 @@ class KubernetesPodOperatorAsync(KubernetesPodOperator):
             else:
                 raise AirflowException(description)
 
-    def define_container_state(self) -> ContainerState:
-        """Define the state of container"""
-        pod_containers = self.pod.status.container_statuses
-        if not pod_containers:
-            return ContainerState.UNDEFINED
-        container = [c for c in pod_containers if c.name == self.BASE_CONTAINER_NAME][0]
-        for state in (ContainerState.RUNNING, ContainerState.WAITING, ContainerState.TERMINATED):
-            state_obj = getattr(container.state, state)
-            if state_obj:
-                if state != ContainerState.TERMINATED:
-                    return state
-                else:
-                    return ContainerState.TERMINATED if state_obj.exit_code == 0 else ContainerState.FAILED
-        return ContainerState.UNDEFINED
-
     def defer(self, last_log_time: Optional[DateTime] = None, **kwargs: Any) -> None:
         """Defers to ``WaitContainerTrigger`` optionally with last log time."""
         if kwargs:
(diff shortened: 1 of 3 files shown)


Lee-W:
圖片

astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py
```diff
@@ -76,10 +95,30 @@ def defer(self, last_log_time: Optional[DateTime] = None, **kwargs: Any) -> None
...
         self.pod_request_obj = self.build_pod_request_obj(context)
         self.pod: k8s.V1Pod = self.get_or_create_pod(self.pod_request_obj, context)
-        self.defer()
+        container_state = self.define_container_state()
+        pod_status = self.pod.status.phase
+        if KubernetesPodTrigger.should_wait(pod_status, container_state):
+            self.defer()
+            return
+
+        if container_state == ContainerState.TERMINATED:
```
pankajastro, line 107: does pod TERMINATED state mean always a success?

Lee-W: Yes, I think so. This is how we decide the state in OSS airflow https://github.com/apache/airflow/blob/main/airflow/providers/cncf/kubernetes/triggers/pod.py#L135 as well

astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py
```diff
@@ -76,10 +95,30 @@ def defer(self, last_log_time: Optional[DateTime] = None, **kwargs: Any) -> None
...
+            event = {
+                "name": self.pod.metadata.name,
+                "namespace": self.pod.metadata.namespace,
+                "status": "success",
+                "message": "All containers inside pod have started successfully.",
```
pankajkoti, line 112: Is the message correct here to say containers have started?

astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py
```diff
@@ -76,10 +95,30 @@ def defer(self, last_log_time: Optional[DateTime] = None, **kwargs: Any) -> None
...
+            }
+        else:
+            event = {
+                "name": self.pod.metadata.name,
+                "namespace": self.pod.metadata.namespace,
+                "status": "failed",
+                "message": self.pod.status.message,
+            }
+        return self.trigger_reentry(context=context, event=event)
```
pankajkoti, line 121: is the event payload that we send to trigger_reentry handled correctly? I am doubting if it is consistent with how we return it from the triggerer and that it gets handled correctly.

astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py
Lee-W: This has been removed

astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py
Lee-W: Thanks for the reminder. In the original implementation, I utilized the logic in oss airflow but I made a change yesterday to use the WaitContainerTrigger logic instead.

This has been tested as well 

image

astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py
```diff
@@ -76,9 +80,20 @@ def defer(self, last_log_time: Optional[DateTime] = None, **kwargs: Any) -> None
             method_name=self.trigger_reentry.__name__,
         )
 
-    def execute(self, context: Context) -> None:  # noqa: D102
+    def execute(self, context: Context) -> Any:  # noqa: D102
         self.pod_request_obj = self.build_pod_request_obj(context)
         self.pod: k8s.V1Pod = self.get_or_create_pod(self.pod_request_obj, context)
+        pod_status = self.pod.status.phase
+        if pod_status in PodPhase.terminal_states or not container_is_running(
+            pod=self.pod, container_name=self.BASE_CONTAINER_NAME
```
pankajkoti, line 88: Will there always be a base container in all the pods? Also, if there is a base container what does it do? I am thinking if there are multiple containers in the pod then should we not check for the running status of all the containers in the pod? Or the base container is meant to keep a check on the running status of other containers in the pod?

If it's not possible to check the status of all containers I think we could just remove the `or` condition to check the container status and then the PR looks good to me.

Lee-W: I think so 🤔 That's how we check when deferred.

https://github.com/astronomer/astronomer-providers/blob/main/astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py#L69
https://github.com/astronomer/astronomer-providers/blob/main/astronomer/providers/cncf/kubernetes/triggers/wait_container.py#L109

pankajkoti: I am afraid we have a wrong implementation there with respect to the above questions. It's done a bit differently in the OSS provider, no?

pankajkoti (approved):
I don't have a strong reservation on this. Since this is consistent with the existing trigger implementation, other issues would be outside the scope of the PR.

astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py
Lee-W: yep, I think we implement the logic in different ways

Lee-W:
Thanks! I'll merge it for now. As [deferrable mode](https://github.com/apache/airflow/blob/d8086a3db5ef020ef28158249105e7ce5639b1a5/airflow/providers/cncf/kubernetes/operators/pod.py#L226) has been added in OSS airflow, I think we'll eventually do something like https://github.com/astronomer/astronomer-providers/pull/1169

PR #1110, opened 2023-05-18
ci(dev/scripts): use pip internal to parse RC_PROVIDER_PACKAGES

Lee-W:
According to the discussion with @pankajkoti yesterday, when testing [apache-airflow-providers-cncf-kubernetes](https://pypi.org/project/apache-airflow-providers-cncf-kubernetes/), we cannot use the pinned version but need to use the [source distribution](https://files.pythonhosted.org/packages/a9/59/fcd15c3d3b9b26bab4df980f59b1e86769bf7b18edbd18afca5db973b746/apache-airflow-providers-cncf-kubernetes-6.1.0.tar.gz) instead. However, our current script does not support parsing this kind of string. This PR replace the original string manipulation and uses pip internal to parse the package string instead

### DIFF
--- a/dev/scripts/replace_dependencies.py
+++ b/dev/scripts/replace_dependencies.py
@@ -1,21 +1,30 @@
+from __future__ import annotations
+
 import argparse
 import fileinput
-from re import search, sub
-from typing import List
+from re import sub
+
+from pip._internal.utils.packaging import get_requirement
 
 
-def update_setup_cfg(rc_provider_packages: List[str]):
+def update_setup_cfg(rc_provider_packages: list[str]):
     """
     Replaces the given provider packages in the setup.cfg with the given pinned RC versions.
     :param rc_provider_packages: list of RC provider packages to be replaced
     """
     for package in rc_provider_packages:
-        pinned_package = package.strip()
-        if not search("==", pinned_package):
+        requirement = get_requirement(package)
+        package_name_to_search = requirement.name
+
+        if requirement.specifier:
+            pinned_package = f"{package_name_to_search}{requirement.specifier}"
+        elif requirement.url:
+            pinned_package = f"{package_name_to_search} @{requirement.url}"
+        else:
             raise Exception(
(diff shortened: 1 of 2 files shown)


dev/scripts/requirements.txt
pankajkoti, line 1: are we using this file?

dev/scripts/requirements.txt
```diff
@@ -0,0 +1,3 @@
+pip==23.1.2
+requests==2.30.0
+beautifulsoup4=4.12.2
```
pankajkoti, line 3: where do we need beautifulsoup4?

dev/scripts/replace_dependencies.py
```diff
@@ -1,21 +1,32 @@
...
 
 
-def update_setup_cfg(rc_provider_packages: List[str]):
+def update_setup_cfg(rc_provider_packages: list[str]):
     """
     Replaces the given provider packages in the setup.cfg with the given pinned RC versions.
     :param rc_provider_packages: list of RC provider packages to be replaced
     """
     for package in rc_provider_packages:
+        requirement = get_requirement(package)
```
pankajkoti, line 16: nice

dev/scripts/replace_dependencies.py
```diff
@@ -1,21 +1,32 @@
...
+        package_name_to_search = requirement.name
+
         pinned_package = package.strip()
```
pankajkoti, line 19: We may remove this line.

pankajkoti (approved):
Looks good to me apart from the minor suggestions.

dev/scripts/requirements.txt
Lee-W: This is dep needed for another feature. Just removed it. Thanks!

dev/scripts/replace_dependencies.py
Lee-W: Just removed

Issue #1119, opened 2023-05-21
dependency conflict between astronomer-provider 1.17.0.dev and airflow issues 31322 providers


Lee-W:
**Describe the bug**
dependency conflict between `protobuf<=3.20.0` and  `google-ads>=20.0.0` (probably from google provider)

**To Reproduce**
1. `git checkout rc-test-2023-05-20T05-10-46UTC`
2. `make build-run`

**Expected behavior**
install without error

**Screenshots**

**Desktop (please complete the following information):**
 - OS: ubuntu-20.04
 - Browser 
 - Version 

**Smartphone (please complete the following information):**
 - Device: 
 - OS: 
 - Browser 
 - Version 

**Additional context**
```
WARNING: openlineage-integration-common 0.26.0 does not provide the extra 'sql'
ERROR: Cannot install apache-airflow-providers-google and astronomer-providers[all]==1.17.0.dev1 because these package versions have conflicting dependencies.

The conflict is caused by:
    astronomer-providers[all] 1.17.0.dev1 depends on protobuf<=3.20.0; extra == "all"
    google-ads 21.1.0 depends on protobuf>=4.21.5
    astronomer-providers[all] 1.17.0.dev1 depends on protobuf<=3.20.0; extra == "all"
    google-ads 21.0.0 depends on protobuf>=4.21.5
    astronomer-providers[all] 1.17.0.dev1 depends on protobuf<=3.20.0; extra == "all"
    google-ads 20.0.0 depends on protobuf>=4.21.5

To fix this you could try to:
1. loosen the range of package versions you've specified
2. remove package versions to allow pip attempt to solve the dependency conflict

ERROR: ResolutionImpossible: for help visit https://pip.pypa.io/en/latest/topics/dependency-resolution/#dealing-with-dependency-conflicts
The command '/bin/sh -c pip install  --upgrade --force-reinstall  --no-cache-dir /tmp/astronomer-providers[all]' returned a non-zero code: 1
```

PR #1126, opened 2023-05-22
build(setup.cfg): remove protobuf constraint

Lee-W:
dependency conflict between `protobuf<=3.20.0` and `google-ads>=20.0.0` after we upgrade google provider to [10.1.0rc2](https://pypi.org/project/apache-airflow-providers-google/10.1.0rc2)

Close: #1119

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -64,7 +64,6 @@ google =
     apache-airflow-providers-google>=8.1.0
     gcloud-aio-storage
     gcloud-aio-bigquery
-    protobuf<=3.20.0  # Bigquery provider isn't compatible with it. Details in https://github.com/apache/airflow/commit/25a9ae3b2eec85dfd500b0a921045fc95ab8ffd6
 http =
     apache-airflow-providers-http
 microsoft.azure =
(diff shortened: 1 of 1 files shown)


setup.cfg
```diff
@@ -137,7 +137,6 @@ all =
     impyla
     openlineage-airflow>=0.12.0
     paramiko
-    protobuf<=3.20.0  # Bigquery provider isn't compatible with it. Details in https://github.com/apache/airflow/commit/25a9ae3b2eec85dfd500b0a921045fc95ab8ffd6
```
pankajkoti, line 140: We would also need to remove this from the Google extras here https://github.com/astronomer/astronomer-providers/pull/1126/files#diff-fa602a8a75dc9dcc92261bac5f533c2a85e34fcceaff63b3a3a81d9acde2fc52R67

pankajkoti: Can we test that all Google Cloud example DAGs (Especially BigQuery ones since the comment says that it was done for BigQuery) run fine after this change?

Lee-W: Yes, I just removed it. We've already tested all the google example DAGS without encountering errors

pankajkoti (approved):

Issue #1136, opened 2023-05-24
[GitHub Actions] reduce the wait time for providers-integration-tests-on-KE if providers-integration-tests is not deployed


Lee-W:
**Is your feature request related to a problem? Please describe.**
We currently wait 2 hours and 30 minutes for providers-integration-tests-on-KE even when providers-integration-tests are not deployed. The reason for the additional 2 hours is for waiting for providers-integration-tests.

**Describe the solution you'd like**
Wait providers-integration-tests-on-KE for 30 minutes if providers-integration-tests is not deployed

**Describe alternatives you've considered**

**Additional context**

PR #1137, opened 2023-05-24
Optimize deployment workflow

Lee-W:
* reduce the wait time to 30 minutes for providers-integration-tests-on-KE if providers-integration-tests is not deployed
* merge wait for deployment and trigger master dag job into the same job but different step

Closes: #1136

### DIFF
--- a/.github/workflows/deploy-integration-tests.yaml
+++ b/.github/workflows/deploy-integration-tests.yaml
@@ -86,12 +86,21 @@ jobs:
     needs: deploy-to-providers-integration-tests-on-KE
     runs-on: 'ubuntu-20.04'
     steps:
-      - name: Sleep for 2 hours 30 minutes
+      - name: Sleep and wait for the deployment to be ready
         run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-integration-tests.yaml
```diff
@@ -81,23 +76,25 @@ jobs:
...
+  wait-for-deployment-to-be-ready-and-trigger-master-dag-for-providers-integration-tests-on-KE:
     if: github.event_name == 'schedule'
     needs: deploy-to-providers-integration-tests-on-KE
     runs-on: 'ubuntu-20.04'
     steps:
-      - name: Sleep for 2 hours 30 minutes
+      - name: Sleep and wait for the deployment to be ready
         run: |
+          if [ "${{ inputs.environment_to_deploy }}" == "providers-integration-tests-on-KE" ]; then
+            # 30 minutes
```
pankajkoti, line 87:
```suggestion
            # This means that only KE deployment was triggered and hence we do not need to wait for CE deployment's master DAG to complete. We just need to wait for 30 minutes for the deployment to be reflected.
```

pankajkoti (approved):

Issue #1120, opened 2023-05-22
example_livy_operator failed with airflow issues 31322 RC providers


Lee-W:
**Describe the bug**
`cluster_creator` failed when testing `example_livy_operator` with the latest RC2 providers in https://github.com/apache/airflow/issues/31322.

**To Reproduce**
1. Update the provider version to the ones in https://github.com/apache/airflow/issues/31322

**Expected behavior**
DAG finishes without failure

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/site-packages/airflow/providers/amazon/aws/operators/emr.py", line 695, in execute
    log_uri=get_log_uri(emr_client=self._emr_hook.conn, job_flow_id=self._job_flow_id),
  File "/usr/local/lib/python3.10/site-packages/airflow/providers/amazon/aws/links/emr.py", line 61, in get_log_uri
    log_uri = S3Hook.parse_s3_url(response["Cluster"]["LogUri"])
KeyError: 'LogUri'
```

PR #1140, opened 2023-05-24
add loguri to EmrCreateJobFlowOperator for bypassing https://github.com/apache/airflow/issues/31480

Lee-W:
Note that this loguri is a fake s3 bucket which doesn't not exist and used only for bypassing this issue

Closes: #1120, #1121, #1123


pankajkoti:
I merged this for the time being to resolve the errors we see in our provider alerts. We can revert this later once we have the fix in the upstream provider.

Lee-W:
> I merged this for the time being to resolve the errors we see in our provider alerts. We can revert this later once we have the fix in the upstream provider.

Another thing we could do is add a valid s3 log uri if we want to test whether the log is sent, but it might need additional effort to clean that up as well

PR #1141, opened 2023-05-25
Integrate tests with runtime

vatsrahul1001:
1. Modifying master DAG Schedule to None when tested with runtime release
2. Updating comments in example_adf_run_pipeline.py DAG for connection usage


.circleci/integration-tests/master_dag.py
```diff
@@ -18,7 +18,8 @@
 SLACK_CHANNEL = os.getenv("SLACK_CHANNEL", "#provider-alert")
 SLACK_WEBHOOK_CONN = os.getenv("SLACK_WEBHOOK_CONN", "http_slack")
 SLACK_USERNAME = os.getenv("SLACK_USERNAME", "airflow_app")
-MASTER_DAG_SCHEDULE = os.getenv("MASTER_DAG_SCHEDULE", None)
+MASTER_DAG_SCHEDULE = os.getenv("MASTER_DAG_SCHEDULE", "0 0 * * *")
```
pankajkoti, line 21: A change was made recently in the GitHub actions deploy image job which triggers the master DAG after deploying. So, we do not need to have a scheduled run. Also, we do not want to run the master DAG at this schedule as @Lee-W starts his day during the run of this DAG and it creates a hurdle for him. So keeping set it to None is what we need at least for the alerts that we have in #provider-alert channel.

pankajkoti, line 21:
```suggestion
MASTER_DAG_SCHEDULE = os.getenv("MASTER_DAG_SCHEDULE", None)
```

.circleci/integration-tests/master_dag.py
```diff
@@ -88,6 +92,9 @@ def prepare_dag_dependency(task_info, execution_time):
     return _task_list, _dag_run_ids
 
 
+if IS_RUNTIME_RELEASE:
+    MASTER_DAG_SCHEDULE = None
+
```
pankajkoti, line 97:
```suggestion
```

pankajkoti: No need for this block as the schedule would be None anyway.

pankajkoti (commented):
Looks good to me, have minor comments inline. 

Thank you for the contribution here :)

.circleci/integration-tests/master_dag.py
```diff
@@ -29,9 +30,12 @@ def get_report(dag_run_ids: List[str], **context: Any) -> None:
 
         airflow_version = context["ti"].xcom_pull(task_ids="get_airflow_version")
         airflow_executor = context["ti"].xcom_pull(task_ids="get_airflow_executor")
-        airflow_version_message = (
-            f"The below run is on Airflow version `{airflow_version} with {airflow_executor} executor`\n\n"
-        )
+
+        if IS_RUNTIME_RELEASE:
+            airflow_version_message = f"Results generated for latest Runtime version {os.environ['ASTRONOMER_RUNTIME_VERSION']} with {airflow_executor}  \n\n"
```
pankajkoti, line 35:
```suggestion
            airflow_version_message = f"Results generated for Runtime version {os.environ['ASTRONOMER_RUNTIME_VERSION']} with {airflow_executor}  \n\n"
```
perhaps this may not be the latest runtime version always, no? Anyway good to skip the work latest since we're reflecting the version anyhow.

pankajkoti (approved):

.circleci/integration-tests/master_dag.py
vatsrahul1001: correct @pankajkoti

PR #1146, opened 2023-05-30
ci(github-action): periodically parse oss airflow issue to check provider release

Lee-W:
Check whether the RC testing announcement (e.g., https://github.com/apache/airflow/issues/31322) has been made on [apache-airflow](https://github.com/apache/airflow/). If there's an announcement and we do not have an existing open issue with the title `[DO NOT MERGE] Test RC provider packages for $rc_issue_url`, then trigger the rest of the RC provider testing workflow

### DIFF
--- a/.github/workflows/test-rc-release.yaml
+++ b/.github/workflows/test-rc-release.yaml
@@ -2,6 +2,8 @@
 name: test-rc-release
 
 on:  # yamllint disable-line rule:truthy
+  schedule:
+    - cron: "0 0 * * *"
   workflow_dispatch:
     inputs:
       rc_testing_branch:
@@ -25,8 +27,73 @@ on:  # yamllint disable-line rule:truthy
         default: 'main'
 
 jobs:
+  check-rc-testing-annocement:
+    runs-on: 'ubuntu-20.04'
+    if: github.event_name == 'schedule'
+    steps:
+      - name: Checkout apache-airflow
+        uses: actions/checkout@v3
+        with:
+          repository: 'apache/airflow'
+
+      - name: Parse the latest 100 GitHub issues from apache-airflow to check providers testing annocement
+        id: parse-airflow-gh-issues
+        run: |
+          # The default limit is 30. Set it to 100 for retrieving more issues
+          rc_issue_url=`gh issue list \
+            --json createdAt,title,url \
+            --limit 100 \
+            --jq 'map(
+                    select(
+                      .title |
+                      contains ("Status of testing Providers that were prepared on ")
+                    )
+                  ) | .[0].url'`
+
+          echo "rc_issue_url=$rc_issue_url" >> $GITHUB_OUTPUT
+
+      - name: Checkout current repo
+        uses: actions/checkout@v3
+        if: steps.parse-airflow-gh-issues.outputs.rc_issue_url != ''
+
+      - name: Parse the latest GitHub issues for checking existing RC provider testing issue
+        id: parse-current-repo
+        if: steps.parse-airflow-gh-issues.outputs.rc_issue_url != ''
+        run: |
+          # The default limit is 30. Set it to 100 for retrieving more issues
+          rc_issue_url="${{ steps.parse-airflow-gh-issues.outputs.rc_issue_url }}"
+          testing_issue_url=`gh issue list \
+            --json createdAt,title,url \
+            --limit 100 \
+            --jq 'map(
+                    select(
+                      .title |
+                      contains (
+                        "[DO NOT MERGE] Test RC provider packages for $rc_issue_url"
+                      )
+                    )
+                  ) | .[0].url'`
+
+          echo "testing_issue_url=$testing_issue_url" >> $GITHUB_OUTPUT
+
+      - name: Export rc_issue_url
+        id: export-rc-issue-url
+        run: |
+          rc_issue_url="${{ steps.parse-airflow-gh-issues.outputs.rc_issue_url }}"
+          testing_issue_url="${{ steps.parse-current-repo.outputs.testing_issue_url }}"
+          if [ "$rc_issue_url" == "" ] ; then
+            echo "No RC providers testing annocement found on apache-airflow"
+          elif [ "$testing_issue_url" != "" ] ; then
+            echo "Branch for testing RC providers has been created"
+            rc_issue_url=""
+          fi
+          echo "rc_issue_url=$rc_issue_url" >> $GITHUB_OUTPUT
+    outputs:
+      rc_issue_url: ${{ steps.export-rc-issue-url.outputs.rc_issue_url }}
+
   validate-manual-input:
     runs-on: 'ubuntu-20.04'
+    if: github.event_name == 'workflow_dispatch'
     steps:
       - name: Validate user input
         if: |
@@ -37,9 +104,14 @@ jobs:
           exit 1
 
   create-branch-for-testing-rc-release:
-    needs: validate-manual-input
+    needs: [validate-manual-input, check-rc-testing-annocement]
     runs-on: 'ubuntu-20.04'
-    if: ${{ inputs.issue_url }}
+    if: |
+      always() &&
+      (
+        (github.event_name == 'workflow_dispatch' && inputs.issue_url != '') ||
+        (github.event_name == 'schedule' && needs.check-rc-testing-annocement.outputs.rc_issue_url != '')
+      )
     steps:
       - name: Checkout
         uses: actions/checkout@v3
@@ -50,6 +122,22 @@ jobs:
         run: |
           python3 -m pip install -r dev/integration_test_scripts/requirements.txt
 
+      - name: Setup Github Actions git user
+        run: |
+          git config --global user.email "action@github.com"
+          git config --global user.name "GitHub Actions"
+
+      - name: Export GitHub RC provider testing url
+        id: export-rc-issue-url
+        run: |
+          if [ "${{ inputs.issue_url }}" != "" ] ; then
+            rc_issue_url="${{ inputs.issue_url }}"
+          else
+            rc_issue_url="${{ needs.check-rc-testing-annocement.outputs.rc_issue_url }}"
+          fi
+
+          echo "rc_issue_url=$rc_issue_url" >> $GITHUB_OUTPUT
+
       - name: Create RC branch
         id: create_rc_branch
         run: |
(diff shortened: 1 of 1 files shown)


Lee-W:
@utkarsharma2 this is the wip commit i mentioned earlier today

Lee-W:
@utkarsharma2 I just wrapped it up after working on issue-airflow. Please take a look if you're interested in it 🙂

.github/workflows/test-rc-release.yaml
```diff
@@ -25,8 +27,73 @@ on:  # yamllint disable-line rule:truthy
...
+        id: parse-airflow-gh-issues
+        run: |
+          # The default limit is 30. Set it to 100 for retrieving more issues
+          rc_issue_url=`gh issue list \
+            --json createdAt,title,url \
+            --limit 100 \
+            --jq 'map(
+                    select(
+                      .title |
+                      contains ("Status of testing Providers that were prepared on ")
```
utkarsharma2, line 49: @Lee-W Thanks for looking into this. 
I feel this is a bit risky since the value of `Status of testing Providers that were prepared on` might change from ticket to ticket and we will miss the testing of RC. WDYT?

utkarsharma2: I can think of a couple of solutions like sentiment analysis, but I think they are overkill for this situation.

Lee-W: Yes, such a check might be risky. I just took a look at the [airflow repo](https://github.com/apache/airflow/issues?q=is%3Aissue+%22Status+of+testing+Providers+that+were+prepared+on%22). There are around 50 issues in this format, so I'd say this is somewhat reliable. Even sentiment analysis might have some risk. Considering this, I think this might be a good enough solution, and we can always trigger manually if we notice this is not triggered.

uranusjr: The issue is actually generated from a script, so the title should be fairly predicatable:

https://github.com/apache/airflow/blob/8dc895cbb64e991893705b04e03a6655c35888fe/dev/breeze/src/airflow_breeze/commands/release_candidate_command.py#L248

If this still seems too unstructured, the *testing status* label can also be checked; I think the label is only used for this particular purpose right now.

.github/workflows/test-rc-release.yaml
```diff
@@ -25,8 +27,73 @@ on:  # yamllint disable-line rule:truthy
...
+  check-rc-testing-announcement:
+    runs-on: 'ubuntu-20.04'
+    if: github.event_name == 'schedule'
+    steps:
+      - name: Checkout apache-airflow
+        uses: actions/checkout@v3
+        with:
+          repository: 'apache/airflow'
+
+      - name: Parse the latest 100 GitHub issues from apache-airflow to check providers testing annocement
```
pankajkoti, line 39:
```suggestion
      - name: Parse the latest 100 GitHub issues from apache-airflow to check providers testing announcement
```

.github/workflows/test-rc-release.yaml
```diff
@@ -25,8 +27,73 @@ on:  # yamllint disable-line rule:truthy
...
+                    )
+                  ) | .[0].url'`
+
+          echo "rc_issue_url=$rc_issue_url" >> $GITHUB_OUTPUT
+
+      - name: Checkout current repo
+        uses: actions/checkout@v3
+        if: steps.parse-airflow-gh-issues.outputs.rc_issue_url != ''
+
+      - name: Parse the latest GitHub issues for checking existing RC provider testing issue
```
pankajkoti, line 59: Shouldn't we check for PRs instead of issues here? Also, where are we checking out the `astronomer-providers` repo for checking the existing PRs in our repo?

.github/workflows/test-rc-release.yaml
```diff
@@ -37,9 +104,14 @@ jobs:
...
   create-branch-for-testing-rc-release:
-    needs: validate-manual-input
+    needs: [validate-manual-input, check-rc-testing-annocement]
     runs-on: 'ubuntu-20.04'
-    if: ${{ inputs.issue_url }}
+    if: |
+      always() &&
+      (
+        (github.event_name == 'workflow_dispatch' && inputs.issue_url != '') ||
+        (github.event_name == 'schedule' && needs.check-rc-testing-annocement.outputs.rc_issue_url != '')
```
pankajkoti, line 113: Shouldn't we also add a check here that there is no existing PR in our repo?

pankajkoti (commented):
Looks great to me. Have some questions inline.

.github/workflows/test-rc-release.yaml
Lee-W: In the current design, I'd say checking issues might make more sense as we use the timestamp as part of the branch name. But I'm thinking of renaming it as `rc-test-for-issue-$issue_number`

.github/workflows/test-rc-release.yaml
Lee-W: if we have existing PR, we'll set this value to ''. https://github.com/astronomer/astronomer-providers/pull/1146/files#diff-7cd74bfb9babddfbfa7911a72e5d0784f450360d0a81e9f8319ecba30711b059R88

.github/workflows/test-rc-release.yaml
Lee-W: If we all agree on this change, I think I'll be able to work on it this week. Or we can merge this one and create a issue for this improvement

pankajkoti: maybe I missing something or not understanding. Are we planning to make a change with the automation that along with PRs we will also create issues in astronomer-providers for the RCs? If yes, how are we planning to create such issues?

Lee-W: Ah, it's me who misunderstood the workflow 🤦‍♂️ We actually create a PR with that title instead of the issue. Thanks for finding it out! I'll fix it late today

pankajkoti: okay, sorry for the late review on this one

Lee-W: Just fixed it. Could you pelase check again when you have time? Thanks!

.github/workflows/test-rc-release.yaml
```diff
@@ -25,8 +27,73 @@ on:  # yamllint disable-line rule:truthy
...
+                    )
+                  ) | .[0].url'`
+
+          echo "rc_issue_url=$rc_issue_url" >> $GITHUB_OUTPUT
+
+      - name: Checkout current repo
+        uses: actions/checkout@v3
+        if: steps.parse-airflow-gh-issues.outputs.rc_issue_url != ''
+
+      - name: Parse the latest GitHub issues for checking existing RC provider testing issue
```
pankajkoti, line 59:
```suggestion
      - name: Parse the latest GitHub pull requests for checking existing RC provider testing pull request
```

.github/workflows/test-rc-release.yaml
```diff
@@ -25,8 +27,73 @@ on:  # yamllint disable-line rule:truthy
...
+        id: parse-current-repo
+        if: steps.parse-airflow-gh-issues.outputs.rc_issue_url != ''
+        run: |
+          # The default limit is 30. Set it to 100 for retrieving more issues
+          rc_issue_url="${{ steps.parse-airflow-gh-issues.outputs.rc_issue_url }}"
+          testing_issue_url=`gh pr list \
```
pankajkoti, line 65:
```suggestion
          testing_pr_url=`gh pr list \
```

.github/workflows/test-rc-release.yaml
```diff
@@ -25,8 +27,73 @@ on:  # yamllint disable-line rule:truthy
...
+            --jq 'map(
+                    select(
+                      .title |
+                      contains (
+                        "[DO NOT MERGE] Test RC provider packages for $rc_issue_url"
+                      )
+                    )
+                  ) | .[0].url'`
+
+          echo "testing_issue_url=$testing_issue_url" >> $GITHUB_OUTPUT
```
pankajkoti, line 77:
```suggestion
          echo "testing_pr_url=$testing_issue_url" >> $GITHUB_OUTPUT
```

pankajkoti: and need to update the references in subsequent steps.

.github/workflows/test-rc-release.yaml
```diff
@@ -25,8 +27,73 @@ on:  # yamllint disable-line rule:truthy
...
+
+      - name: Export rc_issue_url
+        id: export-rc-issue-url
+        run: |
+          rc_issue_url="${{ steps.parse-airflow-gh-issues.outputs.rc_issue_url }}"
+          testing_issue_url="${{ steps.parse-current-repo.outputs.testing_issue_url }}"
+          if [ "$rc_issue_url" == "" ] ; then
+            echo "No RC providers testing annocement found on apache-airflow"
```
pankajkoti, line 85:
```suggestion
            echo "No RC providers testing announcement found on apache-airflow"
```

.github/workflows/test-rc-release.yaml
```diff
@@ -25,8 +27,73 @@ on:  # yamllint disable-line rule:truthy
         default: 'main'
 
 jobs:
+  check-rc-testing-annocement:
```
pankajkoti, line 30:
```suggestion
  check-rc-testing-announcement:
```

.github/workflows/test-rc-release.yaml
```diff
@@ -25,8 +27,73 @@ on:  # yamllint disable-line rule:truthy
...
+    runs-on: 'ubuntu-20.04'
+    if: github.event_name == 'schedule'
+    steps:
+      - name: Checkout apache-airflow
+        uses: actions/checkout@v3
+        with:
+          repository: 'apache/airflow'
+
+      - name: Parse the latest 100 GitHub issues from apache-airflow to check providers testing annocement
```
pankajkoti, line 39:
```suggestion
      - name: Parse the latest 100 GitHub issues from apache-airflow to check providers testing announcement
```

.github/workflows/test-rc-release.yaml
```diff
@@ -37,9 +104,14 @@ jobs:
           exit 1
 
   create-branch-for-testing-rc-release:
-    needs: validate-manual-input
+    needs: [validate-manual-input, check-rc-testing-annocement]
```
pankajkoti, line 107: This needs to be an `or` condition right? validate-manual-input runs only on `workflow_dispatch` whereas `check-rc-testing-annocement` runs only on `schedule`, so both will not be true at the same time and this job will never be executed then, no?

.github/workflows/test-rc-release.yaml
```diff
@@ -25,8 +27,73 @@ on:  # yamllint disable-line rule:truthy
...
+
+      - name: Checkout current repo
+        uses: actions/checkout@v3
+        if: steps.parse-airflow-gh-issues.outputs.rc_issue_url != ''
+
+      - name: Parse the latest GitHub issues for checking existing RC provider testing issue
+        id: parse-current-repo
+        if: steps.parse-airflow-gh-issues.outputs.rc_issue_url != ''
+        run: |
+          # The default limit is 30. Set it to 100 for retrieving more issues
```
pankajkoti, line 63:
```suggestion
          # The default limit is 30. Set it to 100 for retrieving more pull requests
```

.github/workflows/test-rc-release.yaml
pankajkoti, line 65: I think this only lists open PRs. Once will close the PRs, and since the issue could still be there in apache-airflow in top 100 issues, are we at risk of recreating PRs? I think we should list here merged PRs too

pankajkoti (commented):
Will it be possible to test this workflow somehow?

.github/workflows/test-rc-release.yaml
Lee-W: AFAIK, we don't have a way to do it in `needs`. That's why I use `always()` [here](https://github.com/astronomer/astronomer-providers/pull/1146/files#diff-7cd74bfb9babddfbfa7911a72e5d0784f450360d0a81e9f8319ecba30711b059R110)

.github/workflows/test-rc-release.yaml
Lee-W: Yes, that makes sense. I waited for the airflow issue to close before I closes the PR in the previous releases. Let me add it 💪 Thanks!

Lee-W: just added

Lee-W:
> Will it be possible to test this workflow somehow?

Yes, we can simulate through the latest closed issue (we will need to change the code to read closed issue from airflow for testing).

.github/workflows/test-rc-release.yaml
pankajkoti: okay, so just for my understanding, we are guaranteed that this job will run, right? Although `validate-manual-input` or `check-rc-testing-annocement` is skipped based on their if conditions, since we have always, this job will be run, correct? If that is correct, yes, please mark this conversation as resolved. Sorry for my lack of expertise here :)

pankajkoti (approved):
Looks good to me. IMO, we can merge this if we want the code to be available in main to test something. We can always fix something later if something is not working well.

.github/workflows/test-rc-release.yaml
Lee-W: Yes, that's how it works. Thanks for looking so deep into it 🙌 It also took me some time to sort all these stuff out

Lee-W:
@pankajkoti We don't actually need it to be in the main branch for testing, but I think it is better to merge it to the main branch earlier. I'll do some simple tests on another branch to verify the basic functionality again (as we have made some changes these days). Once those work as I expect, I'll merge this one. Thanks for reviewing 👍

PR #1151, opened 2023-05-31
Terminate running virtual clusters launched for EMR containers

pankajkoti:
It has been observed that sometimes the virtual clusters launched 
by EMR containers are not cleaned up. It is not possible to nuke 
them using AWS nuke as this resource type is not exposed, 
additionally, it has not been exposed to delete them in the 
AWS console too. The only way to delete them is using AWS CLI 
and hence we're adding a task in the nuke DAG to delete those 
using AWS CLI.

### DIFF
--- a/astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
@@ -29,6 +29,15 @@ with DAG(
 ) as dag:
     start = DummyOperator(task_id="start")
 
+    terminate_running_emr_virtual_clusters = BashOperator(
+        task_id="terminate_running_emr_virtual_clusters",
+        bash_command=f"set -e; "
+        f"aws configure set aws_access_key_id {AWS_ACCESS_KEY_ID}; "
+        f"aws configure set aws_secret_access_key {AWS_SECRET_ACCESS_KEY}; "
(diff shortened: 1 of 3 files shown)


astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
```diff
@@ -29,6 +29,15 @@
 ) as dag:
     start = DummyOperator(task_id="start")
 
+    terminate_running_emr_virtual_clusters = BashOperator(
+        task_id="terminate_running_emr_virtual_clusters",
+        bash_command=f"set -e; "
+        f"aws configure set aws_access_key_id {AWS_ACCESS_KEY_ID}; "
```
Lee-W, line 35: Do we need to install aws-cli?

Lee-W: If we already have it installed, I'm good with this PR. Thanks for taking care of this issue!

Lee-W (approved):

pankajastro (approved):

astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
pankajkoti: yes, it's installed already with the images we build using the Dockerfiles. e.g. Dockerfile.aws_cloud in the integration_tests or dev/Dockerfile.aws

Issue #1131, opened 2023-05-23
Delete stale VPCs, subnets, route tables created from EMR DAG


pankajkoti:
Currently, the `emr_eks_pi_job` DAG fails intermittently due to leftover VPCs created from earlier runs.
We need to identify a reliable automation (most probably in the same example DAG) to delete these while being extremely careful to not delete the 2 Databricks-related VPCs present in the account.

phanikumv:
we can identify the VPC with the string **emr_eks_ctl**

pankajkoti:
okay, will check this sometime soon.

PR #1153, opened 2023-05-31
Delete leftover/stale VPCs launched by EMR-EKS

pankajkoti:
Delete leftover/stale VPCs launched by EMR-EKS

The PR adds a script to do the following steps for deleting the VPCs:
1. List the VPCs with the prefix `eksctl-providers-team-eks`
2. Firstly delete all the association of the VPCs, viz. subnets,
   Internet gateways, route tables, security groups, and network
   interfaces
3. By default, AWS takes care of the main route table and the default
   security groups are not deleted as they have further dependencies
   associated with them, which we do not touch in our script.
   Hence we're safe here.
4. Finally, delete the VPCs and report the exit status of these final
   deletion commands as the task status. We do not want to `set -e`
   in the beginning of the script as the association failures can be
   ignored if we're able to delete the VPCs.

The PR then adds this script to be called a BashOperator task
in the AWS nuke DAG.

closes: #1131

### DIFF
--- a/astronomer/providers/amazon/aws/example_dags/delete_stale_emr_vpcs.sh
+++ b/astronomer/providers/amazon/aws/example_dags/delete_stale_emr_vpcs.sh
@@ -1,5 +1,24 @@
 #!/bin/bash
 
+# What does the script do?
+# This script automates the deletion of resources associated with VPCs (Virtual Private Clouds) in AWS .
+# It takes a prefix as input and retrieves a list of VPC IDs with the specified prefix. Then, it iterates through each
+# VPC ID and performs the following actions in the same order listed below to delete the VPC successfully.
+#     1. Deletes associated subnets
+#     2. First detach and then delete associated internet gateways
+#     3. Disassociates subnet associations and the main route table if applicable for the route tables and then delete
+#        the associated route tables
+#     4. Deletes associated security groups
+#     5. Deletes associated network interfaces
+#     6. Finally, deletes the VPC itself
+#
+# Why do we need this script?
+# When we create an EKS cluster using eksctl in the "emr_eks_pi_job" DAG , it creates a VPC with the specified prefix.
+# When we delete the cluster as part of its cleanup script task, ideally the delete cluster command should delete the
+# VPC as well. However, there were observed occurrences where the VPCs were not deleted. This script automates the
+# deletion of such stale VPCs and associated resources. We call this script as a task from the AWS nuke DAG at the end
+# of the master DAG run.
+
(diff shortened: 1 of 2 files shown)


pankajkoti:
Verified that the VPC was successfully deleted in the AWS console.

Screenshot 2023-05-31 at 7 38 20 PM

astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
```diff
@@ -46,6 +46,13 @@
...
+    delete_stale_emr_vpcs = BashOperator(
+        task_id="delete_stale_emr_vpcs",
+        bash_command="sh $AIRFLOW_HOME/dags/delete_stale_emr_vpcs.sh ",
+        trigger_rule="all_done",
+    )
+
     end = DummyOperator(task_id="end")
 
-    start >> terminate_running_emr_virtual_clusters >> execute_aws_nuke >> end
+    start >> terminate_running_emr_virtual_clusters >> execute_aws_nuke >> delete_stale_emr_vpcs
```
Lee-W, line 57: May I know why we can't run `delete_stale_emr_vpcs` and `execute_aws_nuke` in parallel? Thanks

Lee-W (commented):
Overall, I'm good with this PR. But I would like to wait for the testing mentioned in Slack to succeed before we merge it. Another thing I'm curious about is whether it makes more sense for us to implement `delete_stale_emr_vpcs.sh` as a Python script/function. We can use the subprocess module and don't need to install `jq`.

astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
pankajkoti: Ideally, aws_nuke should take care of deleting the VPCs too. But, since we observed that sometimes they remain stale, I am running them after the aws nuke. I do not want to delete the VPCs before running AWS nuke because it may affect the nuke if VPCs are deleted before and then the CloudFormation tries deleting them as part of its stack which may fail.

astronomer/providers/amazon/aws/example_dags/delete_stale_emr_vpcs.sh
```diff
@@ -0,0 +1,83 @@
+#!/bin/bash
+
```
phanikumv, line 2: Please give some context of what this script is intended to do at the beginning of the script

Lee-W (approved):

phanikumv (approved):

PR #1156, opened 2023-06-01
Changes for running master dag in parallel

vatsrahul1001:
Added Env vars in a few DAGS which had hardcoded values


pankajkoti:
Thanks a lot for this @vatsrahul1001 . One thing I was wondering where we should mention/note which all environment variables need to have unique values across all the example DAGs for the parallel runs to succeed. Would adding an inline comment for those environment variables in the example DAGs make sense?

This would help us when in future we want to create new deployments and need to know which environment variables need to be set with unique values.

vatsrahul1001:
> Thanks a lot for this @vatsrahul1001 . One thing I was wondering where we should mention/note which all environment variables need to have unique values across all the example DAGs for the parallel runs to succeed. Would adding an inline comment for those environment variables in the example DAGs make sense?
> 
> This would help us when in future we want to create new deployments and need to know which environment variables need to be set with unique values.

@pankajkoti I think we should have one place where we have all the environment vars needed, Putting note in every example will make it difficult for anyone to check later on

pankajkoti:
> > Thanks a lot for this @vatsrahul1001 . One thing I was wondering where we should mention/note which all environment variables need to have unique values across all the example DAGs for the parallel runs to succeed. Would adding an inline comment for those environment variables in the example DAGs make sense?
> > This would help us when in future we want to create new deployments and need to know which environment variables need to be set with unique values.
> 
> @pankajkoti I think we should have one place where we have all the environment vars needed, Putting note in every example will make it difficult for anyone to check later on

makes sense. where can we have such a note? I have a few more inputs on it with respect to the EMR EKS DAG. Wanted to check if they are already covered or we need to add those too in the note.

vatsrahul1001:
> > > Thanks a lot for this @vatsrahul1001 . One thing I was wondering where we should mention/note which all environment variables need to have unique values across all the example DAGs for the parallel runs to succeed. Would adding an inline comment for those environment variables in the example DAGs make sense?
> > > This would help us when in future we want to create new deployments and need to know which environment variables need to be set with unique values.
> > 
> > 
> > @pankajkoti I think we should have one place where we have all the environment vars needed, Putting note in every example will make it difficult for anyone to check later on
> 
> makes sense. where can we have such a note? I have a few more inputs on it with respect to the EMR EKS DAG. Wanted to check if they are already covered or we need to add those too in the note.

1. I was thinking of adding comment in Master DAG for the same
2. for EMR i have set (EKS_CLUSTER_NAME,EKS_NAMESPACE,EMR_VIRTUAL_CLUSTER_NAME,MANAGE_VIRTUAL_CLUSTERS) env as unique

pankajkoti:
Perfect place I think. Can we ade that in this PR. In the EMR EKS the iam role we create also needs to be unique. Also mentioning what is the suffix we add for the deployment would help.

vatsrahul1001:
> Perfect place I think. Can we ade that in this PR. In the EMR EKS the iam role we create also needs to be unique. Also mentioning what is the suffix we add for the deployment would help.

@pankajkoti 
1. Sure will add in master dag
2. I think we can make JOB_EXECUTION_ROLE unique right?
3. Right now I am adding suffix as per the deployment executor type i.e celery or k8s

pankajkoti:
> > Perfect place I think. Can we ade that in this PR. In the EMR EKS the iam role we create also needs to be unique. Also mentioning what is the suffix we add for the deployment would help.
> 
> @pankajkoti
> 
> 1. Sure will add in master dag
> 2. I think we can make JOB_EXECUTION_ROLE unique right?
> 3. Right now I am adding suffix as per the deployment executor type i.e celery or k8s

yes to 1 & 2
For 3, let's add the suffix placeholder e.g ProvidersTeamJobExecutionRole<executor> in the master DAG where you'd add a comment.

.circleci/integration-tests/master_dag.py
```diff
@@ -21,6 +21,17 @@
 MASTER_DAG_SCHEDULE = os.getenv("MASTER_DAG_SCHEDULE", None)
 IS_RUNTIME_RELEASE = bool(os.getenv("IS_RUNTIME_RELEASE", False))
 
+"""
+To run this master dag across multiple deployment parallelly, please set these env var unique per deployment
+
+GCP_BIGQUERY_DATASET_NAME,BATCH_JOB_COMPUTE_ENV, MODEL_NAME,AZURE_DATA_STORAGE_BLOB_NAME,AZURE_DATA_STORAGE_CONTAINER_NAME,
```
Lee-W, line 27: I thought about whether we could add a single env var (e.g., `MASTER_DAG_RUN_SUFFIX`) that can work as a suffix to these variables so that we don't need to change all of it.

maybe something like the following?

```python
import os

MASTER_DAG_RUN_SUFFIX = os.getenv("MASTER_DAG_RUN_SUFFIX", None)

env_keys = [
    "GCP_BIGQUERY_DATASET_NAME",
    "BATCH_JOB_COMPUTE_ENV",
    "MODEL_NAME",
    "AZURE_DATA_STORAGE_BLOB_NAME",
    "AZURE_DATA_STORAGE_CONTAINER_NAME",
    "GCP_TEST_BUCKET",
    "GCP_DATAPROC_CLUSTER_NAME",
    "GKE_CLUSTER_NAME",
    "BATCH_JOB_NAME",
    "BATCH_JOB_QUEUE",
    "REDSHIFT_TABLE_NAME",
    "GKE_POD_NAME",
    "EMR_VIRTUAL_CLUSTER_NAME",
    "REDSHIFT_CLUSTER_IDENTIFIER",
    "EKS_CLUSTER_NAME",
    "EKS_NAMESPACE",
    "MANAGE_VIRTUAL_CLUSTERS",
    "JOB_EXECUTION_ROLE",
]

if MASTER_DAG_RUN_SUFFIX:
    for env_key in env_keys:
        value = os.getenv(env_key, None)
        if value:
            os.environ[env_key] += MASTER_DAG_RUN_SUFFIX

```

Lee-W: After a second thought, this might not be sufficient. We'll need to pass these values to the triggered dag as well. Also it might be handy if the default value of `MASTER_DAG_RUN_SUFFIX` is some random value (maybe a timestamp?) so that all run is parallel by default

pankajkoti: Cloud resources usually have constraints on naming them, so we need to ensure that the timestamp format we wish to specify is accepted by all of them(usually the accept alphanumeric chars, but we observed some resources do not accept upper case characters 😄 ); otherwise resource creation would fail

astronomer/providers/apache/hive/example_dags/example_hive.py
```diff
@@ -37,6 +37,7 @@
 HIVE_SCHEMA = os.getenv("HIVE_SCHEMA", "default")
 HIVE_TABLE = os.getenv("HIVE_TABLE", "zipcode")
 HIVE_PARTITION = os.getenv("HIVE_PARTITION", "state='FL'")
+HIVE_CLUSTER = os.getenv("HIVE_CLUSTER", "example_hive_sensor_cluster")
```
pankajkoti, line 40: Let's also do this for the LIVY DAG. We may not have encountered the parallelism yet but it likely is a candidate. And also let's list that at the top for unique environment values.

.circleci/integration-tests/master_dag.py
pankajkoti, line 27: It's a bit difficult to read these. Can we have 1 variable per line and also possibly group them by DAGs?

Lee-W: Yep, we can encode the timestamp into something acceptable by those resources

astronomer/providers/amazon/aws/example_dags/example_sagemaker.py
```diff
@@ -248,7 +249,12 @@ def delete_logs(task_instance: TaskInstance) -> None:
             if client.describe_log_streams(logGroupName=group)["logStreams"]:
                 client.delete_log_group(logGroupName=group)
         except ClientError as e:
-            raise e
+            if e.response["Error"]["Code"] == "ResourceNotFoundException":
+                # Handle the resource group not found exception
+                print("logs group doesn't exits")
```
pankajkoti, line 254:
```suggestion
                print("logs group doesn't exist")
```

pankajkoti (approved):

astronomer/providers/amazon/aws/example_dags/example_sagemaker.py
```diff
@@ -248,7 +249,12 @@ def delete_logs(task_instance: TaskInstance) -> None:
             if client.describe_log_streams(logGroupName=group)["logStreams"]:
                 client.delete_log_group(logGroupName=group)
         except ClientError as e:
-            raise e
+            if e.response["Error"]["Code"] == "ResourceNotFoundException":
+                # Handle the resource group not found exception
+                print("logs group doesn't exist")
```
Lee-W, line 254: nitpick

```suggestion
                logging.warning("logs group doesn't exist")
```

Lee-W (approved):

.circleci/integration-tests/master_dag.py
Lee-W: Should we make it an issue if we're not going to implement this in this PR?

PR #1157, opened 2023-06-01
Refactor EMR EKS cleanup script to handle failures gracefully

pankajkoti:
This commit refactors the cleanup script that automates 
the deletion of resources launched by the EMR_EKS_PI_JOB DAG. 
The script deletes an IAM role, detaches policies from the role, 
removes the role from associated instance profiles, deletes a 
virtual cluster (EMR Containers), and deletes an EKS cluster and 
its associated node group. The major change with the refactoring 
is to include checks to avoid failures if resources don't exist because 
the creation script for creating the resources might have created 
partial resources and could have failed without fully completing the 
creation.


pankajkoti:
Screenshot 2023-06-02 at 1 19 39 AM

The DAG ran fine and I verified that there are no stale resources left in the AWS account. Checked that the CloufFormation stack, IAM role, VPCs, EMR Container virtual clusters are all successfully cleaned up.

Lee-W (approved):

PR #1158, opened 2023-06-02
Remove AWS nuke DAG from master DAG & accept DAG schedule param 

pankajkoti:
It is observed that when 2 deployments are run in parallel, 
their master DAGs trigger the nuke task which cleanup the 
resources and if one master DAG finished running before 
the other one across the deployments, the other deployment's
 master DAG tasks will fail as the resources will no longer be 
available. To mitigate this, we remove the nuke DAG trigger 
from the master DAG and instead plan to schedule the master 
DAG only from one of the provider integration tests 
deployment (e.g. CeleryExecutor) by using the env var 
"AWS_NUKE_DAG_SCHEDULE". We also want to be alerted in
 our Slack channel if one of the nuke tasks fail and hence also 
add a report task for the DAG.

### DIFF
--- a/.circleci/integration-tests/master_dag.py
+++ b/.circleci/integration-tests/master_dag.py
@@ -238,18 +238,6 @@ with DAG(
     dag_run_ids.extend(ids)
     chain(*dbt_trigger_tasks)
 
-    end_test = DummyOperator(
-        task_id="end_test",
-        trigger_rule="all_success",
-    )
-
-    aws_nuke_task_info = [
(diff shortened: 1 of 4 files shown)


pankajkoti:
Tested this E2E and we received alerts in our Slack channel.

Screenshot 2023-06-02 at 10 01 49 PM

pankajastro (approved):

pankajkoti:
I am merging this for now to test parallel runs. If there are comments, I will address those later in a subsequent PR.

PR #1159, opened 2023-06-02
Scan quay update runtime image

pankajkoti:


phanikumv (approved):
Nice work @pankajkoti

pankajkoti:
> Nice work @pankajkoti 

Thank you Phani for the idea and support!

.github/workflows/runtime-image-update.yaml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Install dependencies
+        run: pip install requests semantic-version
+
+      - name: Get current tag from Dockerfile.astro_cloud
+        id: get_current_tag
+        # yamllint disable rule:line-length
+        run: |
+          CURRENT_TAG=$(awk -F: '/quay.io\/astronomer\/astro-runtime/ {print $NF}' ./.circleci/integration-tests/Dockerfile.astro_cloud)
+          echo "CURRENT_TAG=$CURRENT_TAG"
+          echo "::set-output name=tag::$CURRENT_TAG"
```
Lee-W, line 35: This seems to be deprecating.
https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

```suggestion
          echo "tag=$CURRNET_TAG" >> $GITHUB_OUTPUT
```

Lee-W: same applied to elsewhere

.github/workflows/runtime-image-update.yaml
```diff
@@ -0,0 +1,105 @@
...
+            echo "Tag was not updated"
+            echo "::set-output name=tag_updated::false"
+          fi
+
+      - name: Create runtime image tag update branch
+        id: create_runtime_image_update_branch
+        run: |
+          BRANCH_NAME="update-runtime-image-tag-${{ steps.get_latest_tag.outputs.tag }}"
+          git fetch origin main
+          git checkout main
```
Lee-W, line 63: I'm not sure whether it's necessary. I thought we've checkout to main through https://github.com/astronomer/astronomer-providers/pull/1159/files#diff-09763d6cc66b999842e6f6520c49b659d5d8e89185fb9b64704e07c69f473e42R18-R19

Lee-W (commented):
It's great! 🎉 Even though it has been merged, I am still interested in learning about other automation. I made a few minor comments, but nothing significant.

.github/workflows/runtime-image-update.yaml
```diff
@@ -0,0 +1,105 @@
...
+      - name: Commit changes and create a pull request
+        env:
+          GH_TOKEN: ${{ github.token }}
+        # yamllint disable rule:line-length
+        run: |
+          set -e
+          git add Makefile .circleci/integration-tests/Dockerfile.astro_cloud
+          git commit -m "Update runtime image tag to ${{ steps.get_latest_tag.outputs.tag }}"
+          git push origin ${{ steps.create_runtime_image_update_branch.outputs.runtime_image_update_branch }}
+          gh pr create --base main \
```
Lee-W, line 102: Should we check whether the pr has been created? Otherwise, wouldn't it generate a PR per day if we do not update the image in time? If this is needed in this automation as well, I have implemented similar behavior [here](https://github.com/astronomer/astronomer-providers/pull/1146/files#diff-7cd74bfb9babddfbfa7911a72e5d0784f450360d0a81e9f8319ecba30711b059R64).

.github/workflows/runtime-image-update.yaml
pankajkoti: Thank you for this. Created a PR to address this https://github.com/astronomer/astronomer-providers/pull/1163

.github/workflows/runtime-image-update.yaml
pankajkoti: The Github actions job was failing in my PR without this. Only after introducing this the Actions job finally ran fine.

.github/workflows/runtime-image-update.yaml
pankajkoti: Thanks for this. I was missing another condition for these steps to check whether the tag has been updated, added that now so that if at least the main contains the latest tag, we do not proceed.

Coming back to your question, if we forget to update the image in time (which should not happen ideally) since the branch name would be the same, this step would fail when it tries to push the changes saying that upstream has changes that need to be pulled in first from the same branch. So we're safe that way. I agree, yours is the cleaner way but let's see if I find time later, will make that change too.
e.g. an action run when it failed for the scenario we're discussing https://github.com/astronomer/astronomer-providers/actions/runs/5155815512/jobs/9285972413

PR #1163, opened 2023-06-03
Address @Lee-W's comments on PR #1159

pankajkoti:
The PR address review comments on PR #1159 wrt 
below points:
1. Remove usage of deprecated way to write step outputs.
2. Add a condition to check tags are updated for running the 
step to create a PR.


Lee-W (approved):
Thank you for addressing these comments so quickly!

PR #1168, opened 2023-06-05
Increase wait timeout for EKS cluster deletion to 60 minutes

pankajkoti:
Previously, we were just triggering the deletion of the EKS cluster 
without waiting for it to finish, as a result, the task did not reflect 
the accurate status of whether the EKS cluster got deleted. 
However, in PR #1157, we introduce the `--wait`` flag on the 
delete command so that the Airflow task reflects the correct status. 
However, it has been observed that the deletion takes longer than 
the default 25m0s timeout set by the command. Hence, we're 
increasing the timeout to 60m0s. I think this is a safe number given 
that we also run the Nuke DAG in sometime after the master DAG run 
which should clean up leftover resources.
Additionally, we want to continue the deletion of the cluster in case 
of failures as there are no dependencies on the cluster and hence
 we're also introducing the `--force` flag while deletion. 
I believe with the addition of these 2 flags, the task should not fail again.


astronomer/providers/amazon/aws/example_dags/example_delete_eks_cluster_and_role_policies.sh
```diff
@@ -77,5 +77,5 @@ fi
 
 # Command to delete the EKS cluster and node group attached with it. Make sure to wait for the cluster to be deleted so
 # that the task reports for any potential failures.
-eksctl delete cluster $EKS_CLUSTER_NAME --wait
+eksctl delete cluster $EKS_CLUSTER_NAME --wait --force --timeout 60m0s
```
phanikumv, line 80:
```suggestion
eksctl delete cluster $EKS_CLUSTER_NAME --wait --force --timeout 600ms
```

pankajkoti: The default value is 25 minutes 0 seconds. The syntax is the form X minutes Y seconds -> XmYs

pankajkoti: Screenshot 2023-06-05 at 5 50 18 PM

phanikumv (approved):

Issue #757, opened 2022-11-15
SagemakerProcessingOperatorAsync to honor existing_jobs_found


rajaths010494:
**Is your feature request related to a problem? Please describe.**
Currently, we are failing if the job already exists in SageMaker

**Describe the solution you'd like**
https://github.com/apache/airflow/pull/27456. The same has been fixed in OSS airflow

pankajkoti:
We're waiting for the SageMaker OSS Airflow PRs with the deferrable parameters to be merged and available in the upcoming release. Post that we will inherit the OSS operator here so that we do not have to do double work.

pankajkoti:
PR #1169 was created and is waiting for reviews.

pankajkoti:
Reverting PR #1169 into PR https://github.com/astronomer/astronomer-providers/pull/1175 as we need a major bump of astronomer-providers as per discussion with the team and tagging this issue to milestone `2.0.0`

pankajastro:
@phanikumv do we still want to add this feature? asking since we have already donated this to OSS and the OSS operator take care of this.

phanikumv:
Closing as the functionality as available in OSS airflow

PR #1169, opened 2023-06-05
Extend upstream Amazon provider SageMaker operators; set deferrable to True

pankajkoti:
The PR aims at extending the upstream SageMaker operators of the 
Amazon providers of Apache Airflow for the asynchronous SageMaker
operators in ``astronomer-providers``, setting the `deferrable` param
to `True` while initialising the corresponding parent constructor and 
cleaning up the code for the existing implementation.

With this, the aim is to push support and fixes to the upstream operators 
as much as possible, and also we do not have to keep track of the 
upstream changes and backport them to the operators in this module.

Following is the high-level summary of the changes made in the PR:
1. Modify the `astronomer/providers/amazon/aws/operators/sagemaker.py`
module to incorporate an extension of corresponding upstream operators in the 
asynchronous operators and set `defferable=True` while initialising the parent
class constructor. Since the parent class supports deferrable capabilities well
now, we do not need any additional code for the operators here; so we clean 
that up here
2. Remove the following modules as they are no longer needed. They are 
maintained in the upstream provider.
i.  `astronomer/providers/amazon/aws/hooks/sagemaker.py`
ii. `astronomer/providers/amazon/aws/triggers/sagemaker.py`
4. Regarding the unit tests, since we only extend the upstream provider and 
set `deferrable=True`, we only assert that the deferrable param is set in the 
Async operator by initialising the operator. All the other tests are removed
as the code no longer resides here and they would be covered in the upstream
provider. Hence, we're also removing the test modules for the hook and the 
trigger.
5. Since the deferrable capability became available for these operators starting 
`apache-airflow-providers-amazon 8.1.0`, we set the minimum version for the 
provider in `setup.cfg` to `8.1.0`

closes: #757


pankajkoti:
The code-coverage report says that the patch coverage is `100%` as all the code with the refactored implementation has been covered. 
However, the project coverage seems to be dropping by `-0.08%` and it is because we're cleaning up `>5k` lines, and hence the relative coverage for the project has been affected.

astronomer/providers/amazon/aws/operators/sagemaker.py
```diff
@@ -1,381 +1,66 @@
...
-)
-from astronomer.providers.utils.typing_compat import Context
-
-
-def serialize(result: dict[str, Any]) -> str:
-    """Serialize any objects coming from Sagemaker API response to json string"""
-    return json.loads(json.dumps(result, cls=AirflowJsonEncoder))  # type: ignore[no-any-return]
 
 
 class SageMakerProcessingOperatorAsync(SageMakerProcessingOperator):
```
Lee-W, line 21: I think we might still need the following for backward compatibility. @phanikumv Could you please confirm? Thanks!

```python
class SageMakerProcessingOperatorAsync(SageMakerProcessingOperator):
        def __init__(self, **kwargs: Any) -> None:
            return super().__init__(deferrable=True, **kwargs)
```

pankajkoti: I think the idea to extend the upstream providers was to help us not keep us tracking changes in upstream changes and it's backward compatible functionality wise, already no? Our methods din't have different signatures.

Lee-W: Yep, but I think this code is for users still using `from astronomer... import SageMakerProcessingOperatorAsync`.

pankajkoti: Which part exactly you mean? They can still keep doing it. The class is still there.

pankajkoti: The init method is still there, it accepts both *args and *kwargs for all the operators.

https://github.com/astronomer/astronomer-providers/pull/1169/files#diff-f57678957312458c2669bdee1f0ee1c8a4afe88ea8bf99fef01f69c555b44cddR24

https://github.com/astronomer/astronomer-providers/pull/1169/files#diff-f57678957312458c2669bdee1f0ee1c8a4afe88ea8bf99fef01f69c555b44cddR34

pankajkoti: Since the diff is huge, maybe it is unclear, 
but this is how the operator module looks like if you checkout to the branch in this PR locally
```
"""
This module contains operators to interact with Amazon SageMaker in an asynchronous way.

Starting release ``1.17.0`` of the ``astronomer-providers`` package, the operators in this module extend the
corresponding operators in the Amazon provider of Apache Airflow supporting the deferrable capabilities. With this,
the aim is to push support and fixes to the upstream operators as much as possible, and also we do not have to keep
track of the  upstream changes and backport them to the operators in this module.
"""

from __future__ import annotations

from typing import Any

from airflow.providers.amazon.aws.operators.sagemaker import (
    SageMakerProcessingOperator,
    SageMakerTrainingOperator,
    SageMakerTransformOperator,
)


class SageMakerProcessingOperatorAsync(SageMakerProcessingOperator):
    """The operator is used to start a SageMaker processing job asynchronously."""

    def __init__(self, *args: Any, **kwargs: Any) -> None:
        """
        Initialize a new SageMakerProcessingOperatorAsync instance.

        We set the deferrable flag to True while calling the super class constructor to enable the deferrable
        capabilities.

        :param args: List of arguments to pass to SageMakerProcessingOperator.
        :param kwargs: Keyword arguments to pass to SageMakerProcessingOperator.
        """
        super().__init__(*args, deferrable=True, **kwargs)


class SageMakerTransformOperatorAsync(SageMakerTransformOperator):
    """The operator is used to start a SageMaker transform job asynchronously."""

    def __init__(self, *args: Any, **kwargs: Any) -> None:
        """
        Initialize a new SageMakerTransformOperatorAsync instance.

        We set the deferrable flag to True while calling the super class constructor to enable the deferrable
        capabilities.

        :param args: List of arguments to pass to SageMakerTransformOperator.
        :param kwargs: Keyword arguments to pass to SageMakerTransformOperator.
        """
        super().__init__(*args, deferrable=True, **kwargs)


class SageMakerTrainingOperatorAsync(SageMakerTrainingOperator):
    """The operator is used to start a SageMaker training job asynchronously."""

    def __init__(self, *args: Any, **kwargs: Any) -> None:
        """
        Initialize a new SageMakerTrainingOperatorAsync instance.

        We set the deferrable flag to True while calling the super class constructor to enable the deferrable
        capabilities.

        :param args: List of arguments to pass to SageMakerTrainingOperator.
        :param kwargs: Keyword arguments to pass to SageMakerTrainingOperator.
        """
        super().__init__(*args, deferrable=True, **kwargs)
```

Lee-W: Ah got it. I thought these files are removed 🤦‍♂️ Thanks for your clarification!

Lee-W (approved):

astronomer/providers/amazon/aws/operators/sagemaker.py
pankajkoti: No, no. The operator module is still there and the init methods should support existing implementation. The GitHub review tab does not show well for large diff when so many lines are getting dropped.

pankajastro:
I think we have some [docs here](https://www.notion.so/astronomerio/Draft-Reduce-maintenance-effort-and-keep-backward-Compatibility-for-astronomer-providers-after-con-e16857c5ea814c71afe5009c674d0d67) for the same so I'm not sure if we can release it in minor release

Issue #1166, opened 2023-06-05
check_fn not being used correctly in S3KeySensorAsync


sunank200:
**Describe the bug**
When user passes `check_fn` which is a function that receives the list of the S3 objects, and returns a boolean. Currently the `check_fn` is not being used in the code at all: https://github.com/astronomer/astronomer-providers/blob/main/astronomer/providers/amazon/aws/triggers/s3.py#L70-L76

**Expected behavior**

When user passes `check_fn`, it should use this python callable in sensor.
```
        if self.check_fn is not None:
            return self.check_fn(files)
```

pankajastro:
unfortunately, the trigger can't serialise callable so we can't handle it in the trigger but if we want to implement it we can do it in `execute_complete` and then again defer

pankajastro:
I have PR here https://github.com/astronomer/astronomer-providers/pull/1171 here PTAL if you feel positive then I'll add test etc

PR #1171, opened 2023-06-05
Handle check_fn in S3KeySensor

pankajastro:
closes: https://github.com/astronomer/astronomer-providers/issues/1166

### DIFF
--- a/astronomer/providers/amazon/aws/sensors/s3.py
+++ b/astronomer/providers/amazon/aws/sensors/s3.py
@@ -1,9 +1,8 @@
 from __future__ import annotations
 
-import typing
 import warnings
 from datetime import timedelta
-from typing import Any, Callable, List, Sequence, cast
+from typing import Any, Callable, Sequence, cast
 
 from airflow.exceptions import AirflowException, AirflowSkipException
 from airflow.providers.amazon.aws.sensors.s3 import S3KeySensor, S3KeysUnchangedSensor
@@ -78,6 +77,8 @@ class S3KeySensorAsync(S3KeySensor):
             verify=verify,
             **kwargs,
         )
+        self.check_fn = check_fn
+        self.should_check = True if check_fn else False
 
     def execute(self, context: Context) -> None:
         """Check for a keys in s3 and defers using the trigger"""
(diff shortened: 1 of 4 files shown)


astronomer/providers/amazon/aws/triggers/s3.py
```diff
@@ -61,6 +63,7 @@ def serialize(self) -> tuple[str, dict[str, Any]]:
                 "hook_params": self.hook_params,
                 "poke_interval": self.poke_interval,
                 "soft_fail": self.soft_fail,
+                "should_check": self.should_check,
```
sunank200, line 66: What is `should_check`?

pankajastro: just a boolean based on this I'll see if need to return running status from trigger 
see astronomer/providers/amazon/aws/sensors/s3.py L81

astronomer/providers/amazon/aws/sensors/s3.py
```diff
@@ -89,27 +91,37 @@ def execute(self, context: Context) -> None:
...
         Callback for when the trigger fires - returns immediately.
         Relies on trigger to throw an exception, otherwise it assumes execution was
         successful.
         """
+        if event["status"] == "running":
+            found_keys = self.check_fn(event["files"])
+            if found_keys:
+                return None
+            else:
+                self._defer()
```
sunank200, line 124: This is just a function callable here. So why is it required to defer here? I didn't understand this @pankajastro

astronomer/providers/amazon/aws/triggers/s3.py
```diff
@@ -71,12 +74,12 @@ async def run(self) -> AsyncIterator[TriggerEvent]:
             async with await hook.get_client_async() as client:
                 while True:
                     if await hook.check_key(client, self.bucket_name, self.bucket_key, self.wildcard_match):
-                        if self.check_fn is None:
-                            yield TriggerEvent({"status": "success"})
+                        s3_objects = await hook.get_files(
+                            client, self.bucket_name, self.bucket_key, self.wildcard_match
+                        )
+                        if self.should_check:
+                            yield TriggerEvent({"status": "running", "files": s3_objects})
```
sunank200, line 81: What does `running` mean here?

pankajastro: sensor is not in terminal state

sunank200 (commented):
Another way to serialize python callable - [stackoverflow](https://stackoverflow.com/questions/6234586/we-need-to-pickle-any-sort-of-callable)

pankajastro:
> Another way to serialize python callable - [stackoverflow](https://stackoverflow.com/questions/6234586/we-need-to-pickle-any-sort-of-callable)

There are some security issues with pickle so maybe we can't use always

astronomer/providers/amazon/aws/sensors/s3.py
```diff
@@ -89,35 +90,39 @@ def execute(self, context: Context) -> None:
...
+            timeout=timedelta(seconds=self.timeout),
+            trigger=S3KeyTrigger(
+                bucket_name=cast(str, self.bucket_name),
+                bucket_key=self.bucket_key,
+                wildcard_match=self.wildcard_match,
+                aws_conn_id=self.aws_conn_id,
+                verify=self.verify,
+                poke_interval=self.poke_interval,
+                soft_fail=self.soft_fail,
+                should_check=self.should_check,
```
Lee-W, line 106: What happens if we pass the `check_fn` into `S3KeyTrigger`? I thought we should run this inside `S3KeyTrigger` instead of `execute_complete`

Lee-W:
> > Another way to serialize python callable - [stackoverflow](https://stackoverflow.com/questions/6234586/we-need-to-pickle-any-sort-of-callable)
> 
> There are some security issues with pickle so maybe we can't use always

I've discussed this with @uranusjr . It seems pickle won't work and the current solution might be good enough. Please correct me if I'm wrong. Thanks!

Issue #1154, opened 2023-05-31
Allow regex matching in S3KeySensorAsync


infohash:
**Is your feature request related to a problem? Please describe.**
The wildcard matching uses `fnmatch.fnmatch` that uses unix wildcard pattern to filter bucket keys but it has a limitation. It cannot distinguish between `archived_uploads/dataset.csv` and `dataset.csv` when your wildcard pattern is `*.csv`. As S3 does not use directories, that wildcard pattern cannot filter "top level" bucket keys that do not have the delimeter `/`. Because of this, S3KeySensorAsync always goes in success state even if you have not uploaded anything new.

However, there is a workaround, which is to use some prefix in your bucket keys for new file uploads and to not upload at the top level. E.g. If you are uploading new files, upload them as `new/dataset.csv` and the wildcard pattern for this case will be `new/*.csv`.

**Describe the solution you'd like**
A flag variable `use_regex` when set to `True` should use regex pattern.

https://github.com/astronomer/astronomer-providers/blob/6b29b0428ca3a2f26c31dc6040f3a6f9b5d01d67/astronomer/providers/amazon/aws/hooks/s3.py#L151-L157

```python
                 elif use_regex:
                     keys = await self.get_file_metadata(client, bucket_name, key)
                     key_matches = [k for k in keys if re.match(pattern=key, string=k["Key"])]
                     if not key_matches:
                         return False
```

E.g. If you have bucket keys `archived_uploads/dataset1.csv`, `archived_uploads/dataset2.csv`, `archived_uploads/dataset3.csv` and `dataset4.csv`, this is the pattern you will need to only poke for keys that do not have the delimeter '/', in this case `dataset4.csv` does not have this delimeter: 

```python
S3KeySensorAsync(
    task_id='sense_datasets',
    bucket_name='default',
    # Negative Lookbehind regex pattern which tells regex to not match names that are being preceded by the delimeter "/".
    bucket_key=r'(?<!/)[a-zA-Z0-9_]+\.csv',
    use_regex=True,
)
```

**Describe alternatives you've considered**
Solved it via `check_fn`.

```python
def check_for_files_without_delimeter(files: list[dict[str, datetime.datetime | int | str | dict[str, str]]]) -> bool:
    if any((file_metadata for file_metadata in files if re.match(pattern=r'(?<!/)[\w_]+\.csv', string=file_metadata ['Key'], flags=re.IGNORECASE))):
        return True
[...]

PR #1172, opened 2023-06-06
Add use_regex param in S3KeySensor

pankajastro:
closes: https://github.com/astronomer/astronomer-providers/issues/1154

### DIFF
--- a/astronomer/providers/amazon/aws/hooks/s3.py
+++ b/astronomer/providers/amazon/aws/hooks/s3.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import asyncio
 import fnmatch
 import os
@@ -5,7 +7,7 @@ import re
 from datetime import datetime
 from functools import wraps
 from inspect import signature
-from typing import Any, Dict, List, Optional, Set, TypeVar, cast
+from typing import Any, TypeVar, cast
 
 from aiobotocore.client import AioBaseClient
 from airflow.providers.amazon.aws.hooks.s3 import S3Hook, unify_bucket_name_and_key
@@ -54,8 +56,8 @@ class S3HookAsync(AwsBaseHookAsync):
     @provide_bucket_name_async
     @unify_bucket_name_and_key
     async def get_head_object(
-        self, client: AioBaseClient, key: str, bucket_name: Optional[str] = None
-    ) -> Optional[Dict[str, Any]]:
+        self, client: AioBaseClient, key: str, bucket_name: str | None = None
+    ) -> dict[str, Any] | None:
         """
         Retrieves metadata of an object
 
@@ -63,7 +65,7 @@ class S3HookAsync(AwsBaseHookAsync):
         :param bucket_name: Name of the bucket in which the file is stored
         :param key: S3 key that will point to the file
         """
-        head_object_val: Optional[Dict[str, Any]] = None
+        head_object_val: dict[str, Any] | None = None
         try:
             head_object_val = await client.head_object(Bucket=bucket_name, Key=key)
             return head_object_val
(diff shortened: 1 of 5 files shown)


astronomer/providers/amazon/aws/hooks/s3.py
```diff
@@ -114,15 +116,17 @@ async def list_prefixes(
...
+    ) -> list[Any]:
         """
         Gets a list of files that a key matching a wildcard expression exists in a bucket asynchronously
 
         :param client: aiobotocore client
         :param bucket_name: the name of the bucket
         :param key: the path to the key
         """
-        prefix = re.split(r"[\[\*\?]", key, 1)[0]
+        prefix = re.split(r"[\[\*\?]", key, 1)[0] if key else ""
```
Lee-W, line 129: Should we set the default value of `key` to "" instead of adding an if here? Or is there anything I missed?

Lee-W (approved):

astronomer/providers/amazon/aws/hooks/s3.py
pankajastro: yes, this make sense

astronomer/providers/amazon/aws/hooks/s3.py
```diff
@@ -148,13 +153,20 @@ async def _check_key(
         :param bucket_val: the name of the bucket
         :param key: S3 keys that will point to the file
         :param wildcard_match: the path to the key
+        :param use_regex: whether to use regex to check bucket
         """
+        pattern = key
```
infohash, line 158: No need to reassign the variable. Put it directly in `re.match(pattern=key, string=k["Key"])`.

tests/amazon/aws/hooks/test_s3_hooks.py
```diff
@@ -494,3 +494,27 @@ async def test_s3__check_key_with_wild_card(
             mock_client.return_value, "test_bucket", True, "test/example_s3_test_file.txt"
         )
         assert response is False
+
+    @pytest.mark.parametrize("key,pattern,expected", [("test.csv", "tes", True), ("test.txt", "abc", False)])
```
infohash, line 498: Use regex patterns for tests so that we can know that `re.match` is actually distinguishing between filenames with and without the delimeters:

```python
@pytest.mark.parametrize("key, pattern, expected", [("test.csv", r"[a-z]+\.csv", True), ("test.csv", r"test/[a-z]+\.csv", False), ("test.csv", r"test/[a-z]+\.csv", True)])
```

E.g.

```python
import re

# Check for keys without the delimeter "/".
>>> bool(re.match(pattern=r"[a-z]+\.csv", string="test.csv"))
    True
# Check for keys with the delimeter "/".
>>> bool(re.match(pattern=r"test/[a-z]+\.csv", string="test.csv"))
    False

>>> bool(re.match(pattern=r"test/[a-z]+\.csv", string="test/test.csv"))
    True
```

astronomer/providers/amazon/aws/hooks/s3.py
Lee-W: Ah yes, I misunderstood this initialization as well 🤦‍♂️ Thanks for clarification

tests/amazon/aws/hooks/test_s3_hooks.py
```diff
@@ -494,3 +494,30 @@ async def test_s3__check_key_with_wild_card(
             mock_client.return_value, "test_bucket", True, "test/example_s3_test_file.txt"
         )
         assert response is False
+
+    @pytest.mark.parametrize(
+        "key,pattern,expected",
```
Lee-W, line 499: nitpick

```suggestion
        "key, pattern, expected",
```

Lee-W (approved):

astronomer/providers/amazon/aws/hooks/s3.py
```diff
@@ -148,13 +153,19 @@ async def _check_key(
...
+        :param use_regex: whether to use regex to check bucket
         """
         bucket_name, key = S3Hook.get_s3_bucket_key(bucket_val, key, "bucket_name", "bucket_key")
         if wildcard_match:
             keys = await self.get_file_metadata(client, bucket_name, key)
             key_matches = [k for k in keys if fnmatch.fnmatch(k["Key"], key)]
             if len(key_matches) == 0:
                 return False
+        elif use_regex:
+            keys = await self.get_file_metadata(client, bucket_name)
```
infohash, line 165: Third parameter `key` (optional) is missing. Was this intentional?

PR #1174, opened 2023-06-07
Use min airflow version 2.4 for astronomer-providers

pankajkoti:
With the addition of PR #1169, we set the dependency 
`apache-airflow-providers-amazon>=8.1.0` which requires
min airflow version as `2.4.0`. Hence, we are setting the min 
airflow version to `2.4.0` for `astronomer-providers` too so that 
we're able to generate constraints successfully which failed generation 
for the PR upon merge to the `main`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,7 +34,7 @@ packages = find_namespace:
 include_package_data = true
 namespace_packages = astronomer,astronomer.providers
 install_requires =
-    apache-airflow>=2.2.0
+    apache-airflow>=2.4.0
     aiohttp
     aiofiles
     asgiref
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ workflows:
             parameters:
               python_version: ["3.8", "3.9", "3.10"]
               airflow_version: ["2.4.2", "2.5.3", "2.6.1"]
-          <<: *main_and_release_branches
+          <<: *all_branches_and_version_tag  # TODO: Set this back to *main_and_release_branches
       - build-docs:
           <<: *all_branches_and_version_tag
       - test:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,7 @@ workflows:
             parameters:
               python_version: ["3.8", "3.9", "3.10"]
               airflow_version: ["2.4.2", "2.5.3", "2.6.1"]
-          <<: *all_branches_and_version_tag  # TODO: Set this back to *main_and_release_branches
+          <<: *main_and_release_branches
       - build-docs:
           <<: *all_branches_and_version_tag
       - test:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,7 +38,7 @@ workflows:
           matrix:
             parameters:
               python_version: ["3.8", "3.9", "3.10"]
-              airflow_version: ["2.3.4", "2.4.2", "2.5.3"]
+              airflow_version: ["2.4.2", "2.5.3", "2.6.1"]
           <<: *main_and_release_branches
       - build-docs:
           <<: *all_branches_and_version_tag


setup.cfg
```diff
@@ -34,7 +34,7 @@ packages = find_namespace:
 include_package_data = true
 namespace_packages = astronomer,astronomer.providers
 install_requires =
-    apache-airflow>=2.2.0
+    apache-airflow>=2.4.0
```
sunank200, line 37: Should we add a comment on why we changed the minimum version to 2.4.0?

pankajkoti: I am not sure if we need to add a comment every time we change the min version. I think git blame on the line should take us to the PR and we could then read the PR description on why the change was made.
WDYT? 

Additionally, @sunank200, I would also like to hear your opinion on that since we're changing the minimum version, is it fine to push this PR and PR #1169 in release `1.17.0` of `astronomer-providers` or do we need to wait until the `2.0.0` major release for us to upgrade the base Airflow version?

@phanikumv @pankajastro @Lee-W please also share your opinion on this one.

pankajkoti: I have seen Airflow OSS providers bumping the minimum Airflow version in minor releases 
e.g. https://airflow.apache.org/docs/apache-airflow-providers-postgres/stable/index.html#id1 The Airflow Postgres provider bumped min Airflow version from `2.3.0` to `2.4.0` in minor release `5.5.0` of the provider.

So, ideally, it should not stop us from including the change in a minor release, too. WDYT?

pankajkoti:
Constraints generation was successful with this change as can be seen in the run. Hence, I will now revert the commit which was introduced to run the constraints generation CircleCI branch on this PR branch. 

Screenshot 2023-06-07 at 4 14 17 PM

setup.cfg
Lee-W: Based on the guidelines provided on [semver.org](https://semver.org/), bump a `MAJOR version when you make incompatible API changes`. Therefore, I think it's probably ok for us to bump the version to `1.17.0`. If we are comfortable with bumping lots of major versions, upgrading to `2.0.0` may be preferable. But I'm not sure whether it's the case

phanikumv (changes requested):
I think PR#https://github.com/astronomer/astronomer-providers/pull/1169 is wrongly upgrading the amazon provider.what abt users who are on 2.2.0?

setup.cfg
pankajkoti: That is the question @phanikumv. Would PR #1169 and this PR #1174 have to wait until release 2.0.0 of providers?

Wouldn't constraints help here? Saying if they want to use `1.17.0`, they need to have Airflow `2.4.0`?

Also, we don't have upper bounds for upstream providers. So, users may still be installed with newer versions of upstream providers, no?

pankajkoti:
> I think PR##1169 is wrongly upgrading the amazon provider.what abt users who are on 2.2.0?

I attempted to answer this question in the ongoing conversation thread https://github.com/astronomer/astronomer-providers/pull/1174#discussion_r1221432700

setup.cfg
pankajkoti: Also, as I mentioned in the earlier comment https://github.com/astronomer/astronomer-providers/pull/1174#discussion_r1221367447, Airflow OSS providers had minor releases bumping the min Airflow version requirement.

pankajkoti: Airflow docs say here https://github.com/apache/airflow/blob/main/PROVIDERS.rst#minimum-supported-version-of-airflow-for-community-managed-providers the below:
```
When we increase the minimum Airflow version, this is not a reason to bump MAJOR version of the providers (unless there are other breaking changes in the provider). The reason for that is that people who use older version of Airflow will not be able to use that provider (so it is not a breaking change for them) and for people who are using supported version of Airflow this is not a breaking change on its own - they will be able to use the new version without breaking their workflows. When we upgraded min-version to 2.2+, our approach was different but as of 2.3+ upgrade (November 2022) we only bump MINOR version of the provider when we increase minimum Airflow version.
```

pankajkoti:
Closing this as we need to wait until the major release of providers 2.0.0 as per discussion with the team.

setup.cfg
pankajkoti: @pankajastro could you link the document you were mentioning please?

pankajkoti: okay found it. I think it was this one https://www.notion.so/astronomerio/Draft-Reduce-maintenance-effort-and-keep-backward-Compatibility-for-astronomer-providers-after-con-e16857c5ea814c71afe5009c674d0d67

PR #1176, opened 2023-06-12
Use env variable to control GCP nuke DAG schedule

pankajkoti:
The current GCP Nuke DAG schedule is colluding with our
night runs and hence we're observing GCP DAG failures
at times. This allows us to set the nuke dag schedule with
an environment variable and we plan to run it at the same
time as AWS nuke DAG which runs twice; once before and once
after our nightly run.

### DIFF
--- a/astronomer/providers/google/cloud/example_dags/example_gcp_nuke.py
+++ b/astronomer/providers/google/cloud/example_dags/example_gcp_nuke.py
@@ -1,3 +1,4 @@
+"""DAG to delete stale GCP resources."""
 import os
 from datetime import timedelta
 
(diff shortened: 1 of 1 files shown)


astronomer/providers/google/cloud/example_dags/example_gcp_nuke.py
```diff
@@ -7,6 +8,7 @@
 from airflow.utils.timezone import datetime
 
 REGION = os.getenv("GCP_LOCATION", "us-central1")
+GCP_NUKE_DAG_SCHEDULE = os.getenv("GCP_NUKE_DAG_SCHEDULE", None)
```
pankajastro, line 11: why did you choose the default value None and not `30 20 * * *`?

pankajastro (approved):

astronomer/providers/google/cloud/example_dags/example_gcp_nuke.py
pankajkoti: Yes, we do the same in the AWS Nuke schedule. If we set the default schedule, since we now have 2 deployments, it will trigger the DAG from both deployments at the same time and it would be redundant(might also cause conflicts). We only schedule it from the CeleryExecutor deployment by setting this env var there, whereas the KubernetesExecutor deployment does not have this env var, so it won't be scheduled from there/.

PR #1179, opened 2023-06-13
Mark Async Redshift Cluster DAG failed if any of the tasks fail

vatsrahul1001:
fail.

### DIFF
--- a/astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
@@ -149,5 +149,3 @@ with DAG(
         >> dag_final_status
     )
 
-    [resume_redshift_cluster, redshift_sensor] >> dag_final_status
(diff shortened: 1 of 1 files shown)


astronomer/providers/amazon/aws/example_dags/example_redshift_cluster_management.py
```diff
@@ -128,5 +148,5 @@
         >> delete_redshift_cluster
     )
```
pankajkoti, line 149:
```suggestion
        >> delete_redshift_cluster
        >> dag_final_status
    )
```

pankajkoti:
Also, could we please add a screenshot of the DAG run or the Graph view to ensure that these changes do not cause a regression.

vatsrahul1001:
> Also, could we please add a screenshot of the DAG run or the Graph view to ensure that these changes do not cause a regression.

@pankajkoti Yes I did check this DAG locally 
image

pankajkoti (approved):
Thank you for this change, Rahul!

PR #1180, opened 2023-06-13
Adjust S3 file size check check

pankajastro:
Look like our file size is ~30B but check_fn is checking for 1024B

### DIFF
--- a/astronomer/providers/amazon/aws/example_dags/example_s3.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_s3.py
@@ -70,7 +70,7 @@ with DAG(
         :return: true if the criteria is met
         :rtype: bool
         """
-        return all(f.get("Size", 0) > 1024 for f in files)
+        return all(f.get("Size", 0) > 20 for f in files)
 
     # [END howto_sensor_s3_key_function_definition]
     create_bucket = S3CreateBucketOperator(


Lee-W (approved):

astronomer/providers/amazon/aws/example_dags/example_s3.py
```diff
@@ -70,7 +70,7 @@ def check_fn(files: List[Any]) -> bool:
         :return: true if the criteria is met
         :rtype: bool
         """
-        return all(f.get("Size", 0) > 1024 for f in files)
+        return all(f.get("Size", 0) > 20 for f in files)
```
pankajkoti, line 73: Can we update the docstring then?

pankajastro: I was thinking if we should change it to `f.get("Size", 0) > 0`. This way it might be more stable and we are also testing the check function wdyt?

sunank200 (approved):

astronomer/providers/amazon/aws/example_dags/example_s3.py
pankajkoti: yes, was going to suggest > 0, but thought you want to intentionally keep it to that size based on the files we want to test :)

PR #1181, opened 2023-06-13
Setting Factory name as ENV Var

vatsrahul1001:
We have recently observed that the current usage of the "factory_name" as a static value. Taking it from an environment variable will help us run the DAG in parallel.

### DIFF
--- a/.circleci/integration-tests/master_dag.py
+++ b/.circleci/integration-tests/master_dag.py
@@ -45,6 +45,7 @@ MODEL_NAME
 REDSHIFT_CLUSTER_IDENTIFIER
 REDSHIFT_TABLE_NAME
 RESOURCE_GROUP_NAME
+ADF_FACTORY_NAME
 
 Additionally, ensure to have unique "cluster_identifier" values for each deployment in the below Airflow connections:
 aws_default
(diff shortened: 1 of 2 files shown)


.circleci/integration-tests/master_dag.py
```diff
@@ -45,6 +45,7 @@
 REDSHIFT_CLUSTER_IDENTIFIER
 REDSHIFT_TABLE_NAME
 RESOURCE_GROUP_NAME
+ADF_FACTORY_NAME
```
pankajkoti, line 48: Can we place it at the right location? This listed is alphabetically sorted

astronomer/providers/microsoft/azure/example_dags/example_adf_run_pipeline.py
```diff
@@ -23,6 +23,7 @@
 TENANT_ID = os.getenv("TENANT_ID", "")
 SUBSCRIPTION_ID = os.getenv("SUBSCRIPTION_ID", "")
 RESOURCE_GROUP_NAME = os.getenv("RESOURCE_GROUP_NAME", "team_provider_resource_group_test_1")
+ADF_FACTORY_NAME = os.getenv("ADF_FACTORY_NAME", "ADFProvidersTeamDataFactoryTest")
```
pankajkoti, line 26: We already have a variable below `DATAFACTORY_NAME` (which is also set in our Astro Deployments), perhaps we could just use that in the `default_args` dictionary below.

vatsrahul1001: @pankajkoti fixed

PR #1185, opened 2023-06-15
Mark External Task DAG failed if any of the tasks fail

vatsrahul1001:
fail.

Tested DAG locally after the update
image

### DIFF
--- a/astronomer/providers/core/example_dags/example_external_task.py
+++ b/astronomer/providers/core/example_dags/example_external_task.py
@@ -1,12 +1,15 @@
 import os
 import time
 from datetime import timedelta
+from typing import Any
 
 from airflow import DAG
 from airflow.operators.dummy import DummyOperator
 from airflow.operators.python import PythonOperator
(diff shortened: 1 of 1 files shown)


Lee-W (approved):

astronomer/providers/core/example_dags/example_external_task.py
```diff
@@ -62,9 +76,15 @@
...
+        retries=0,
+    )
 
-    start >> [waiting_for_task, wait_for_me] >> end
+    start >> [waiting_for_task, wait_for_me] >> dag_final_status
 
-    start >> wait_for_dag >> end
-    start >> wait_to_defer >> external >> end
+    start >> wait_for_dag >> dag_final_status
+    start >> wait_to_defer >> external >> dag_final_status
```
pankajkoti, line 90: There should be a way to maintain these dependencies in a cleaner way. But, it's okay for now. I will take a look at it later to clean up. Looks fine to merge for now.

pankajkoti (approved):

PR #1186, opened 2023-06-15
Mark DBT Cloud DAG failed if any of the tasks fail

vatsrahul1001:
fail.

Tested DAG locally after the update
image

### DIFF
--- a/astronomer/providers/dbt/cloud/example_dags/example_dbt_cloud.py
+++ b/astronomer/providers/dbt/cloud/example_dags/example_dbt_cloud.py
@@ -5,7 +5,7 @@ from datetime import timedelta
 from typing import Any
 
 from airflow import DAG
-from airflow.operators.empty import EmptyOperator
+from airflow.operators.dummy import DummyOperator
 from airflow.operators.python import PythonOperator
 from airflow.providers.dbt.cloud.operators.dbt import DbtCloudRunJobOperator
 from airflow.utils.state import State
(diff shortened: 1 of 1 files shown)


Lee-W (approved):

pankajkoti (approved):

astronomer/providers/dbt/cloud/example_dags/example_dbt_cloud.py
```diff
@@ -2,11 +2,15 @@
 
 import os
 from datetime import timedelta
+from typing import Any
 
 from airflow import DAG
 from airflow.operators.empty import EmptyOperator
```
pankajkoti, line 8:
```suggestion
from airflow.operators.dummy import DummyOperator
```

astronomer/providers/dbt/cloud/example_dags/example_dbt_cloud.py
```diff
@@ -34,7 +49,7 @@
     catchup=False,
 ) as dag:
     start = EmptyOperator(task_id="start")
```
pankajkoti, line 51:
```suggestion
    start = DummyOperator(task_id="start")
```

pankajkoti (approved):

pankajkoti:
@vatsrahul1001 I applied a small patch to use DummyOperator instead of EmptyOperator as it's usage is deprecated.

PR #1190, opened 2023-06-19
Run EMR DAGs in parallel to reduce Master DAG run duration

pankajkoti:
Currently, we chain the EMR DAGs, and the EMR EKS PI Job takes 
between 30 to 40 mins and the EMR Sensor DAGs takes about 
8-10 mins. As a result, due to these long-running DAGs, the 
Master DAG takes 38 to 50 mins. The EMR EKS DAG takes the 
longest duration across all the DAGs in Master DAG. With 
parallelising this chain, we're expecting the Master DAG will run 
for a duration at most the same as the EMR EKS PI job and then it 
does not have to wait for additional 8-10 mins for the EMR sensor 
DAG to finish as it would be run parallelly and would have 
completed by then.

The commit also removes additional free lines in the Slack report as it looks unnecessary.

### DIFF
--- a/.circleci/integration-tests/master_dag.py
+++ b/.circleci/integration-tests/master_dag.py
@@ -69,7 +69,7 @@ def get_report(dag_run_ids: List[str], **context: Any) -> None:
             airflow_version_message = f"The below run is on Airflow version `{airflow_version} with {airflow_executor} executor`\n\n"
 
         master_dag_deployment_link = f"{os.environ['AIRFLOW__WEBSERVER__BASE_URL']}/dags/example_master_dag/grid?search=example_master_dag"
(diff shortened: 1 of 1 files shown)


pankajkoti:
Gannt chart before when EMR Sensor DAG run after EMR EKS PI job. Took about 40-45 mins on average.

Screenshot 2023-06-19 at 1 00 11 PM

pankajkoti:
Gantt chart now when EMR EKS and EMR sensor run in parallel. The master DAG run duration is down to 33 mins in this instance.

Screenshot 2023-06-19 at 12 59 58 PM

Lee-W (approved):

PR #1191, opened 2023-06-19
Delete stale IAM roles left by EMR EKS PI job DAG

pankajkoti:
Ideally, the "emr_eks_pi_job" DAG should delete the IAM roles it creates. 
However, there were observed occurrences where the IAM roles were not 
deleted, most probably because we did not allow to run the delete cluster 
task to completion or did not allow the task to run altogether due to killing 
of the deployment before it finished running. With this commit, we add a 
script that automates the deletion of such stale IAM roles by first detaching 
its associated policies. We call this script as a task from the AWS nuke DAG 
which is scheduled to run after the end of the master DAG run.

### DIFF
--- a/astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
@@ -139,6 +139,11 @@ with DAG(
         trigger_rule="all_done",
     )
 
+    delete_stale_emr_iam_roles = BashOperator(
+        task_id="delete_stale_emr_iam_roles",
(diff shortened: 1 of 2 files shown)


pankajkoti:
Stale IAM roles from the last 3-4 days
Screenshot 2023-06-19 at 10 08 44 PM

If the environment variable generates the same 2-digit random number as a suffix for the IAM role as these stale ones, the DAG fails to say the role already exists.

pankajkoti:
DAG run for the nuke task of deleting stale IAM roles created by EMR EKS PI job
Screenshot 2023-06-19 at 10 28 01 PM

Lee-W (approved):

PR #1193, opened 2023-06-21
ci(github-actions): fix wrong variable used in test-rc workflow

Lee-W:
This is the root cause of the failure https://github.com/astronomer/astronomer-providers/actions/runs/5328666545/jobs/9653537662

### DIFF
--- a/.github/workflows/test-rc-release.yaml
+++ b/.github/workflows/test-rc-release.yaml
@@ -170,7 +170,7 @@ jobs:
           git add setup.cfg
           git commit -m "Updating setup.cfg with RC provider packages on $rc_issue_url"
           git push origin ${{ steps.create_rc_branch.outputs.rc_testing_branch }}
-          gh pr create --base ${{ inputs.base_git_rev }} \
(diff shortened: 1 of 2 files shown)


pankajastro (approved):

pankajkoti (commented):
Was the PR https://github.com/astronomer/astronomer-providers/pull/1194 created after this change? Mean to ask if we verified that our script works now? :)

Lee-W:
No, we'll need to wait for tomorrow for it to run. Thus, I run the workflow manually with the branch created by the bot for testing it earlier.

pankajkoti:
But tomorrow it won't run since we already have the branch/PR created now, right? 🤔

Lee-W:
Ideally, yes; if not, then we have more fixes to go for the workflow 🤦‍♂️ So for actually test the full round, I think we'll need to wait for next provider release

pankajkoti:
okay, understood, thank you! :)

Lee-W:
Here's the pr for that fixing https://github.com/astronomer/astronomer-providers/pull/1207 🤦‍♂️

PR #1194, opened 2023-06-21
[DO NOT MERGE] Test RC provider packages for https://github.com/apache/airflow/issues/32030

Lee-W:

PR #1195, opened 2023-06-21
Fix mypy errors observed in June Providers wave

pankajkoti:
We observed mypy failures while testing the June wave of providers
in our RC testing PR https://github.com/astronomer/astronomer-providers/pull/1194.

We try to fix those issues here and will be applied once the providers are released


pankajkoti:
cc: @Lee-W

Lee-W (approved):

Lee-W:
Let's merge it after the release 🙂

pankajkoti:
The providers are released now. So merging this PR now.

PR #1196, opened 2023-06-21
Automate deletion of RC testing branch once PR is closed

pankajkoti:
Our current GitHub configuration automatically deletes the branches
when PRs are merged. But for closed PRs, the branches are not deleted
and they remain stale. We have cases like testing the RCs, where we do
not want to merge the PR but only close it after verifying that our RCs
do not cause static check or tests failure. For such RC testing branches, 
this action will delete the branches upon closing the PR. 

In case, we want to reopen the PR, we need to click the button to 
restore the branch first and only then the reopen PR option will be
enabled by GitHub. I made an attempt to restore the branch if the event
is of type PR reopen, but that event cannot happen when the branch
no longer exists. Hence, we will have to manually click on the restore
button first if we want to reopen the PR.

### DIFF
--- a/.github/workflows/delete-branch-on-pr-close.yaml
+++ b/.github/workflows/delete-branch-on-pr-close.yaml
@@ -1,0 +1,27 @@
+---
+name: Delete RC testing branch upon PR closure
+
+on:  # yamllint disable-line rule:truthy
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  delete-branch:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+
+      - name: Verify and Delete RC testing branch
+        run: |
+          git config --global user.name "GitHub Actions"
+          git config --global user.email "action@github.com"
+          git fetch --prune --all
+
+          if [[ "${{ github.event.pull_request.head.ref }}" == rc-test-* ]]; then
+            git push origin --delete ${{ github.event.pull_request.head.ref }}
+          else
+            echo "Branch does not have the required RC testing prefix. Skipping branch deletion."
+          fi


pankajkoti:
cc: @Lee-W following up with the discussion in PR #1194

.github/workflows/delete-branch-on-pr-close.yaml
```diff
@@ -0,0 +1,22 @@
+---
```
Lee-W, line 1: I'm not sure whether we should delete the branch in such an aggressive way. For example, @phanikumv wants to keep the https://github.com/astronomer/astronomer-providers/tree/add-postgres-async, but with this action, we'll remove that as well. I'm thinking of doing it only for the testing rc branches

phanikumv: yes we need to cleanup for rc test branches only

pankajkoti: we can always restore the branch if we accidentally delete the branch or upon PR closure. I don't think deletion is permanent delete. But will check that we've such a prefix.

pankajkoti: This action will only delete the branch when a PR is closed. And branches deleted from close PRs can be restored AFAIK https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-branches-in-your-repository/deleting-and-restoring-branches-in-a-pull-request#restoring-a-deleted-branch

Lee-W: Wow, I didn't know it. Thanks for sharing this!

.github/workflows/delete-branch-on-pr-close.yaml
```diff
@@ -0,0 +1,27 @@
...
+    types:
+      - closed
+
+jobs:
+  delete-branch:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v2
```
Lee-W, line 15: I notice there's already v3 for this action. Maybe we could try to use the latest version?

Lee-W (approved):

.github/workflows/delete-branch-on-pr-close.yaml
pankajkoti, line 15:
```suggestion
        uses: actions/checkout@v3
```

pankajkoti: updated, thank you !

pankajkoti:
As we can see the branch of this PR was not deleted upon PR closure
since it does not have the `rc-test-` prefix. I created another PR having
the branch name prefix as `rc-test-` and verified that such branch was
deleted upon PR closure in PR #1197

Lee-W (approved):

PR #1198, opened 2023-06-21
Add Dataproc job object in DataProcSubmitTrigger yield message

pankajastro:
- Remove param location and wait_timeout from the DataProcSubmitTrigger docs string since these do not belong to DataProcSubmitTrigger class. 
- Add Dataproc job object in DataProcSubmitTrigger yield event this will have some info about the job 

https://cloud.google.com/python/docs/reference/dataproc/latest/google.cloud.dataproc_v1.types.Job 
https://cloud.google.com/python/docs/reference/dataproc/latest/google.cloud.dataproc_v1.types.JobStatus

Screenshot 2023-07-04 at 7 54 01 PM

### DIFF
--- a/astronomer/providers/google/cloud/triggers/dataproc.py
+++ b/astronomer/providers/google/cloud/triggers/dataproc.py
@@ -278,9 +278,7 @@ class DataProcSubmitTrigger(BaseTrigger):
     :param region: Required. The Cloud Dataproc region in which to handle the request. (templated)
     :param project_id: The ID of the google cloud project in which
         to create the cluster. (templated)
-    :param location: (To be deprecated). The Cloud Dataproc region in which to handle the request. (templated)
     :param gcp_conn_id: The connection ID to use connecting to Google Cloud Platform.
-    :param wait_timeout: How many seconds wait for job to be ready.
     :param impersonation_chain: Optional service account to impersonate using short-term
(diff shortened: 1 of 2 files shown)


sunank200:
> Can you please enhance the description as to what all and why we're making these changes?
> 
> Also, can you please paste a screenshot of the details we will have now that solves the issue/ helps debug the issue?

this change LGTM once the above changes are done.

Lee-W (approved):

pankajkoti (commented):
Were we able to reproduce the issue and see some logs? Or are we not able to reproduce and want to merge this to see the logs when the issue comes for further debugging?

Can we verify once running the DAG locally/cloud to make sure the DAG runs fine with these changes?

pankajastro:
> Were we able to reproduce the issue and see some logs? Or are we not able to reproduce and want to merge this to see the logs when the issue comes for further debugging?

I was not able to reproduce so thought of adding this to get some insight when error occur

> Can we verify once running the DAG locally/cloud to make sure the DAG runs fine with these changes?

I do it sometime tomorrow

pankajkoti:
okay, understood, I think we can merge this once we verify that the DAG runs fine with these changes. @Lee-W has already approved it so we should be able to merge this one then.

PR #1206, opened 2023-06-22
ci(github-actions): fix wrong base branch used when creating pull request

Lee-W:
we should set the main branch as base instead of the changed branch

### DIFF
--- a/.github/workflows/test-rc-release.yaml
+++ b/.github/workflows/test-rc-release.yaml
@@ -170,7 +170,7 @@ jobs:
           git add setup.cfg
           git commit -m "Updating setup.cfg with RC provider packages on $rc_issue_url"
           git push origin ${{ steps.create_rc_branch.outputs.rc_testing_branch }}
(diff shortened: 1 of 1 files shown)


.github/workflows/test-rc-release.yaml
```diff
@@ -170,7 +170,7 @@ jobs:
           git add setup.cfg
           git commit -m "Updating setup.cfg with RC provider packages on $rc_issue_url"
           git push origin ${{ steps.create_rc_branch.outputs.rc_testing_branch }}
-          gh pr create --base ${{ steps.create_rc_branch.outputs.rc_testing_branch }} \
+          gh pr create --base main \
```
pankajkoti, line 173: Do we need to specify base explicitly as I think it's our default branch and will be used as base anyway, no? If that is the case we may not need to specify --base argument.

Lee-W: yes, I think we can ignore it. let me remove it.

Lee-W: just removed

pankajkoti (approved):

Issue #1208, opened 2023-06-22
KubernetesPodOperatorAsync failing in astronomer-providers version :1.17.0


vatsrahul1001:
**Describe the bug**
**KubernetesPodOperatorAsync** failing in provider version:1.17.0 

**To Reproduce**
1. Create a deployment with astronomer-providers version 1.17.0 
2. Deploy the DAG mentioned below in the deployment
3. Try running the DAG

You can check this [deployment](https://astro01.astronomer-stage.run/dib05g0v/dags/async_kpo/grid?root=&dag_run_id=manual__2023-06-22T10%3A07%3A06.940234%2B00%3A00&task_id=simple_async&tab=logs)  for quick check

Looks like issue is related to [commit](https://github.com/astronomer/astronomer-providers/commit/89ccc7ec1ad5c38a1665b50f649e9153d3ca2f37#diff-655f3679acaf9e6930348f717981308b8e223241734253af938b2db193ef61e1R86)

**DAG CODE**
```
`from datetime import datetime

from airflow import DAG
from airflow.configuration import conf
try:
    from astronomer.providers.cncf.kubernetes.operators.kubernetes_pod import (
        KubernetesPodOperatorAsync,
    )
except ImportError:
    from astronomer_operators.cncf.kubernetes.operators.kubernetes_pod import (
        KubernetesPodOperatorAsync,
    )

namespace = conf.get("kubernetes", "NAMESPACE")

if namespace == "default":
    config_file = None
    in_cluster = False
else:
    in_cluster = True
    config_file = None


with DAG(
    dag_id="async_kpo",
    start_date=datetime(2022, 1, 1),
    schedule_interval="@daily",
    tags=["k8s", "core", "async"],
    max_active_runs=1,
) as dag:
    simple_async = KubernetesPodOperatorAsync(
        task_id="simple_async",
        namespace=namespace,
        in_cluster=in_cluster,
        config_file=config_file,
        name="simple_async",
        image="ubuntu",
        cmds=["/bin/sh"],
        arguments=["-c", "sleep 30"],
    )`
```

**Expected behavior**
DAG should run without any error

**LOGS**
`172.20.14.173
*** Reading remote log from s3://airflow-logs-cli988ld400jv01m4t55v2hc3/clj6tzyfg2845601mzviib05g0v/dag_id=async_kpo/run_id=manual__2023-06-22T10:07:06.940234+00:00/task_id=simple_async/attempt=1.log.
[...]

PR #1209, opened 2023-06-22
Revert "feat(kubernetes): check state before deferring KubernetesPodOperatorAsync (#1104)"

pankajkoti:
This reverts commit 89ccc7ec1ad5c38a1665b50f649e9153d3ca2f37.

PR #1104 adds logic to poke for Pod status before putting the status 
check on deferral. However, it is observed that our DAG fails always 
when we go on checking the status.phase on pods as Pod.status is None 
before the pod gets scheduled on a node. So, in most scenarios it does 
not make sense to check for the pod status immediately to verify that the 
pod has completed its desired execution and hence, we revert this poke 
in case of Kubernetes Pod operator.

closes: #1208


phanikumv (approved):
Let's also remove retries from the corresponding integration DAG

Issue #1201, opened 2023-06-21
[Releasing process] automatically create corresponding git tag after merging releasing branches


Lee-W:
**Is your feature request related to a problem? Please describe.**
Automate [this step](https://github.com/astronomer/astronomer-providers/blob/main/RELEASING.rst#tag-and-push-the-commit)

**Describe the solution you'd like**
Parse the merged pull request with branch name "release-a-b-c" and create git tag `a.b.c`

**Describe alternatives you've considered**

**Additional context**

pankajkoti:
Perhaps we could also check if we could automate the dev version bump post the release.

Lee-W:
Yep, that shouldn't be too difficult and should be included in this change as well. Thanks for your suggestion!

Lee-W:
Another thing I'm thinking of (not related to this issue) is whether we could somehow tag a PR as included in the changelog before merging so that we don't need to check all the commits and simply parse the commits with that pattern

pankajkoti:
yes, I think we can refer to Airflow OSS on how they achieve building the ChangeLog for RCs. I believe they also add a label `changelog:skip` to PRs that they want to be skipped in ChangeLog. I think these ideas can be separate issues and separate PRs that will allow us to have incremental improvements :)

PR #1214, opened 2023-06-26
Automate part of the release process

Lee-W:
* add `release` workflow for automating tagging process
* fix missing `needs` in the test rc release workflow
* rename workflows

closes: #1201 

We'll need to allow `action@github.com` to push tag and commit to main

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,37 @@
+---
+name: Trigger release and bump version
+
+on:  # yamllint disable-line rule:truthy
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  trigger-release-process:
+    if: github.event.pull_request.merged == true && startswith(github.event.head_commit.message, 'Release ')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out
+        uses: actions/checkout@v3
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Install commitizen
+        run: |
+          pip3 install commitizen
+
+      - name: Setup Github Actions git user
+        run: |
+          git config --global user.email "action@github.com"
+          git config --global user.name "GitHub Actions"
+
+      - name: Tag and push tag for triggering release
+        run: |
+          git tag `cz version -p`
+          git push origin `cz version -p`
+
+      - name: Bump version for new development
+        run: |
+          make ASTRO_PROVIDER_VERSION=dev bump-version
+          git add .
+          git push origin main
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,34 @@
+---
+name: Trigger release and bump version
+
+on:
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  trigger-release-rpceoss:
```
phanikumv, line 10:
```suggestion
  trigger-release-process:
```

Lee-W:
> Should we club test-rc-release workflow changes in this PR? If it is not needed/related to this work I think good to have it in a separate PR. WDYT?

Sounds good 🙂

.github/workflows/test-rc-release.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-name: test-rc-release
+name: Test providers RC releases
```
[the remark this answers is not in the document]
Lee-W: Sure 🙂

Lee-W:
I've done some investigation and testing with the permission. It seems `GITHUB_TOKEN` does not have enough permission to push the tag/branch to a protected tag/branch. Thus, we have 2 ways to achieve it.

1. Use the access token https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens created by someone with the write permission and set this value to GitHub secret (like what we did for the deployment secret). We can name it as `GH_TOKEN` or something else (I'll go with `GH_TOKEN`). If we opt to go this route, we might need @phanikumv 's help on setting these up.
2. Create a GitHub account for astronomer, grant it with such permission https://github.blog/changelog/2022-08-18-bypass-branch-protections-with-a-new-permission/ or allow it to bypass rule (which can be set through settings in this repo)

Lee-W:
> Add the instruction on how to use/trigger this process in `RELEASING.rst`

This is automatically triggered. I just updated the document.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,37 @@
...
+name: Trigger release and bump version
+
+on:  # yamllint disable-line rule:truthy
+  pull_request:
+    types:
+      - closed
+
+jobs:
+  trigger-release-process:
+    if: github.event.pull_request.merged == true && startswith(github.event.head_commit.message, 'Release ')
```
phanikumv, line 11: Can we make this condition more tighter somehow - I am a bit concerned with the condition for a new release being just a PR being merged with "Release" in the message

Lee-W: I don't think we can easily do so through GitHub actions. Thus, I implement a simple python script for this validation.

phanikumv (approved):

PR #1216, opened 2023-06-26
Use upstream aws base hook session implementation 

pankajastro:
closes: https://github.com/astronomer/astronomer-providers/issues/1227

We have missing implementation for some auth type so I was thinking if it makes sense to use upstream provider base hook implementation. 

Screenshot 2023-07-05 at 12 40 33 AM

### DIFF
--- a/astronomer/providers/amazon/aws/example_dags/example_batch.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_batch.py
@@ -288,8 +288,8 @@ with DAG(
     )
     # [END howto_operator_batch_async]
 
-    submit_batch_job_assume_cred = BatchOperatorAsync(
-        task_id="assume_role_cred",
+    submit_job_with_assume_role = BatchOperatorAsync(
+        task_id="submit_job_with_assume_role",
         job_name=JOB_NAME,
         job_queue=JOB_QUEUE,
         job_definition=JOB_DEFINITION,
@@ -347,7 +347,7 @@ with DAG(
         >> submit_batch_job
         >> list_jobs
         >> batch_job_sensor
-        >> submit_batch_job_assume_cred
+        >> submit_job_with_assume_role
         >> disable_compute_environment
         >> disable_job_queue
         >> delete_job_queue
--- a/astronomer/providers/amazon/aws/example_dags/example_batch.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_batch.py
@@ -18,6 +18,7 @@ from astronomer.providers.amazon.aws.sensors.batch import BatchSensorAsync
 
 AWS_DEFAULT_REGION = os.getenv("AWS_DEFAULT_REGION", "us-east-2")
 AWS_CONN_ID = os.getenv("ASTRO_AWS_CONN_ID", "aws_default")
+AWS_CONN_ID_ASSUME_ROLE = os.getenv("ASTRO_AWS_ASSUME_ROLE_CONN_ID", "aws_assume_role_conn")
 AWS_ACCESS_KEY_ID = os.getenv("AWS_ACCESS_KEY_ID", "xxxxxxx")
 AWS_SECRET_ACCESS_KEY = os.getenv("AWS_SECRET_ACCESS_KEY", "xxxxxxxx")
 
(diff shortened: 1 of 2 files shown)


pankajastro:
@phanikumv wdyt?

phanikumv:
> @phanikumv wdyt?

yes it makes sense

pankajkoti (commented):
Let's test all our example DAGs on Astro before merging these.

astronomer/providers/amazon/aws/hooks/base_aws.py
```diff
@@ -34,6 +34,12 @@ class AwsBaseHookAsync(AwsBaseHook):
 
     async def get_client_async(self) -> AioBaseClient:
         """Create an Async Client object to communicate with AWS services."""
+        # Try to use airflow provider client implementation
+        try:
+            return self._get_aioclient()
```
phanikumv, line 39: Let's also include assume role config in our aws connection so that we can test it

pankajastro: I added some  https://github.com/apache/airflow/pull/28850/files#diff-4e7fd0a0e31b60e534f7cab71ee4d3591c0758c1e76054f2403b6669546cb2d2 do you want me to replicate the same here

phanikumv: I was looking for an integration test for `assume_role` functionality

pankajastro: created connection in cloud account and added a operator please take a look

astronomer/providers/amazon/aws/example_dags/example_batch.py
```diff
@@ -287,6 +288,16 @@ def check_dag_status(**kwargs: Any) -> None:
     )
     # [END howto_operator_batch_async]
 
+    submit_batch_job_assume_cred = BatchOperatorAsync(
+        task_id="assume_role_cred",
```
pankajastro, line 292:
```suggestion
        task_id="submit_job_with_assume_role",
```

astronomer/providers/amazon/aws/example_dags/example_batch.py
```diff
@@ -287,6 +288,16 @@ def check_dag_status(**kwargs: Any) -> None:
     )
     # [END howto_operator_batch_async]
 
+    submit_batch_job_assume_cred = BatchOperatorAsync(
```
pankajastro, line 291:
```suggestion
    submit_job_with_assume_role = BatchOperatorAsync(
```

astronomer/providers/amazon/aws/example_dags/example_batch.py
```diff
@@ -336,6 +347,7 @@ def check_dag_status(**kwargs: Any) -> None:
         >> submit_batch_job
         >> list_jobs
         >> batch_job_sensor
+        >> submit_batch_job_assume_cred
```
pankajastro, line 350:
```suggestion
        >> submit_job_with_assume_role
```

astronomer/providers/amazon/aws/hooks/base_aws.py
phanikumv: @pankajastro Could not find it on our deployments. Can you please let me know where this is added

astronomer/providers/amazon/aws/hooks/base_aws.py
```diff
@@ -152,3 +158,12 @@ async def get_role_credentials(
                 )
                 return_response = response["Credentials"]
             return return_response
+
+    def _get_aioclient(self) -> AioBaseClient:
+        session = self.get_session(region_name=self.region_name, deferrable=True)
```
pankajkoti, line 163: Would this require bumping our minimum amazon provider version? I think the deferrable flag would have been added recently right?

astronomer/providers/amazon/aws/hooks/base_aws.py
pankajastro: It was https://org-airflow-team-org.astronomer.run/dfi0rona/connection/list/

astronomer/providers/amazon/aws/hooks/base_aws.py
pankajkoti: Not sure if it has all the needed pieces, I just extracted custom pieces and tailored in that PR to add support for refreshable credentials for only the `assume_role` method

pankajastro: I think we should do a similar for other methods too which have been not covered

pankajastro:
changing this to draft to avoid accidental merge

PR #1222, opened 2023-06-29
fix triggering failure due to astro_cloud subdomain change

Lee-W:
* ci(github-actions):
    * **fix** fix triggering failure due to astro_cloud subdomain change 
    * **refactor**: extract common wait and trigger logic as a reusable workflow

For this workflow to work, we must set the latest subdomain to github secret as `ASTRO_SUBDOMAIN`. we used to hardcode it in the script, but I think it might be better for use to keep it secret


pankajkoti (approved):
Should we add a default value for now as we might now have access to secrets? And we can keep changing the default values in our branch

Lee-W:
That's something I'd like to discuss as well. I'd opt to keep such a thing secret. As we don't change this value often, I think that should be fine

Lee-W:
I'll temporary add the value for testing and drop that commit

PR #1229, opened 2023-07-06
ci(github-actions): increate the frequency of airflow rc testing issue check

Lee-W:
Today's announcement happen to be a bit later than our regular check. thus, @phanikumv suggests we can increase the check frequency


Lee-W:
For this round of testing, I've triggered the action manually

pankajkoti (approved):

phanikumv (approved):

PR #1231, opened 2023-07-06
ci(github-actions): fix rc testing branch existence check

Lee-W:
the original logic cannot correctly load bash variables


pankajkoti (approved):
I was wondering something could be wrong when it created a duplicate PR. I think now we verified by running the job manually that it does not create a duplicate PR, right?

Looks good to me then!

PR #1233, opened 2023-07-08
Adding region option in eksctl commands

vatsrahul1001:
It appears that the default region support in the latest release of eksctl (v0.148.0) is broken. A defect has already been raised for this issue, which you can find [here](https://github.com/weaveworks/eksctl/issues/6774). As a result of this problem, all of our eksctl commands were failing because we didn't have the "--region" command line option included. To work around this issue, I have temporarily added the "--region" option explicitly


pankajkoti (approved):
Nice work @vatsrahul1001! 👏🏽 

Let's merge this after a successful DAG run on Astro Cloud.

PR #1240, opened 2023-07-12
Adding logic to poll deployment health instead of static wait

vatsrahul1001:
Currently, we have a static sleep time of [1800](https://github.com/astronomer/astronomer-providers/blob/d50a8b6150b541ca8cdc9c826a3e7c3bda7a00d3/.github/workflows/trigger-dag-reuse-wf.yaml#L45) seconds for the deployment to reach a healthy state. However, instead of relying on this fixed duration, we can implement a polling mechanism to check the health of the deployment. If the deployment is found to be unhealthy, we can Exit 1. This improvement will address the following issues:

1. We observed today that even when the deployment was unhealthy, our workflow still succeeded. This PR will prevent such occurrences in the future.
2. By implementing polling, we can reduce the overall execution time of the workflow. In most cases, deployments become healthy within 5-10 minutes, and with polling, we can optimize this process.

### DIFF
--- a/.github/workflows/deploy-integration-tests.yaml
+++ b/.github/workflows/deploy-integration-tests.yaml
@@ -57,6 +57,8 @@ jobs:
       deployment_id: ${{ secrets.PROVIDER_INTEGRATION_TESTS_DEPLOYMENT_ID }}
       astronomer_key_id: ${{ secrets.PROVIDER_INTEGRATION_TESTS_ASTRONOMER_KEY_ID }}
       astronomer_key_secret: ${{ secrets.PROVIDER_INTEGRATION_TESTS_ASTRONOMER_KEY_SECRET }}
+      organization_id: ${{ secrets.ORGANIZATION_ID }}
+      bearer_token: ${{ secrets.BEARER_TOKEN }
 
   deploy-to-providers-integration-tests-on-KE:
     if: |
--- a/.github/workflows/deploy-integration-tests.yaml
+++ b/.github/workflows/deploy-integration-tests.yaml
@@ -58,7 +58,7 @@ jobs:
       astronomer_key_id: ${{ secrets.PROVIDER_INTEGRATION_TESTS_ASTRONOMER_KEY_ID }}
       astronomer_key_secret: ${{ secrets.PROVIDER_INTEGRATION_TESTS_ASTRONOMER_KEY_SECRET }}
       organization_id: ${{ secrets.ORGANIZATION_ID }}
-      bearer_token: ${{ secrets.BEARER_TOKEN }
+      bearer_token: ${{ secrets.BEARER_TOKEN }}
 
   deploy-to-providers-integration-tests-on-KE:
     if: |
@@ -91,4 +91,4 @@ jobs:
       astronomer_key_id: ${{ secrets.PROVIDER_INTEGRATION_TESTS_ON_KE_ASTRONOMER_KEY_ID }}
       astronomer_key_secret: ${{ secrets. PROVIDER_INTEGRATION_TESTS_ON_KE_ASTRONOMER_KEY_SECRET }}
       organization_id: ${{ secrets.ORGANIZATION_ID }}
-      bearer_token: ${{ secrets.BEARER_TOKEN }
+      bearer_token: ${{ secrets.BEARER_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-integration-tests.yaml
```diff
@@ -57,6 +57,8 @@ jobs:
       deployment_id: ${{ secrets.PROVIDER_INTEGRATION_TESTS_DEPLOYMENT_ID }}
       astronomer_key_id: ${{ secrets.PROVIDER_INTEGRATION_TESTS_ASTRONOMER_KEY_ID }}
       astronomer_key_secret: ${{ secrets.PROVIDER_INTEGRATION_TESTS_ASTRONOMER_KEY_SECRET }}
+      organization_id: ${{ secrets.ORGANIZATION_ID }}
+      bearer_token: ${{ secrets.BEARER_TOKEN }
```
pankajkoti, line 61: We need to pass this for KE deployment too

vatsrahul1001: thanks, @pankajkoti added

Lee-W: I though we could get it through this https://github.com/astronomer/astronomer-providers/blob/main/dev/integration_test_scripts/trigger_dag.py#L83?

.github/workflows/trigger-dag-reuse-wf.yaml
```diff
@@ -9,11 +9,6 @@ on:  # yamllint disable-line rule:truthy
         required: false
         type: string
         default: ''
-      wait_time:
```
Lee-W, line 12: I thought we still need this wait time for the pods to be ready. Or does  the `HEALTHY` state indicate that the whole deployment is ready?

.github/workflows/trigger-dag-reuse-wf.yaml
```diff
@@ -34,17 +29,45 @@ on:  # yamllint disable-line rule:truthy
       astronomer_key_secret:
         description: 'astro cloud astronomer_key_secret'
         required: true
+      organization_id:
+        description: 'astro cloud organization_id'
+        required: true
+      bearer_token:
```
pankajkoti, line 35: Do we know if this Bearer token has an expiry and if yes, how frequently we need to update this?

.github/workflows/deploy-integration-tests.yaml
vatsrahul1001: @Lee-W i have created a workspace token that will always be the same and we can use that as well instead of making a new call every time

.github/workflows/trigger-dag-reuse-wf.yaml
vatsrahul1001: @pankajkoti i have set Expiry to None for now

.github/workflows/trigger-dag-reuse-wf.yaml
vatsrahul1001: @Lee-W this core API endpoint check pods health only and then only status is returned as HEALTHY

.github/workflows/trigger-dag-reuse-wf.yaml
```diff
@@ -45,30 +45,25 @@ jobs:
         run: |
           astro_core_api="https://api.astronomer.io/v1alpha1/organizations/${{secrets.organization_id }}/\
           deployments"
-          tries=60
+          tries=15
```
pankajkoti, line 48: It might make sense to wait for an hour maybe in some scenarios where it might take longer. So tries=30?

.github/workflows/trigger-dag-reuse-wf.yaml
```diff
@@ -34,17 +29,42 @@ on:  # yamllint disable-line rule:truthy
...
-          echo "Sleeping for ${{ inputs.wait_time }}"
-          echo "allowing the deployed image to be updated across all Airflow components."
-          sleep ${{ inputs.wait_time }}
+          astro_core_api="https://api.astronomer.io/v1alpha1/organizations/${{secrets.organization_id }}/\
+          deployments"
+          tries=15
+          health_flag=false
+
+          while [[ $tries -gt 0 && $health_flag == false ]]; do
+              sleep 120
```
pankajkoti, line 52: I think would be good to place this log at the end of while loop, just to see logs saying Deployment status is different than Health and it's waiting for deployment to be healthy in case if it becomes healthy in <2 mins.

pankajkoti (approved):

PR #1243, opened 2023-07-13
Add actionlint

Lee-W:
Failures due to https://github.com/astronomer/astronomer-providers/pull/1240 changes (fixed in https://github.com/astronomer/astronomer-providers/pull/1242) can be detected through actionlint. Thus, adding this lint for preventing this kind of error in the future

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,7 +92,7 @@ jobs:
       - run:
           name: Run pre-commit
           command: |
-            pip install pre-commit
+            pip install "pre-commit>=3.0.0"
             pre-commit run markdown-link-check --all-files || { git --no-pager diff && false ; }
       - save_cache:
           key: v1-pc-cache-{{ checksum "pre-commit-cache-key.txt" }}-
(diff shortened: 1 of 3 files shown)


phanikumv (approved):

pankajkoti (commented):
I did not understand exactly on what are we checking with actionlint here. Can you please help me understand?

Lee-W:
> I did not understand exactly on what are we checking with actionlint here. Can you please help me understand?

yamls inside `.github/workflows`

pankajkoti:
> > I did not understand exactly on what are we checking with actionlint here. Can you please help me understand?
> 
> yamls inside `.github/workflows`

and what do we check for exactly in those yamls?

pankajkoti:
Is it that we have the template yaml that we use and it will report for missing secrets that need to be passed from other yamls that use the template YAML?

Lee-W:
> Is it that we have the template yaml that we use and it will report for missing secrets that need to be passed from other yamls that use the template YAML?

Including but not limited to this check, the check you mentioned is https://github.com/rhysd/actionlint/blob/main/docs/checks.md#check-inputs-and-secrets-in-workflow-call, but it can also check a bunch of potential failures

pankajkoti (approved):

Issue #1259, opened 2023-07-25
`snowflake` not included in the `extras_require` of `amazon` extra


ReadytoRocc:
**Describe the bug**
The `RedshiftDataOperatorAsync` utilizes the `RedshiftDataHook`, which has a [dependency](https://github.com/astronomer/astronomer-providers/blob/1.17.1/astronomer/providers/amazon/aws/hooks/redshift_data.py#L12) on the `snowflake` python package. This is [not included](https://github.com/astronomer/astronomer-providers/blob/1.17.1/setup.cfg#L46-L48) in the `extras_require` of the `amazon` extra. This results in an import failure.

**To Reproduce**
- Install `astronomer-providers==1.17.1`
- Include the `amazon` extra.
- Import `RedshiftDataOperatorAsync`:
```
>>> from astronomer.providers.amazon.aws.operators.redshift_data import RedshiftDataOperatorAsync
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.10/site-packages/astronomer/providers/amazon/aws/operators/redshift_data.py", line 6, in <module>
    from astronomer.providers.amazon.aws.hooks.redshift_data import RedshiftDataHook
  File "/usr/local/lib/python3.10/site-packages/astronomer/providers/amazon/aws/hooks/redshift_data.py", line 12, in <module>
    from snowflake.connector.util_text import split_statements
ModuleNotFoundError: No module named 'snowflake'
>>> 
```

pankajkoti:
Instead of adding the extra, I worked on removing the snowflake dependency as suggested in Slack thread https://astronomer.slack.com/archives/C0387PMM2E7/p1690293653544339?thread_ts=1690244271.090079&cid=C0387PMM2E7 

Have created a PR for the same #1260

PR #1260, opened 2023-07-26
Use DbApiHook's split_sql_string instead of snowflake utility in amazon provider

pankajkoti:
Currently, the redshift_data hook in the Amazon provider has 
a dependency on snowflake for it's utility to split sql strings. 
The PR removes this need by instead using the DbApiHook's 
`split_sql_string` which reliably parses the sql strings for us.

closes: #1259

### DIFF
--- a/astronomer/providers/amazon/aws/hooks/redshift_data.py
+++ b/astronomer/providers/amazon/aws/hooks/redshift_data.py
@@ -1,15 +1,14 @@
 from __future__ import annotations
 
 import asyncio
-from io import StringIO
 from typing import Any, Iterable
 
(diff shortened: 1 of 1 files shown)


pankajkoti:
Verified that our Redshift SQL [example DAG](https://github.com/astronomer/astronomer-providers/blob/main/astronomer/providers/amazon/aws/example_dags/example_redshift_sql.py) which passes sql strings in various formats runs well.


Screenshot 2023-07-26 at 10 11 03 AM

pankajastro (approved):

Lee-W (approved):

PR #1266, opened 2023-07-28
Pin OL to 0.29.2

pankajkoti:


pankajkoti:
This was to test if OL 0.29.2 does not cause issues for the Snowflake DAG on Astro. And verified that it does not after deploying this pinned version. Closing the PR.

PR #1268, opened 2023-07-30
Execute Airflow Providers RC tests with a GCP cluster deployment

vatsrahul1001:
As per the current process of providers' release testing, our workflow only updates the AWS deployment with RC Providers versions for testing. However, we want this update to be executed across all cloud providers. Currently, we are using a hosted organization, and Azure is not supported. We will enable this feature for Azure when it becomes available.


pankajkoti (commented):
Another thought I had is we currently test this RCs only on CeleryExecutor deployments. Do you think we need to also test these on KubernetesExecutor deployments?

pankajastro:
> Another thought I had is we currently test this RCs only on CeleryExecutor deployments. Do you think we need to also test these on KubernetesExecutor deployments?

IMO makes sense for Airflow rc but not sure about providers

vatsrahul1001:
> Another thought I had is we currently test this RCs only on CeleryExecutor deployments. Do you think we need to also test these on KubernetesExecutor deployments?

Correct Pankaj, we can do that, should I add that to this PR only?

pankajkoti:
> > Another thought I had is we currently test this RCs only on CeleryExecutor deployments. Do you think we need to also test these on KubernetesExecutor deployments?
> 
> Correct Pankaj, we can do that, should I add that to this PR only?

maybe we can create another PR later since we have mixed opinions.

Lee-W:
Do we have the secrets set up ready? Could we test it before merging it? Thanks!

vatsrahul1001:
> Do we have the secrets set up ready? Could we test it before merging it? Thanks!

@Lee-W I have asked @phanikumv to add secrets, sure will test post that

Lee-W (approved):

phanikumv:
> > Do we have the secrets set up ready? Could we test it before merging it? Thanks!
> 
> @Lee-W I have asked @phanikumv to add secrets, sure will test post that

secrets have been added

PR #1274, opened 2023-08-01
Fix impersonation failure on GKE start pod operator async

Lee-W:
**What's the issue?**

The argument `impersonation_chain` does not work. The operator still uses the gcp_conn to start the pod.

**What's changed**
We suspect the issue is due to that the token is not correctly generated. Thus, we try to write this token into kube_config


Lee-W:
圖片

It works fine on my side. But would like to mark it as "Ready for review" after testing by CRE.

phanikumv (commented):
can you add more context on why this code change is necessary in the description?

Lee-W:
Screenshot 2023-08-07 at 2 45 46 PM

astronomer/providers/google/cloud/gke_utils.py
```diff
@@ -0,0 +1,154 @@
...
+        for context in config_content["contexts"]:
+            if context["name"] == cluster_context:
+                user_name = context["context"]["user"]
+                break
+
+        user_index = 0
+        for u_index, user in enumerate(config_content["users"]):
+            if user["name"] == user_name:
+                user_index = u_index
+                break
```
pankajkoti, line 60: Is the user guaranteed to be present in the config_content?

If yes, then I don't think we need another user_index var, we could just use u_index. And if no, then user_index points to 0 which is wrong.

astronomer/providers/google/cloud/gke_utils.py
```diff
@@ -0,0 +1,154 @@
...
+        client = ClusterManagerClient(credentials=impersonated_creds)
+        name = f"projects/{project_id}/locations/{location}/clusters/{cluster_name}"
+        client.get_cluster(name=name)
+    except Exception as e:
+        raise Exception(f"Error while creating impersonated creds: {e}") from e
+    return impersonated_creds.token  # type: ignore[no-any-return]
+
+
+def _write_token_into_config(config_name: str, token: str, cluster_context: str | None = None) -> None:
+    # breakpoint()
```
pankajkoti, line 42: We can remove this.

astronomer/providers/google/cloud/gke_utils.py
```diff
@@ -0,0 +1,154 @@
...
+        target_principal=impersonation_account,
+        target_scopes=["https://www.googleapis.com/auth/cloud-platform"],
+    )
+    try:
+        client = ClusterManagerClient(credentials=impersonated_creds)
+        name = f"projects/{project_id}/locations/{location}/clusters/{cluster_name}"
+        client.get_cluster(name=name)
+    except Exception as e:
+        raise Exception(f"Error while creating impersonated creds: {e}") from e
+    return impersonated_creds.token  # type: ignore[no-any-return]
```
pankajkoti, line 38: Can we cast this to str and remove the type ignore?

astronomer/providers/google/cloud/gke_utils.py
```diff
@@ -0,0 +1,154 @@
...
+                break
+
+        # update the token
+        config_content["users"][user_index]["user"]["auth-provider"] = {
+            "name": "gcp",
+            "config": {"access-token": token},
+        }
+    except KeyError as e:
+        logging.exception(e)
+        return
```
pankajkoti, line 69: Do we need to reraise the exception here?

dev/Dockerfile.google_cloud
```diff
@@ -13,7 +13,7 @@ RUN apt-get install -y --no-install-recommends \
         echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] http://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list && \
         curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg  add - && \
         apt-get update -y && \
-        apt-get install google-cloud-sdk -y && \
+        apt-get install python3 google-cloud-sdk -y && \
```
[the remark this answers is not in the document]
Lee-W: Not sure about the base image, but without this. It doesn't work on my local machine

astronomer/providers/google/cloud/gke_utils.py
Lee-W: Thanks! Just updated.

astronomer/providers/google/cloud/gke_utils.py
Lee-W: Thanks! Just updated.

astronomer/providers/google/cloud/gke_utils.py
Lee-W: let me check it again 👀

Lee-W: Yep, `u_index` should be sufficient. But after a second thought, I tried to rewrite it with `KubeConfigLoader` in `kubernetes_asyncio.config.kube_config` to make use of their more robust loader

astronomer/providers/google/cloud/gke_utils.py
Lee-W: This is removed as I rewrite this section with `kubernetes_asyncio.config.kube_config.KubeConfigLoader`

pankajastro (approved):

Issue #1170, opened 2023-06-05
Handle soft fail for sensors


pankajastro:
currently, we always fail in execute_comeplet if trigger a yield error which not respect  the soft_fail sensors param.

phanikumv:
@utkarsharma2 lets make progress on this

infohash:
Is this the reason why 'S3KeySensorAsync` goes to failed state instead of skipped state when `airflow.exceptions.TaskDeferralError: Trigger/execution timeout` occurs and `soft_fail` is set to `True`? I encountered this recently and I assumed it that it is by design to alert the user about missing bucket keys. But I agree, it should go to skipped state if soft_fail is set to `True`.

infohash:
@pankajastro I tested soft fail for `S3KeySensorAsync` and it is not respecting `soft_fail` when the task execution times out and `airflow.exceptions.TaskDeferralError: Trigger/execution timeout` is raised. I tested it with both `wildcard_match` and `use_regex` flag.

infohash:
I have tested it in airflow 2.7.1. It has been fixed in upstream by apache/airflow#33196 merged in 2.7.1 release.

pankajastro:
Thanks for letting us know!

PR #1276, opened 2023-08-02
Handle soft fail

pankajastro:
closes: https://github.com/astronomer/astronomer-providers/issues/1170
Handle soft_fail param for sensor

TODO
- [x] run test on astro

### DIFF
--- a/astronomer/providers/apache/hive/sensors/hive_partition.py
+++ b/astronomer/providers/apache/hive/sensors/hive_partition.py
@@ -1,13 +1,14 @@
-from datetime import timedelta
-from typing import Dict, Optional
+from __future__ import annotations
 
-from airflow.exceptions import AirflowException
+from datetime import timedelta
+
 from airflow.providers.apache.hive.sensors.hive_partition import HivePartitionSensor
 
 from astronomer.providers.apache.hive.hooks.hive import HiveCliHookAsync
 from astronomer.providers.apache.hive.triggers.hive_partition import (
     HivePartitionTrigger,
 )
+from astronomer.providers.utils.sensor_util import handle_error
 from astronomer.providers.utils.typing_compat import Context
 
 
@@ -57,7 +58,7 @@ class HivePartitionSensorAsync(HivePartitionSensor):
                 method_name="execute_complete",
             )
 
-    def execute_complete(self, context: Context, event: Optional[Dict[str, str]] = None) -> str:
+    def execute_complete(self, context: Context, event: dict[str, str] | None = None) -> str:  # type: ignore[return]
         """
         Callback for when the trigger fires - returns immediately.
         Relies on trigger to throw an exception, otherwise it assumes execution was
@@ -69,5 +70,4 @@ class HivePartitionSensorAsync(HivePartitionSensor):
                     "Success criteria met. Found partition %s in table: %s", self.partition, self.table
                 )
                 return event["message"]
-            raise AirflowException(event["message"])
-        raise AirflowException("No event received in trigger callback")
+            handle_error(self.soft_fail, event["message"])
--- a/astronomer/providers/apache/hive/sensors/hive_partition.py
+++ b/astronomer/providers/apache/hive/sensors/hive_partition.py
@@ -8,7 +8,7 @@ from astronomer.providers.apache.hive.hooks.hive import HiveCliHookAsync
 from astronomer.providers.apache.hive.triggers.hive_partition import (
     HivePartitionTrigger,
 )
-from astronomer.providers.utils.sensor_util import handle_error
+from astronomer.providers.utils.sensor_util import raise_error_or_skip_exception
 from astronomer.providers.utils.typing_compat import Context
 
 
@@ -70,4 +70,4 @@ class HivePartitionSensorAsync(HivePartitionSensor):
                     "Success criteria met. Found partition %s in table: %s", self.partition, self.table
                 )
                 return event["message"]
-            handle_error(self.soft_fail, event["message"])
+            raise_error_or_skip_exception(self.soft_fail, event["message"])
--- a/astronomer/providers/sftp/sensors/sftp.py
+++ b/astronomer/providers/sftp/sensors/sftp.py
@@ -6,7 +6,7 @@ from airflow.providers.sftp.sensors.sftp import SFTPSensor
 
 from astronomer.providers.sftp.hooks.sftp import SFTPHookAsync
 from astronomer.providers.sftp.triggers.sftp import SFTPTrigger
-from astronomer.providers.utils.sensor_util import handle_error
+from astronomer.providers.utils.sensor_util import raise_error_or_skip_exception
 from astronomer.providers.utils.typing_compat import Context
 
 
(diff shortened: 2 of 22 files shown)


phanikumv (commented):
LGTM,please test it on our deployments just to rule out any regression

astronomer/providers/sftp/sensors/sftp.py
```diff
@@ -79,11 +79,7 @@ def execute_complete(self, context: Dict[str, Any], event: Any = None) -> None:
...
             if "status" in event and event["status"] == "error":
-                raise AirflowException(event["message"])
-
+                handle_error(self.soft_fail, event["message"])
             if "status" in event and event["status"] == "success":
                 self.log.info("%s completed successfully.", self.task_id)
                 self.log.info(event["message"])
-                return None
-
-        raise AirflowException("No event received in trigger callback")
```
Lee-W, line 89: Should we handle the case of missing events?

astronomer/providers/snowflake/sensors/snowflake.py
```diff
@@ -90,20 +90,13 @@ def execute(self, context: Context) -> None:
                 method_name=self.execute_complete.__name__,
             )
 
-    def execute_complete(
-        self,
-        context: Context,
-        event: dict[str, str | list[str]] | None = None,
-    ) -> Any:
+    def execute_complete(self, context: Context, event: dict[str, Any] | None = None) -> Any:
```
Lee-W, line 93: Why do we change the type annotation from `Any` to `list[str]`? Is it no longer accurate?

astronomer/providers/snowflake/sensors/snowflake.py
```diff
@@ -90,20 +90,13 @@ def execute(self, context: Context) -> None:
...
         successful.
         """
         if event:
             if "status" in event and event["status"] == "error":
-                raise AirflowException(event["message"])
-
+                handle_error(self.soft_fail, event["message"])
             self.log.info(event["message"])
-        else:
-            self.log.info("%s completed successfully.", self.task_id)
```
Lee-W, line 109: By removing these lines, our docstring seems to be inaccurate. Should we keep these lines? Or update docstring

astronomer/providers/utils/sensor_util.py
```diff
@@ -0,0 +1,26 @@
...
+
+from typing import TYPE_CHECKING, Any
+
+from airflow.exceptions import AirflowException, AirflowSkipException
+
+if TYPE_CHECKING:  # pragma: no cover
+    from astronomer.providers.utils.typing_compat import Context
+
+
+def poke(cls: Any, context: Context) -> bool | None:  # pragma: no cover
```
Lee-W, line 11: Also, I'm wondering why we need this `poke` function. In what scenario will `poke` raise an exception?

astronomer/providers/utils/sensor_util.py
```diff
@@ -0,0 +1,26 @@
...
+    try:
+        return cls.poke(context)  # type: ignore[no-any-return]
+    except Exception as e:
+        if cls.soft_fail:
+            raise AirflowSkipException(str(e))
+        else:
+            raise e
+
+
+def handle_error(soft_fail: bool, error_message: str) -> None:  # pragma: no cover
```
Lee-W, line 22: Would it be possible for us to rename this function? `handle_error` seems to be an broad name.

Lee-W (commented):
Should we add test cases to `soft_fail`?

astronomer/providers/apache/hive/sensors/hive_partition.py
```diff
@@ -69,5 +70,4 @@ def execute_complete(self, context: Context, event: Optional[Dict[str, str]] = N
                     "Success criteria met. Found partition %s in table: %s", self.partition, self.table
                 )
                 return event["message"]
-            raise AirflowException(event["message"])
-        raise AirflowException("No event received in trigger callback")
+            handle_error(self.soft_fail, event["message"])
```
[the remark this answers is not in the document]
pankajastro: I thought of removing this because 1) it is not consistent in the repo i.e some places we raise errors and someplace we mark success 2) an empty event from trigger should not be possible until we yield empty

astronomer/providers/utils/sensor_util.py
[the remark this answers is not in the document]
pankajastro: I think mypy was not happy but will retry

astronomer/providers/snowflake/sensors/snowflake.py
pankajastro: I took a quick look and look like it yield 
```
yield TriggerEvent(
                            {
                                "status": "success",
                                "message": "Found expected markers.",
                            }
                        )
```
so was not correct?

astronomer/providers/utils/sensor_util.py
pankajastro: This is a wrapper to call the upstream poke method. We call poke method in almost all sensors before deferring so we need to handle the error if any occurs. This  poke method helps to avoid writing try-catch at  multiple places

pankajastro: > In what scenario will poke raise an exception?

maybe a call to s3 and call fail either network or something else

astronomer/providers/utils/sensor_util.py
pankajastro: I agree would you like to suggest something

pankajastro:
> Should we add test cases to `soft_fail`?

Added some unit tests

astronomer/providers/snowflake/sensors/snowflake.py
Lee-W: got it. then should it be `dict[str, str]` instead?

astronomer/providers/utils/sensor_util.py
Lee-W: Got it. Sounds good 👍

astronomer/providers/utils/sensor_util.py
Lee-W: `raise_error_or_skip_exception`? No better idea as of now 😞

astronomer/providers/snowflake/sensors/snowflake.py
pankajastro: This is just a log statement and will no event on end task status but I have restored it.

astronomer/providers/utils/sensor_util.py
```diff
@@ -0,0 +1,27 @@
...
+from typing import TYPE_CHECKING
+
+from airflow.exceptions import AirflowException, AirflowSkipException
+from airflow.sensors.base import BaseSensorOperator
+
+if TYPE_CHECKING:  # pragma: no cover
+    from astronomer.providers.utils.typing_compat import Context
+
+
+def poke(cls: BaseSensorOperator, context: Context) -> bool:  # pragma: no cover
```
Lee-W, line 12: May I know why we mark them as no cover?

astronomer/providers/utils/sensor_util.py
```diff
@@ -0,0 +1,27 @@
...
+    try:
+        return cls.poke(context)  # type: ignore[return-value]
+    except Exception as e:
+        if cls.soft_fail:
+            raise AirflowSkipException(str(e))
+        else:
+            raise e
+
+
+def raise_error_or_skip_exception(soft_fail: bool, error_message: str) -> None:  # pragma: no cover
```
Lee-W, line 23: May I know why we mark them as no cover?

pankajastro: I thought this not cover by the test but we have a test for this. I have removed it.

astronomer/providers/utils/sensor_util.py
pankajastro: tests are missing for a few lines. I have reduced the scope.

Lee-W (approved):

PR #1279, opened 2023-08-03
Clean snowflake query status check

pankajastro:


astronomer/providers/snowflake/operators/snowflake.py
```diff
@@ -43,25 +41,10 @@ def _check_queries_finish(conn: SnowflakeConnection, query_ids: list[str]) -> bo
...
-                    "DISCONNECTED: The session’s connection is broken. "
-                    "The query’s state will change to “FAILED_WITH_ERROR” soon."
-                )
-            elif status == QueryStatus.SUCCESS:
-                pass
-            elif status in (
+            status = conn.get_query_status_throw_if_error(query_id)
+            if status == QueryStatus.SUCCESS:
+                continue
+            if status in (
```
Lee-W, line 47:
```suggestion
            if status == QueryStatus.SUCCESS:
                continue
            elif status in (
```

astronomer/providers/snowflake/operators/snowflake.py
```diff
@@ -71,9 +54,6 @@ def _check_queries_finish(conn: SnowflakeConnection, query_ids: list[str]) -> bo
                 QueryStatus.FAILED_WITH_INCIDENT,
             ):
                 return False
-            else:
-                raise ValueError(f"Unexpected QueryStatus: {str(status)}")
```
Lee-W, line 75: May I know why should we remove these lines?

tests/snowflake/operators/test_snowflake.py
```diff
@@ -32,49 +32,13 @@
...
-    "mock_status",
-    (
-        QueryStatus.RUNNING,
-        QueryStatus.QUEUED,
-        QueryStatus.RESUMING_WAREHOUSE,
-        QueryStatus.BLOCKED,
-        QueryStatus.NO_DATA,
-        QueryStatus.FAILED_WITH_INCIDENT,
-    ),
+    "query_status,expected_response", [(QueryStatus.SUCCESS, True), (QueryStatus.RUNNING, False)]
```
Lee-W, line 36: I think we should still check other status also return False

astronomer/providers/snowflake/operators/snowflake.py
pankajastro: I'm expecting the query state will be either fail, succeed or running in case of fail/error `get_query_status_throw_if_error` will throw an error and the task will fail and succeed and running case we are handling. About this block, I'm not sure if execution reaches here

Lee-W: I think the idea of having this block is to avoid silently passing API changes (e.g., what happen if we change the name or add a new status)

pankajastro: I think the snowflake connector provides `is_still_running` method and we should use this method to avoid issues like a new status introduced, wdyt?

Lee-W: Sounds good

tests/snowflake/operators/test_snowflake.py
pankajastro: Since we do not handle error cases here so not sure if a test requires. quick exploration says they have some tests in https://github.com/snowflakedb/snowflake-connector-python/blob/9b586d4c37f5206742f418ae1d66fc83248cacbe/test/unit/test_connection.py

Lee-W: It would still be better if we could verify those value returns `False`. But I can go with the current implementation as well

Lee-W (approved):

PR #1280, opened 2023-08-04
Fix automate process

Lee-W:
* fix release commit validation step failure
    * we'll need to quote the commit message when running the bash script
* change the regex to check the release commit to include the PR number
    * Our release message format is something like "Release 1.0.0 (#123)" instead of "Release 1.0.0" 
* use BOT_ACCESS_TOKEN to push the git tag
* remove bump dev version step for now as it will fail due to commitizen behavior

### DIFF
--- a/.github/scripts/validate_release_commit.py
+++ b/.github/scripts/validate_release_commit.py
@@ -2,7 +2,7 @@ import re
 import sys
 
 if __name__ == "__main__":
-    pattern = r"^Release \d{1,}\.\d{1,}\.\d{1,}$"
+    pattern = r"^Release \d{1,}\.\d{1,}\.\d{1,}( \(#\d+\)){0,1}$"
     if len(sys.argv) == 2:
         commit_message = sys.argv[1]
         if not re.fullmatch(pattern, commit_message):
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -14,7 +14,7 @@ jobs:
       - name: Check out
         uses: actions/checkout@v3
         with:
-          token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ secrets.BOT_ACCESS_TOKEN }}
 
       - name: Check head commit message
         id: check_head_commit_message
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -21,7 +21,7 @@ jobs:
         continue-on-error: true
         run: |
           head_commit_message=`git log --format=%s -1`
-          python ./.github/scripts/validate_release_commit.py $head_commit_message
+          python ./.github/scripts/validate_release_commit.py "$head_commit_message"
           if [ $? -eq 1 ]
           then
             echo "valid_release_message=False" >> $GITHUB_OUTPUT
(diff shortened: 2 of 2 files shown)


.github/scripts/validate_release_commit.py
```diff
@@ -2,7 +2,7 @@
 import sys
 
 if __name__ == "__main__":
-    pattern = r"^Release \d{1,}\.\d{1,}\.\d{1,}$"
+    pattern = r"^Release \d{1,}\.\d{1,}\.\d{1,}( \(#\d+\)){0,1}$"
```
pankajkoti, line 5: Can you please elaborate this? I did not understand this part. 

Looking at the PR description, it also has non-digit part showing in the description, but here we have only digits.

I am checking the previous release PR. Do we mean to say it is the PR number?

`Release 1.17.2 #1262`

If yes, it does not seem to have closing round brackets around the PR number hash. Do we need those brackets?

Screenshot 2023-08-07 at 12 33 45 PM

pankajkoti: Screenshot 2023-08-07 at 12 37 02 PM
 This is how the PR description reads and it contains non-digit characters too.

sunank200: +1

.github/workflows/release.yaml
```diff
@@ -14,14 +14,14 @@ jobs:
...
-          token: ${{ secrets.GITHUB_TOKEN }}
+          token: ${{ secrets.BOT_ACCESS_TOKEN }}
 
       - name: Check head commit message
         id: check_head_commit_message
         continue-on-error: true
         run: |
           head_commit_message=`git log --format=%s -1`
-          python ./.github/scripts/validate_release_commit.py $head_commit_message
+          python ./.github/scripts/validate_release_commit.py "$head_commit_message"
```
sunank200, line 24: why do we need the quotes?

.github/scripts/validate_release_commit.py
Lee-W: Yes, I think you're right. Let me fix it now

.github/workflows/release.yaml
Lee-W: Yes, I think this is one of the main reason why it fails last time. It seems to be passed as `python ./.github/scripts/validate_release_commit.py Release 1.17.2 #123123` which causes an error

.github/scripts/validate_release_commit.py
Lee-W: Just pushed the fix

pankajkoti (approved):

.github/scripts/validate_release_commit.py
Lee-W: @pankajkoti @sunank200  I just checked again. We actually need those brackets. We're checking the commit message. not the PR title
圖片

pankajkoti: okay @Lee-W

We can merge and test with what you feel works :)

Lee-W: Thanks! Let me merge it now.

PR #1285, opened 2023-08-06
Fix Python3.11 constraints

pankajastro:
job: https://app.circleci.com/pipelines/github/astronomer/astronomer-providers/6270/workflows/9e4ee250-9b09-4512-a7a7-bb911e3b2cf9

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,8 +37,15 @@ workflows:
           name: constraints-<< matrix.python_version >>-<< matrix.airflow_version >>
           matrix:
             parameters:
-              python_version: ["3.8", "3.9", "3.10", "3.11"]
+              python_version: ["3.8", "3.9", "3.10"]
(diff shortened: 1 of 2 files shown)


pankajkoti (commented):
I think with this we need to mention somewhere in our public docs that we don't support Hive provider with Python 3.11

pankajastro:
> I think with this we need to mention somewhere in our public docs that we don't support Hive provider with Python 3.11

yes, good point. shall we add a note in readme?

pankajastro:
> I think we could add this Hive provider limitation in our public doc limitations
> 
> https://astronomer-providers.readthedocs.io/en/stable/#limitations

This is README only, I have added in limitations section also and kept the note too because I feel that would be more visible

pankajkoti (approved):
Looks good to me after checks passing.

pankajastro:
merging this one since currently main is broken but feel free to suggest any improvement

PR #1294, opened 2023-08-10
Fix style, test changes due to https://github.com/apache/airflow/issues/33257

Lee-W:
* fix mypy warning
* remove redundant mocking which cause test failed

https://github.com/astronomer/astronomer-providers/pull/1293


pankajastro (approved):

sunank200 (approved):

pankajkoti:
Mypy might fail on main due to merging this PR, but it will work fine after the google provider RC3 is released.

PR #1295, opened 2023-08-11
Use personal access token in rc workflow and extract it as a separate reusable workflow

Lee-W:
* extract check rc announcement as a reusable workflow
* use BOT_ACCESS_TOKEN for test-rc-release
* replace the fake account action@github.com with airflow-oss-bot@astronomer.io
* rename Reusable workflows to separate them from other workflows
* fix typo


Lee-W:
圖片

pankajastro (approved):

PR #1296, opened 2023-08-11
Adding Python version in Master DAG slack report

vatsrahul1001:
Recently we started testing example dags with runtime Python version images and we observed a need to get Python version info in report

### DIFF
--- a/.circleci/integration-tests/master_dag.py
+++ b/.circleci/integration-tests/master_dag.py
@@ -70,6 +70,7 @@ def get_report(dag_run_ids: List[str], **context: Any) -> None:
             f"*{header}:* `{value}`\n"
             for header, value in [
                 ("Runtime version", os.getenv("ASTRONOMER_RUNTIME_VERSION", "N/A")),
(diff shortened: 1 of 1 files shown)


.circleci/integration-tests/master_dag.py
```diff
@@ -70,6 +70,7 @@ def get_report(dag_run_ids: List[str], **context: Any) -> None:
             f"*{header}:* `{value}`\n"
             for header, value in [
                 ("Runtime version", os.getenv("ASTRONOMER_RUNTIME_VERSION", "N/A")),
+                ("Python version", os.getenv("PYTHON_VERSION", "N/A")),
```
Lee-W, line 73: I'm good with this change, but may I know where is this `PYTHON_VERSION` set?

Lee-W (approved):

.circleci/integration-tests/master_dag.py
vatsrahul1001: @Lee-W its set in runtime image
image

PR #1302, opened 2023-08-11
Use env var for cache version in cache key for CircleCI jobs

pankajkoti:
Currently, main branch is using an old cache for CircleCI jobs. 
We already have a hash for setup.cfg in the key, but this does 
not allow us to use newer upstream dependencies when they 
are released as there is no change in setup.cfg due to no version 
pins. To allow bursting cache, leveraging the use of environment 
variables. Have added a CACHE_VERSION env variable in our 
project settings and set the value for it to today's date. We can 
update it as and when needed by changing the value to new date 
the next time we feel the need to burst the cache.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -121,8 +121,8 @@ jobs:
       - checkout
       - restore_cache:
           keys:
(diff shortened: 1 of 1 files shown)


pankajkoti:
This will allow us to burst cache for main branch which is currently using older dependencies and hence mypy is failing. Main is red at the moment.

Lee-W (approved):

PR #1304, opened 2023-08-14
Add support for Aio refreshable creds for assume role

pankajkoti:


pankajkoti:
Verified that credentials get refreshed by having added some temporary logs and tweaking around expiration.

Screenshot 2023-08-14 at 10 14 46 PM

pankajkoti:
will work next on tests

pankajkoti:
Cc: @abhishekbhakat

pankajastro (approved):
LGTM.
1. we should test the existing aws integration test on Astro once
2. what do you think about adding an integration test or using some existing one to use this method of authentication?

Lee-W (approved):

pankajkoti:
> LGTM.
> 
> 1. we should test the existing aws integration test on Astro once
> 2. what do you think about adding an integration test or using some existing one to use this method of authentication?

1. I tested existing AWS integration locally on runtime image without the role_arn i.e. existing AWS integration and it runs fine too.
2. Yes perhaps, we need to can use the same example DAG with a different connection. Will create a follow-up PR for this.

pankajkoti:
@abhishekbhakat would you like to test this PR?

PR #1306, opened 2023-08-15
Fix ExternalDeploymentTaskSensorAsync

abhishekbhakat:
2 Main points.

- ExternalDeploymentTaskSensorAsync does not need a poke. We can directly defer I think. Otherwise we'll need to make the parameter `request_check` mandatory.
- Using aiohttp seems unnecessary to me. A simple request can get the results. Creating aiohttp session get's closed by Astro.

### DIFF
--- a/astronomer/providers/core/sensors/external_task.py
+++ b/astronomer/providers/core/sensors/external_task.py
@@ -115,20 +115,19 @@ class ExternalDeploymentTaskSensorAsync(HttpSensorAsync):
 
     def execute(self, context: Context) -> None:
         """Defers trigger class to poll for state of the job run until it reaches a failure state or success state"""
-        if not self.poke(context=context):
-            self.defer(
-                timeout=self.execution_timeout,
-                trigger=ExternalDeploymentTaskTrigger(
-                    http_conn_id=self.http_conn_id,
-                    method=self.hook.method,
-                    endpoint=self.endpoint,
-                    data=self.request_params,
-                    headers=self.headers,
-                    extra_options=self.extra_options,
-                    poke_interval=self.poke_interval,
-                ),
-                method_name="execute_complete",
-            )
+        self.defer(
+            timeout=self.execution_timeout,
+            trigger=ExternalDeploymentTaskTrigger(
+                http_conn_id=self.http_conn_id,
+                method=self.hook.method,
+                endpoint=self.endpoint,
+                data=self.request_params,
+                headers=self.headers,
+                extra_options=self.extra_options,
+                poke_interval=self.poke_interval,
+            ),
+            method_name="execute_complete",
+        )
 
     def execute_complete(self, context: "Context", event: Optional[Dict[str, Any]] = None) -> Any:
         """
--- a/astronomer/providers/core/triggers/external_task.py
+++ b/astronomer/providers/core/triggers/external_task.py
@@ -12,6 +12,7 @@ from sqlalchemy import func
 from sqlalchemy.orm import Session
 
 from astronomer.providers.http.triggers.http import HttpTrigger
+from airflow.providers.http.hooks.http import HttpHook
 
 
 class TaskStateTrigger(BaseTrigger):
@@ -169,21 +170,24 @@ class ExternalDeploymentTaskTrigger(HttpTrigger):
 
     async def run(self) -> AsyncIterator["TriggerEvent"]:
         """
-        Makes a series of asynchronous http calls via an http hook poll for state of the job
+        Makes a series of http calls via an http hook poll for state of the job
         run until it reaches a failure state or success state. It yields a Trigger if response state is successful.
         """
         from airflow.utils.state import State
 
-        hook = self._get_async_hook()
+        hook = HttpHook(
+            method="GET", 
+            http_conn_id=self.http_conn_id
+            )
         while True:
             try:
-                response = await hook.run(
+                response = hook.run(
                     endpoint=self.endpoint,
                     data=self.data,
                     headers=self.headers,
                     extra_options=self.extra_options,
                 )
-                resp_json = await response.json()
+                resp_json = response.json()
                 if resp_json["state"] in State.finished:
                     yield TriggerEvent(resp_json)
                 else:


abhishekbhakat:
CC @pankajastro

astronomer/providers/core/sensors/external_task.py
```diff
@@ -115,20 +115,19 @@ class ExternalDeploymentTaskSensorAsync(HttpSensorAsync):
...
-                    method=self.hook.method,
-                    endpoint=self.endpoint,
-                    data=self.request_params,
-                    headers=self.headers,
-                    extra_options=self.extra_options,
-                    poke_interval=self.poke_interval,
-                ),
-                method_name="execute_complete",
-            )
+        self.defer(
```
pankajastro, line 118: I think we decided to check before deferring because we were hitting some issues with Kubernetes executer when a task was spending very small time on triggerer.  @Lee-W / @phanikumv what do you think about this change?

pankajastro:
> Using aiohttp seems unnecessary to me

I'm ok either way with or without aiohttp given that API call does not take significant time/

pankajastro:
Also, I feel since this sensor is part of the core sensor so we should avoid depending on `apache-airflow-providers-http`

abhishekbhakat:
> Also, I feel since this sensor is part of the core sensor so we should avoid depending on `apache-airflow-providers-http`

Any particular reason? We'll need to have a redundant code then, isn't it?

astronomer/providers/core/sensors/external_task.py
Lee-W: agree

astronomer/providers/core/triggers/external_task.py
```diff
@@ -169,21 +170,21 @@ def serialize(self) -> Tuple[str, Dict[str, Any]]:
...
     async def run(self) -> AsyncIterator["TriggerEvent"]:
         """
-        Makes a series of asynchronous http calls via an http hook poll for state of the job
+        Makes a series of http calls via an http hook poll for state of the job
         run until it reaches a failure state or success state. It yields a Trigger if response state is successful.
         """
         from airflow.utils.state import State
 
-        hook = self._get_async_hook()
+        hook = HttpHook(method="GET", http_conn_id=self.http_conn_id)
```
Lee-W, line 178: If we're not running it in an async way, is there any reason we need to defer it to the trigger?

pankajastro:
> Any particular reason? We'll need to have a redundant code then, isn't it?

I have not tested it but I think if you install this package like `pip install 'astronomer-providers[amazon]'` then http provider might not be available but the core will be and it would cause an issue

astronomer/providers/core/triggers/external_task.py
pankajastro: to free worker slot, no?

abhishekbhakat:
> > Using aiohttp seems unnecessary to me
> 
> I'm ok either way with or without aiohttp given that API call does not take significant time/

Aiohttp is just not working for me with Astro. Astro is force closing the session. I tried to limit the amount of concurrent calls with asyncio's [semaphores](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Semaphore), it also failed.

Creating concurrent calls seems like forcing endpoints(Astro in this case), which seems a little overkill to me.

astronomer/providers/core/triggers/external_task.py
Lee-W: Yep, but one of the points of this change is that ` A simple request can get the results. `. I thought that it indicates the load would be super light and why can't we use the HTTP provider instad

abhishekbhakat: Leaving it to worker will cause a lot of risk to reliability. Workers may get killed for multiple reasons I guess.

abhishekbhakat:
> > Any particular reason? We'll need to have a redundant code then, isn't it?
> 
> I have not tested it but I think if you install this package like `pip install 'astronomer-providers[amazon]'` then http provider might not be available but the core will be and it would cause an issue

Ouch! I'll try it out. Maybe then we'll have to define a [TCPConnector](http://aiohttp.readthedocs.io/en/stable/client_reference.html#tcpconnector) when we create an aiohttp.ClientSession if using aiohttp. [The default threshold is 100](https://docs.aiohttp.org/en/stable/http_request_lifecycle.html#:~:text=By%20default%20the,needs%20resource%20adjustment.).

abhishekbhakat:
> > > Any particular reason? We'll need to have a redundant code then, isn't it?
> > 
> > 
> > I have not tested it but I think if you install this package like `pip install 'astronomer-providers[amazon]'` then http provider might not be available but the core will be and it would cause an issue
> 
> Ouch! I'll try it out. Maybe then we'll have to define a [TCPConnector](http://aiohttp.readthedocs.io/en/stable/client_reference.html#tcpconnector) when we create an aiohttp.ClientSession if using aiohttp. [The default threshold is 100](https://docs.aiohttp.org/en/stable/http_request_lifecycle.html#:~:text=By%20default%20the,needs%20resource%20adjustment.).

Ok! I tested it with amazon-providers, seems to be working fine for me.

astronomer/providers/core/triggers/external_task.py
abhishekbhakat: @Lee-W I think we can use the Http provider with some little extra work. But the idea is to leverage the ExternalTaskSensor.

astronomer/providers/core/sensors/external_task.py
abhishekbhakat: I tried it with KubernetesExecutor too, looks working just fine to me.

Dockerfile:

```
FROM quay.io/astronomer/astro-runtime:8.8.0
USER root
RUN pip install git+https://github.com/abhishekbhakat/astronomer-providers.git@fix-externaldeploymenttasksensorasync
USER astro
```

Should I be looking for anything specific ?

pankajkoti:
I am wondering why Astro closes Aiohttp sessions. Should we also investigate on that front?

abhishekbhakat:
> I am wondering why Astro closes Aiohttp sessions. Should we also investigate on that front?

We can raise it Astro team, there could be limits. I even tried with `aiohttp.TCPConnector(limit=1)` or with `asyncio.Semaphore()` but still got connection being force closed. 

But I believe async calls are not even needed, as it's just a GET request on-repeat in certain interval and not concurrent.

astronomer/providers/core/triggers/external_task.py
pankajastro: yes, but the trigger can also get killed. and if we use sync API we might get "Triggerer's async thread was blocked for %.2f seconds" warning log in trigger

abhishekbhakat: Will that cause an issue? I think I saw that log popping up in the triggerer, even though the task was running just fine.

pankajastro: I have created an issue to investigate aiohttp https://github.com/astronomer/astronomer-providers/issues/1308

pankajastro (approved):

PR #1307, opened 2023-08-16
Remove sync wait for deletion of EKS cluster in EMR EKS DAG

pankajkoti:
It is observed that the EKS cluster DAGs sometimes hangs 
on the CloudFormation Stack to be deleted. It just prolongs 
the task from its completion. Removing the sync wait as we 
could rely on it finishing async and if it fails, the Nuke DAG 
can take care of the deletion.


pankajkoti:
cc: @vatsrahul1001

vatsrahul1001 (approved):

Issue #1312, opened 2023-08-24
Bring back hive support for Python 3.11


pankajastro:
check if we can bring back hive support for Python 3.11 once PR https://github.com/apache/airflow/pull/32607 is released

pankajkoti:
Do we need to wait for it to be released or can we pin a min version for the dependency in our list?

pankajastro:
> Do we need to wait for it to be released or can we pin a min version for the dependency in our list?

OSS PR changes
```
- sasl>=0.3.1; python_version>="3.9"
+ pyhive[hive_pure_sasl]>=0.7.0
```
I'm wondering if the installation will still try to install the airflow hive provider old one which has sasl>=0.3.1 as a dependency and it causes a problem?

pankajastro:
If I remember correctly `sasl` was causing issue and pip would still try to install it but we can try

PR #1317, opened 2023-08-29
Bring back hive provider for Python3.11

pankajastro:
closes: https://github.com/astronomer/astronomer-providers/issues/1312


pankajkoti (commented):
yes either we should bump the provider or add the needed minimum version of pyhive/sasl dependencies that supports Python3.11

pankajkoti (approved):

phanikumv (approved):

PR #1325, opened 2023-09-12
Add some logs and return after yield in deployment trigger

pankajastro:


phanikumv (commented):
can you check the mypy failures?

abhishekbhakat:
Looks good to me. The mypy error appears to be unrelated to this trigger it seems.

astronomer/providers/core/triggers/external_task.py
```diff
@@ -187,9 +187,17 @@ async def run(self) -> AsyncIterator["TriggerEvent"]:
                 resp_json = response.json()
                 if resp_json["state"] in State.finished:
                     yield TriggerEvent(resp_json)
+                    return
                 else:
+                    self.log.info(
+                        "The current status is %s. Sleeping for %s seconds",
+                        resp_json.get("state"),
+                        self.poke_interval,
+                    )
```
pankajkoti, line 196:
```suggestion
                self.log.info(
                    "The current status is %s. Sleeping for %s seconds",
                    resp_json.get("state"),
                    self.poke_interval,
                )
```
Suggestion: Since we have a return above, we can avoid this else branching. We can remove the else conditional and reduce 1 level of branching for below lines.

astronomer/providers/core/triggers/external_task.py
```diff
@@ -187,9 +187,17 @@ async def run(self) -> AsyncIterator["TriggerEvent"]:
...
                     await asyncio.sleep(self.poke_interval)
```
pankajkoti, line 197:
```suggestion
                self.log.info(
                    "The current status is %s. Sleeping for %s seconds",
                    resp_json.get("state"),
                    self.poke_interval,
                )
                await asyncio.sleep(self.poke_interval)
```
Suggestion: Since we have a return above, we can avoid this else branching. We can remove the else conditional and reduce 1 level of branching for below lines.

PR #1341, opened 2023-10-16
Fix azure provider breaking change

Lee-W:
This PR fixes the failure due to https://pypi.org/project/apache-airflow-providers-microsoft-azure/8.0.0rc1/

### DIFF
--- a/astronomer/providers/microsoft/azure/example_dags/example_adf_run_pipeline.py
+++ b/astronomer/providers/microsoft/azure/example_dags/example_adf_run_pipeline.py
@@ -7,6 +7,7 @@ from typing import cast
 from airflow import DAG
 from airflow.models.xcom_arg import XComArg
 from airflow.operators.python import PythonOperator
+
 from astronomer.providers.microsoft.azure.operators.data_factory import (
     AzureDataFactoryRunPipelineOperatorAsync,
 )
--- a/astronomer/providers/microsoft/azure/example_dags/example_adf_run_pipeline.py
+++ b/astronomer/providers/microsoft/azure/example_dags/example_adf_run_pipeline.py
@@ -40,8 +40,6 @@ df_params = {"location": LOCATION}
 default_args = {
     "execution_timeout": timedelta(hours=EXECUTION_TIMEOUT),
     "azure_data_factory_conn_id": "azure_data_factory_default",
-    "factory_name": DATAFACTORY_NAME,  # This can also be specified in the ADF connection.
-    "resource_group_name": RESOURCE_GROUP_NAME,  # This can also be specified in the ADF connection.
     "retries": int(os.getenv("DEFAULT_TASK_RETRIES", 2)),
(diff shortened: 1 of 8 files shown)


astronomer/providers/microsoft/azure/example_dags/example_adf_run_pipeline.py
```diff
@@ -97,17 +95,17 @@ def create_adf_storage_pipeline() -> None:
...
-        properties=AzureStorageLinkedService(connection_string=storage_string)
+        properties=AzureStorageLinkedService(connection_string=storage_string)  # type: ignore[arg-type]
     )
     adf_client.linked_services.create_or_update(
         RESOURCE_GROUP_NAME, DATAFACTORY_NAME, STORAGE_LINKED_SERVICE_NAME, ls_azure_storage
     )
 
     # Create an Azure blob dataset (input)
-    ds_ls = LinkedServiceReference(reference_name=STORAGE_LINKED_SERVICE_NAME)
+    ds_ls = LinkedServiceReference(type="LinkedServiceReference", reference_name=STORAGE_LINKED_SERVICE_NAME)
```
pankajkoti, line 105: Do we need to support/take care of backward compatibility for the previous versions of the provider?

astronomer/providers/microsoft/azure/example_dags/example_adf_run_pipeline.py
```diff
@@ -125,8 +123,8 @@ def create_adf_storage_pipeline() -> None:
     # Create a copy activity
     blob_source = BlobSource()
     blob_sink = BlobSink()
-    ds_in_ref = DatasetReference(reference_name=DATASET_INPUT_NAME)
-    ds_out_ref = DatasetReference(reference_name=DATASET_OUTPUT_NAME)
+    ds_in_ref = DatasetReference(type="DatasetReference", reference_name=DATASET_INPUT_NAME)
+    ds_out_ref = DatasetReference(type="DatasetReference", reference_name=DATASET_OUTPUT_NAME)
```
pankajkoti, line 127: same question for backward compatibility.

pankajkoti:
Looks like there is a mypy failure, also could we add a screenshot of a successful DAG run (either local or on Astro Cloud)?

Lee-W:
> Looks like there is a mypy failure, also could we add a screenshot of a successful DAG run (either local or on Astro Cloud)?

I think it's because it was not released by then. Yep, let me deploy one more time. The DAG disappeared (as expected) . cannot screenshot the last successful DAG run

astronomer/providers/microsoft/azure/example_dags/example_adf_run_pipeline.py
Lee-W: And we used to install azure-mgmt-datafactory >= 1.0.0 here.

Lee-W: Also as an example DAG, I think we should use the newer version. WDYT?

pankajastro (approved):
LGTM

PR #1367, opened 2023-11-23
Fixes to move deployment to stage env

vatsrahul1001:
As part of moving the deployments to the stage, we had to make some fixes in the script. sh and GitHub Actions files.


vatsrahul1001:
Tested my Changes working fine - https://github.com/astronomer/astronomer-providers/actions/runs/6970676392/job/18969102769

pankajkoti (approved):

PR #1376, opened 2023-12-05
feat(databricks): include 403 as retryable error

Lee-W:
according to user feedbacks, 403 sometimes works after retry


pankajkoti (approved):
Is there an issue or a Zendesk ticket for this? Would be nice to link it somewhere.

Lee-W:
> Is there an issue or a Zendesk ticket for this? Would be nice to link it somewhere.

Sure!
https://astronomer.zendesk.com/agent/tickets/32794

PR #1390, opened 2023-12-20
Depreacate EmrContainerSensorAsync and EmrStepSensorAsync

Lee-W:
# What's changed
Deprecate `EmrContainerSensorAsync` and `EmrStepSensorAsync` and fallback to `EmrContainerSensor` and `EmrStepSensor` with `deferrable=True`

# Why this change
The logic of `EmrContainerSensorAsync` and `EmrStepSensorAsync` has been contributed back to OSS airflow as part of `EmrContainerSensor` and `EmrStepSensor`

### EmrContainerSensorAsync
https://github.com/astronomer/astronomer-providers/blob/b5a9417919d8a13b8e519d2582ce705c4689b7c6/astronomer/providers/amazon/aws/sensors/emr.py#L38-L65

### EmrContainerSensor
https://github.com/apache/airflow/blob/999b70178a1f5d891fd2c88af4831a4ba4c2cbc9/airflow/providers/amazon/aws/sensors/emr.py#L318-L346

### EmrStepSensorAsync
https://github.com/astronomer/astronomer-providers/blob/b5a9417919d8a13b8e519d2582ce705c4689b7c6/astronomer/providers/amazon/aws/sensors/emr.py#L87-L113

### EmrStepSensor
https://github.com/apache/airflow/blob/999b70178a1f5d891fd2c88af4831a4ba4c2cbc9/airflow/providers/amazon/aws/sensors/emr.py#L644-L668


phanikumv:
Please add description about why you are doing this change

pankajkoti (commented):
We will also need to remove corresponding trigger modules, no?

Lee-W:
> what about the integration DAGs that might be using this class, are we planning to remove them at a later point of time?

Yes, I think we should keep it for a while and remove it after we deprecate the whole repo.

> Please add description about why you are doing this change

Sure. let me reformat and reword it to make it easier to read.

Lee-W:
> We will also need to remove corresponding trigger modules, no?

Yep, we should remove it as well. Thanks for reminding me!

pankajkoti (approved):

PR #1391, opened 2023-12-20
Deprecate BatchSensorAsync

Lee-W:
# What's changed
Deprecate `BatchSensorAsync` and fallback to `BatchSensor` with `deferrable=True`

# Why this change
The logic of `BatchSensorAsync` bas been contributed back to OSS airflow as part of `BatchSensor`

### BatchSensorAsync
https://github.com/astronomer/astronomer-providers/blob/b5a9417919d8a13b8e519d2582ce705c4689b7c6/astronomer/providers/amazon/aws/sensors/batch.py#L46-L68

### BatchSensor
https://github.com/apache/airflow/blob/999b70178a1f5d891fd2c88af4831a4ba4c2cbc9/airflow/providers/amazon/aws/sensors/batch.py#L98-L133

### DIFF
--- a/astronomer/providers/amazon/aws/sensors/batch.py
+++ b/astronomer/providers/amazon/aws/sensors/batch.py
@@ -1,68 +1,22 @@
 import warnings
-from datetime import timedelta
-from typing import Any, Dict
 
 from airflow.providers.amazon.aws.sensors.batch import BatchSensor
-
-from astronomer.providers.amazon.aws.triggers.batch import BatchSensorTrigger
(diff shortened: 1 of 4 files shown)


phanikumv (commented):
what about the integration DAGs that might be using this class, are we planning to remove them at a later point of time?

Lee-W:
> what about the integration DAGs that might be using this class, are we planning to remove them at a later point of time?

Yes, I think we should keep it for a while and remove it after we deprecate the whole repo.

> Please add description about why you are doing this change

Sure. let me reformat and reword it to make it easier to read.

pankajkoti (approved):

PR #1392, opened 2023-12-21
Deprecate S3KeysUnchangedSensorAsync

Lee-W:
# What's changed
Deprecate `S3KeysUnchangedSensorAsync` and fallback to `S3KeysUnchangedSensor` with `deferrable=True`

# Why this change
The logic of  `S3KeysUnchangedSensorAsync` has been contributed back to OSS airflow as part of `S3KeysUnchangedSensor`

### S3KeysUnchangedSensorAsync
https://github.com/astronomer/astronomer-providers/blob/b5a9417919d8a13b8e519d2582ce705c4689b7c6/astronomer/providers/amazon/aws/sensors/s3.py#L195-L223

### S3KeysUnchangedSensor
https://github.com/apache/airflow/blob/999b70178a1f5d891fd2c88af4831a4ba4c2cbc9/airflow/providers/amazon/aws/sensors/s3.py#L340-L374


tests/amazon/aws/sensors/test_s3_sensors.py
```diff
@@ -293,80 +293,12 @@ def test_soft_fail_enable(self, context):
...
-    def test_s3_keys_unchanged_sensor_check_trigger_instance(self, mock_hook, mock_poke, context):
-        """
-        Asserts that a task is deferred and an S3KeysUnchangedTrigger will be fired
-        when the S3KeysUnchangedSensorAsync is executed.
-        """
-        mock_hook.check_for_key.return_value = False
-
-        sensor = S3KeysUnchangedSensorAsync(
+    def test_init(self):
+        task = S3KeysUnchangedSensorAsync(
```
pankajkoti, line 297: Should we import here the OSS sensor and pass deferrable=True instead?

pankajkoti (approved):

phanikumv:
This PR is breaking our integration DAGs , @Lee-W to take a look

pankajkoti:
It maybe an issue with OSS sensor, will leave it to Wei to check once. If it’s an upstream issue, I would suggest reverting this PR for the time being until a new OSS amazon provider including the fix is released.

PR #1398, opened 2023-12-25
Deprecate s3 keys unchanged sensor async

Lee-W:
This PR is blocked by the release of ~~https://github.com/apache/airflow/pull/36363~~ https://github.com/apache/airflow/issues/36467

upgrade minimum apache-airflow-providers-amazon to 8.15.0 and deprecate S3KeysUnchangedSensorAsync

PR #640, opened 2022-09-15
Use `poke_interval` instead of `poll_interval` for sensor

bharanidharan14:
Added deprecated warning and used `poke_interval` where ever  `poll_interval` is used in the sensor.

closes: #613

PR #1406, opened 2023-12-29
Deprecate EmrJobFlowSensorAsync

Lee-W:
# What's changed
Deprecate EmrJobFlowSensorAsync and fallback to their OSS counterpart `EmrJobFlowSensor` with `deferrable=True`
 
# Why this change
Most of the logic of these operators has been contributed back to OSS airflow

### DIFF
--- a/astronomer/providers/amazon/aws/hooks/emr.py
+++ b/astronomer/providers/amazon/aws/hooks/emr.py
@@ -1,3 +1,4 @@
+import warnings
 from typing import Any, Dict, Optional
 
 from botocore.exceptions import ClientError
@@ -146,6 +147,14 @@ class EmrJobFlowHookAsync(AwsBaseHookAsync):
     """
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
+        warnings.warn(
+            (
+                "This module is deprecated and will be removed in 2.0.0."
+                "Please use :class: `~airflow.providers.amazon.aws.hooks.emr.EmrHook`."
+            ),
+            DeprecationWarning,
+            stacklevel=2,
+        )
         kwargs["client_type"] = "emr"
         super().__init__(*args, **kwargs)
 
--- a/astronomer/providers/amazon/aws/sensors/emr.py
+++ b/astronomer/providers/amazon/aws/sensors/emr.py
@@ -16,6 +16,16 @@ class EmrContainerSensorAsync(EmrContainerSensor):
     """
 
     def __init__(self, *args, **kwargs) -> None:  # type: ignore[no-untyped-def]
+        poll_interval = kwargs.pop("poll_interval")
+        if poll_interval:
+            self.poke_interval = poll_interval
+            warnings.warn(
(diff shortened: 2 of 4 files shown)


Lee-W:
Just triggered the testing workflow
圖片

Lee-W:
圖片

astronomer/providers/amazon/aws/sensors/emr.py
```diff
@@ -59,65 +49,18 @@ def __init__(self, *args, **kwargs) -> None:  # type: ignore[no-untyped-def]
...
-    :param failed_states: the failure states, sensor fails when
-        job flow reaches any of these states
+    This class is deprecated.
+    Please use :class: `~airflow.providers.amazon.aws.sensors.emr.EmrJobFlowSensor`.
     """
 
-    def __init__(
-        self,
-        *,
-        poll_interval: float = 5,
```
pankajastro, line 79: EmrJobFlowSensor uses poke_interval not poll_interval so need to that case too I feel https://github.com/apache/airflow/blob/b260367208f9c3c09bc1da2a32abf59867ddd789/airflow/providers/amazon/aws/sensors/emr.py#L524C46-L524C46

pankajkoti: +1

pankajkoti (commented):
same comment https://github.com/astronomer/astronomer-providers/pull/1398#pullrequestreview-1814774953 regarding removal of hook specific code?

astronomer/providers/amazon/aws/sensors/emr.py
Lee-W: indeed, it might be missed in https://github.com/astronomer/astronomer-providers/pull/640. let me add a similar deprecation here.

Lee-W:
> same comment [#1398 (review)](https://github.com/astronomer/astronomer-providers/pull/1398#pullrequestreview-1814774953) regarding removal of hook specific code?

Just removed. Thanks!

astronomer/providers/amazon/aws/hooks/emr.py
```diff
@@ -146,6 +147,11 @@ class EmrJobFlowHookAsync(AwsBaseHookAsync):
     """
 
     def __init__(self, *args: Any, **kwargs: Any) -> None:
+        warnings.warn(
+            "This module is deprecated and will be removed in 2.0.0.",
```
pankajkoti, line 151: Should we also suggest them to use the OSS hook instead?

astronomer/providers/amazon/aws/triggers/emr.py
```diff
@@ -57,6 +58,11 @@ def __init__(
         target_states: Optional[Iterable[str]] = None,
         failed_states: Optional[Iterable[str]] = None,
     ):
+        warnings.warn(
+            "This module is deprecated and will be removed in 2.0.0.",
```
pankajkoti, line 62: same question if we should suggest using them the OSS trigger

pankajastro (approved):
LGTM

PR #1407, opened 2023-12-29
Nuke EKS clusters 

vatsrahul1001:
This PR addresses the deletion of an EKS cluster from the nuke DAG.

### DIFF
--- a/astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
+++ b/astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
@@ -22,6 +22,12 @@ SLACK_CHANNEL = os.getenv("SLACK_CHANNEL", "#provider-alert")
 SLACK_USERNAME = os.getenv("SLACK_USERNAME", "airflow_app")
 SLACK_WEBHOOK_CONN = os.getenv("SLACK_WEBHOOK_CONN", "http_slack")
 
+REGRESSION_CLUSTER_AWS_ACCESS_KEY = os.getenv("REGRESSION_CLUSTER_AWS_ACCESS_KEY", "**********")
+REGRESSION_CLUSTER_AWS_SECRET_ACCESS_KEY = os.getenv("REGRESSION_CLUSTER_AWS_SECRET_ACCESS_KEY", "***********")
+REGRESSION_CLUSTER_AWS_DEFAULT_REGION = os.getenv("REGRESSION_CLUSTER_AWS_DEFAULT_REGION", "us-east-1")
+
+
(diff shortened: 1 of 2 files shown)


phanikumv (approved):

astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
```diff
@@ -125,6 +131,15 @@ def check_dag_status(**kwargs: Any) -> None:
         f"aws emr-containers list-virtual-clusters --state RUNNING --region {AWS_DEFAULT_REGION} | jq -r '.virtualClusters[].id' | xargs -I % aws emr-containers delete-virtual-cluster --id % --region {AWS_DEFAULT_REGION}; ",
     )
 
+    terminate_regression_clusters = BashOperator(
+        task_id="terminate_regression_clusters",
```
pankajkoti, line 135: let's mention here that these are from Astro SDK
e.g. terminate_astro_sdk_regression_clusters or terminate_dag_authoring_regression_clusters

astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
```diff
@@ -125,6 +131,15 @@ def check_dag_status(**kwargs: Any) -> None:
...
+        bash_command=f"set -e; "
+        f"aws configure set aws_access_key_id {REGRESSION_CLUSTER_AWS_ACCESS_KEY}; "
+        f"aws configure set aws_secret_access_key {REGRESSION_CLUSTER_AWS_SECRET_ACCESS_KEY}; "
+        f"aws configure set default.region {REGRESSION_CLUSTER_AWS_DEFAULT_REGION}; "
+        f"sh $AIRFLOW_HOME/dags/delete_eks_cluster_and_nodes.sh {REGRESSION_CLUSTER_AWS_DEFAULT_REGION}",
```
pankajastro, line 140: do we need to pass `REGRESSION_CLUSTER_AWS_DEFAULT_REGION ` as a parameter can't we access it in the bash file directly with something like `${REGRESSION_CLUSTER_AWS_DEFAULT_REGION}`?

astronomer/providers/amazon/aws/example_dags/example_aws_nuke.py
vatsrahul1001: terminate_dag_authoring_regression_clusters would be better

pankajastro (approved):
LGTM, just a small comment but that should not be a blocker

PR #1409, opened 2023-12-30
Adding packages libxml2,libxslt, zlib1g-dev 

vatsrahul1001:
It seems that there is a breaking change in the latest version of lxml (5.0.0), which now requires libxml2 and libxslt. Unfortunately, these packages are not available in Python 3.8 and Python 3.9, causing a failure. I am adding these libxml2,libxslt, zlib1g-dev as temp fix, any user with python3.8, 3.9  tries to install amazon provider they will encounter this issue.

For permanent solutions:-
1. we can add these packages in airflow repo
2. Pin lxml version==4.9.4 in redshift_connector till it get resolved in lxml

### DIFF
--- a/.circleci/integration-tests/Dockerfile.astro_cloud
+++ b/.circleci/integration-tests/Dockerfile.astro_cloud
@@ -8,7 +8,7 @@ USER root
 
 RUN apt-get update -y \
     && apt-get install -y software-properties-common \
-    && apt-get install -y wget procps gnupg2
+    && apt-get install -y wget procps gnupg2 \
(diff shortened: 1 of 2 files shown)


pankajkoti (approved):

vatsrahul1001:
It looks like there was an issue with the lxml wheel, as mentioned [here](https://github.com/lxml/lxml/pull/388#issuecomment-1872472747). Initially, our job attempted to build from source because the wheel was not available, resulting in the problem. However, the wheel issue has now been fixed, and it started working fine again without any further intervention. I will close the PR as its not needed now

PR #1427, opened 2024-01-12
Add non-ASCII logs check task in master dag

vatsrahul1001:
This PR is to test the astronomer logging provider for non-ASCII character, as we have providers tests running on all cloud providers daily, This will help us test daily without creating additional deployments.

### DIFF
--- a/.circleci/integration-tests/master_dag.py
+++ b/.circleci/integration-tests/master_dag.py
@@ -199,13 +199,9 @@ with DAG(
         ],
     )
 
-    """
-    This task is specifically added to test the astronomer logging
-    provider for non-ASCII character testing, as we have providers'
-    tests running on all cloud providers, and we do not want
-    multiple deployments for this.
-    """
-
+    # The below task `check_non_ascii_data` is specifically added to test the astronomer logging
+    # providers for non-ASCII characters, as we have providers' tests running on all
+    # cloud providers, and we do not want multiple deployments for this.
     check_non_ascii_data = PythonOperator(
         task_id="check_non_ascii_data",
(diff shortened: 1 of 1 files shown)


.circleci/integration-tests/master_dag.py
```diff
@@ -199,7 +199,20 @@ def prepare_dag_dependency(task_info, execution_time):
         ],
     )
 
-    airflow_version_check = (get_airflow_version, check_logs_data)
+    # The below task `check_non_ascii_data` is specifically added to test the astronomer logging
+    # providers for non-ASCII characters, as we have providers' tests running on all
+    # cloud providers, and we do not want multiple deployments for this.
+    check_non_ascii_data = PythonOperator(
+        task_id="check_non_ascii_data",
+        python_callable=check_log,
```
pankajkoti, line 207: Where is this callable check_log defined?

pankajkoti (commented):
I pushed a commit to transform the comment from docstring to comment.

.circleci/integration-tests/master_dag.py
vatsrahul1001: we were already using this callable from [here](https://github.com/astronomer/astronomer-providers/blob/a49fe81d86ae49e4ddba7ae988914d703e79c05f/.circleci/integration-tests/airflow_dag_introspection.py#L36)

pankajkoti (approved):

.circleci/integration-tests/master_dag.py
pankajkoti: sorry forgot to check the imports.

PR #1402, opened 2023-12-26
Deprecate EmrContainerOperatorAsync

Lee-W:
# What's changed
Deprecate `EmrContainerOperatorAsync` and fallback to `EmrContainerOperator` with `deferrable=True`
This is blocked by the release of ~~https://github.com/apache/airflow/pull/36417~~ https://github.com/apache/airflow/issues/36467

upgrade minimum apache-airflow-providers-amazon to 8.15.0

# Why this change
The logic of  `EmrContainerOperatorAsync` has been contributed back to OSS airflow as part of `EmrContainerOperator`

### EmrContainerOperatorAsync
https://github.com/astronomer/astronomer-providers/blob/b5a9417919d8a13b8e519d2582ce705c4689b7c6/astronomer/providers/amazon/aws/sensors/emr.py#L38-L65

### EmrContainerOperator
https://github.com/apache/airflow/blob/999b70178a1f5d891fd2c88af4831a4ba4c2cbc9/airflow/providers/amazon/aws/sensors/emr.py#L318-L346

PR #1430, opened 2024-01-18
refactor(amazon/emr): remove unnecessary return in EmrContainerOperatorAsync and Redshift Cluster operators

Lee-W:
address https://github.com/astronomer/astronomer-providers/pull/1402/files/d6f074c32db15394696360919f47216ba026b392#r1457142029


pankajkoti (approved):

pankajkoti (approved):
Let's also update the PR title to make it generic as we have changes in multiple operators now :)

Issue #1432, opened 2024-01-19
task "resume_redshift_cluster" in DAG "example_async_redshift_cluster_management" failed


Lee-W:
**Describe the bug**

**To Reproduce**

**Expected behavior**

Run without failure

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

Lee-W:
I've deployed and tested it on Astro Cloud, which looks fine. cc @vatsrahul1001

PR #1433, opened 2024-01-19
ci(providers/amazon): extend max_attempts for resume_redshift_cluster to 20

Lee-W:
closes: #1432

The default max_attempts is 10. If the status is still in the state of "resuming," it raises a failure. Extend this max_attempts for a longer waiting time and fix the master_dag failure.


pankajkoti (approved):
Can we add some description to the PR. What's the default value and how we are anticipating that this would resolve the issue.

Lee-W:
> Can we add some description to the PR. What's the default value and how we are anticipating that this would resolve the issue.

sure! I forgot to mark it as a draft first 🤦‍♂️ I've triggered the test on the cloud. WIll update more detail after testing

Lee-W:
image

Issue #1413, opened 2024-01-02
Deprecate core async sensors


Lee-W:
**Is your feature request related to a problem? Please describe.**

 as most of the functionality has been contributed back to OSS airflow, we're going to leverage the code there and deprecate this project

**Describe the solution you'd like**

1. Contribute the delta to the OSS provider
2. Point the astronomer provider to the OSS provider
3. Do the release and wait for a month or so before we remove the astronomer provider in a major release

**Describe alternatives you've considered**

**Additional context**

-----

## Current Progress
* Number of operator changes on OSS airflow needed: 2
* Number of operators to deprecate: 2

1. [ ] `FileSensorAsync`
    * **Note**: not yet contributed back
    * **Action**: contribute back
    * **Deprecation PR**:
    * **OSS Airflow PR**: - https://github.com/apache/airflow/pull/36840
    * **Minimum Provider Version**: -
1. [ ] `ExternalTaskSensorAsync`
    * **Note**: minor difference, difference in DagStateTrigger
    * **Action**: dig deeper
    * **Deprecation PR**:
    * **OSS Airflow PR**: https://github.com/apache/airflow/pull/36916
    * **Minimum Provider Version**: -
1. [x] `ExternalDeploymentTaskSensorAsync`
    * **Note**: other, astro-specific operator
    * **Action**: Do not contribute back to OSS as it is Astronomer specific. Keep it as it is.
    * **Deprecation PR**: -
    * **OSS Airflow PR**: -
    * **Minimum Provider Version**: -

pankajastro:
This will be released as core airflow release

phanikumv:
PR https://github.com/apache/airflow/pull/36840 created for `FileSensorAsync` 

`ExternalTaskSensorAsync` is WIP

phanikumv:
@pankajastro can you add the PR created for this issue in the above note

pankajastro:
PR has been merged. closing it

PR #1439, opened 2024-01-19
Add deprecation warning in core provider

pankajastro:
This PR just added a deprecation warning in the core provider to warn users about removal and encourage them to use the upcoming OSS sensor

closes: https://github.com/astronomer/astronomer-providers/issues/1413


Lee-W (approved):

Lee-W:
There are some test fails, but I guess they can be resolved by rebasing 🤔

PR #1444, opened 2024-01-22
Deprecate gcs sensors

Lee-W:


pankajkoti:
Mypy failures would be fixed by https://github.com/astronomer/astronomer-providers/pull/1443/files

astronomer/providers/google/cloud/hooks/gcs.py
```diff
@@ -14,6 +15,17 @@
 class GCSHookAsync(GoogleBaseHookAsync):
     """GCSHookAsync run on the trigger worker, inherits from GoogleBaseHookAsync"""
 
+    def __init__(self, **kwargs: Any):
+        warnings.warn(
+            (
+                "This class is deprecated and will be removed in 2.0.0."
```
pankajkoti, line 21: Should we also add this to the docstring of the class for both the hook and triggers?

astronomer/providers/google/cloud/sensors/gcs.py
```diff
@@ -12,38 +11,17 @@
...
-        account from the list granting this role to the originating account (templated).
+    This class is deprecated.
+    Please use :class: `~airflow.providers.google.cloud.sensors.gcs.GCSObjectExistenceSensor`
+    and set `deferrable` param to `True` instead.
     """
 
     def __init__(
         self,
+        *args: Any,
         polling_interval: float = 5.0,
```
pankajkoti, line 25: can we check if we need to pass this in the kwargs["poke_interval"] = polling interval instead of self.poke_interval=polling_interval?

Meaning to check if self.poke_interval will be used or if we need to set it in kwargs before calling super.init

astronomer/providers/google/cloud/triggers/gcs.py
```diff
@@ -26,16 +29,25 @@ def __init__(
         object_name: str,
         poke_interval: float,
         google_cloud_conn_id: str,
-        hook_params: Dict[str, Any],
+        hook_params: dict[str, Any],
     ):
+        warnings.warn(
+            (
+                "This class is deprecated and will be removed in 2.0.0."
```
pankajkoti, line 36: same question to check if we would need to add this to the docstring of the class too?

Lee-W: yep, sounds good! just added

astronomer/providers/google/cloud/hooks/gcs.py
Lee-W: yep, sounds good! just added

astronomer/providers/google/cloud/sensors/gcs.py
Lee-W: yes, we'll need to do that. just updated it

Lee-W:
image

pankajkoti (approved):

pankajkoti:
looks like we have some mypy failures that would need to be resolved.

Lee-W:
https://github.com/python/mypy/issues/6799#issuecomment-1882059741

Lee-W:
It seems to be an open issue of mypy. Will add type ignore to this one

Issue #1415, opened 2024-01-02
Deprecate snowflake async provider


Lee-W:
**Is your feature request related to a problem? Please describe.**

 as most of the functionality has been contributed back to OSS airflow, we're going to leverage the code there and deprecate this project

**Describe the solution you'd like**

1. Contribute the delta to the OSS provider
2. Point the astronomer provider to the OSS provider
3. Do the release and wait for a month or so before we remove the astronomer provider in a major release

**Describe alternatives you've considered**

**Additional context**

-----

## Current Progress
* Number of operator changes on OSS airflow needed: 3
* Number of operators to deprecate: 3

1. [ ] `SnowflakeSensorAsync`
    * **Note**: not yet contributed back
    * **Action**: contribute back
    * **Deprecation PR**:
    * **OSS Airflow PR**:
    * **Minimum Provider Version**:
1. [ ] `SnowflakeOperatorAsync`
    * **Note**: minor difference, check the state before deferring
    * **Action**: contribute back
    * **Deprecation PR**:
    * **OSS Airflow PR**:
    * **Minimum Provider Version**:
1. `SnowflakeSqlApiOperatorAsync`
    * **Note**: minor difference, check the state before deferring
    * **Action**: contribute back
    * **Deprecation PR**:
    * **OSS Airflow PR**:
    * **Minimum Provider Version**:

jschenz-chwy:
@Lee-W I was unaware of this change.  Is this functionality moving to common SQL?  I do not see support for deferrable SQL operators there yet.

Lee-W:
@jschenz-chwy Yes, it's not yet there. We're planning to contribute to the functionality and make the operator in this repo a proxy to that.

vatsrahul1001:
Created OSS [PR](https://github.com/apache/airflow/pull/36850) for SnowflakeSqlApiOperatorAsync

vatsrahul1001:
Raised deprecation [PR](https://github.com/astronomer/astronomer-providers/pull/1447)

PR #1447, opened 2024-01-22
Deprecate SnowflakeSqlApiOperatorAsync

vatsrahul1001:
Deprecate `SnowflakeSqlApiOperatorAsync` and fallback to  its Airflow OSS provider's counterpart `SnowflakeSqlApiOperator` with `deferrable=True`

closes: #1415


astronomer/providers/snowflake/triggers/snowflake_trigger.py
```diff
@@ -99,6 +94,14 @@ def __init__(
         token_life_time: timedelta,
         token_renewal_delta: timedelta,
     ):
+        warnings.warn(
+            (
+                "This class is deprecated and will be removed in 2.0.0."
+                "Use :class: `~airflow.providers.snowflake.triggers.snowflake_trigger.SnowflakeSqlApiTrigger` instead"
```
Lee-W, line 100:
```suggestion
                "Use `airflow.providers.snowflake.triggers.snowflake_trigger.SnowflakeSqlApiTrigger` instead"
```

Lee-W (commented):
Should we add deprecation warning to `SnowflakeSqlApiHookAsync` as well?

astronomer/providers/snowflake/operators/snowflake.py
```diff
@@ -224,183 +219,20 @@ def execute_complete(self, context: Context, event: dict[str, str | list[str]] |
             raise AirflowException("Did not receive valid event from the trigerrer")
 
 
-class SnowflakeSqlApiOperatorAsync(SnowflakeOperator):
+class SnowflakeSqlApiOperatorAsync(SnowflakeSqlApiOperator):
+    """
+    This class is deprecated and will be removed in 2.0.0.
```
pankajkoti, line 224: Why will we remove this in 2.0.0? I think the operators and sensors will keep proxying it right? Only the hooks & triggers would be removed, right?

astronomer/providers/snowflake/operators/snowflake.py
```diff
@@ -224,183 +219,20 @@ def execute_complete(self, context: Context, event: dict[str, str | list[str]] |
...
+    Use :class: `~airflow.providers.snowflake.operators.snowflake.SnowflakeSqlApiOperator` instead.
```
pankajkoti, line 225: Let's also mention here that they need to pass deferrable param as True like mentioned below.
This will come in the docs and would be nice to tell them what is the deferrable alternative for this deprecated operator.

vatsrahul1001:
> Should we add deprecation warning to `SnowflakeSqlApiHookAsync` as well?

adding warning

astronomer/providers/snowflake/operators/snowflake.py
vatsrahul1001: done

astronomer/providers/snowflake/operators/snowflake.py
vatsrahul1001: yes you are correct, thanks, fixed it

pankajkoti:
Verified running a DAG locally with these changes that it enters deferred state and then succeeds

Screenshot 2024-01-29 at 3 01 06 PM

Screenshot 2024-01-29 at 3 04 09 PM

astronomer/providers/snowflake/operators/snowflake.py
```diff
@@ -2,18 +2,17 @@
...
 
-import requests
 from airflow.exceptions import AirflowException
 
 from snowflake.connector import SnowflakeConnection
 from snowflake.connector.constants import QueryStatus
 
 try:
-    from airflow.providers.snowflake.operators.snowflake import SnowflakeOperator
+    from airflow.providers.snowflake.operators.snowflake import SnowflakeOperator, SnowflakeSqlApiOperator
```
Lee-W, line 15: I think we'll need to move this SnowflakeSqlApiOperator import outside of this block

Lee-W: I think we can remove SnowflakeOperator here. We've update our snowflake provider version

Lee-W (approved):

Issue #1420, opened 2024-01-02
Deprecate http async provider


Lee-W:
**Is your feature request related to a problem? Please describe.**

 as most of the functionality has been contributed back to OSS airflow, we're going to leverage the code there and deprecate this project

**Describe the solution you'd like**

1. Contribute the delta to the OSS provider
2. Point the astronomer provider to the OSS provider
3. Do the release and wait for a month or so before we remove the astronomer provider in a major release

**Describe alternatives you've considered**

**Additional context**

-----

## Current Progress
* Number of operator changes on OSS airflow needed: 1
* Number of operators to deprecate: 1

1. [x] `HttpSensorAsync`
    * **Note**: not yet contributed back
    * **Action**: contribute back
    * **Deprecation PR**: https://github.com/astronomer/astronomer-providers/pull/1459
    * **OSS Airflow PR**: https://github.com/apache/airflow/pull/36904
    * **Minimum Provider Version**: 4.9.0

vatsrahul1001:
Raised [PR](https://github.com/apache/airflow/pull/36904) in OSS

phanikumv:
@vatsrahul1001 to link the relevant PR for this

pankajkoti:
PRs linked and merged. Closing the issue.

PR #1459, opened 2024-01-25
Deprecate HttpSensorAsync

vatsrahul1001:
This PR deprecates HttpSensorAsync by proxying them to their Airflow OSS provider's.

closes:https://github.com/astronomer/astronomer-providers/issues/1420


astronomer/providers/http/hooks/http.py
```diff
@@ -43,6 +35,14 @@ def __init__(
         *,
         keep_response: bool = False,
     ) -> None:
+        warnings.warn(
+            (
+                "This class is deprecated. "
+                "Use `airflow.providers.dbt.cloud.hooks.dbt.DbtCloudHook` instead."
```
pankajkoti, line 41: We need to correct it to Http 🙂

pankajastro:
Note: bump http provider too https://github.com/astronomer/astronomer-providers/blob/main/setup.cfg

astronomer/providers/sftp/triggers/sftp.py
pankajastro, line 1: Why we are deleting this file?

vatsrahul1001: It should not, I will fix this

vatsrahul1001:
Ran example DAG to verify DAG is doing in deferred state

image

pankajkoti (approved):

Issue #1418, opened 2024-01-02
Deprecate kubernetes async provider


Lee-W:
**Is your feature request related to a problem? Please describe.**

 as most of the functionality has been contributed back to OSS airflow, we're going to leverage the code there and deprecate this project

**Describe the solution you'd like**

1. Contribute the delta to the OSS provider
2. Point the astronomer provider to the OSS provider
3. Do the release and wait for a month or so before we remove the astronomer provider in a major release

**Describe alternatives you've considered**

**Additional context**

-----

## Current Progress
* Number of operator changes on OSS airflow needed: 1
* Number of operators to deprecate: 1

1. [ ] `KubernetesPodOperatorAsync`
    * **Note**: different, using different triggerer
    * **Action**: dig deeper
    * **Deprecation PR**: https://github.com/astronomer/astronomer-providers/pull/1465
    * **OSS Airflow PR**: 
    1. https://github.com/apache/airflow/pull/37279
    2. https://github.com/apache/airflow/pull/37363
    3. https://github.com/apache/airflow/pull/37514
    4. https://github.com/apache/airflow/pull/37454
    5. https://github.com/apache/airflow/pull/37508
    * **Minimum Provider Version**:

pankajastro:
OSS KPO prints the container log at the end of execution, specifically when the execute_complete method runs. I am considering logging the container log in the trigger, but doing so would result in the log being printed twice: once when the trigger runs and again from the execute_complete method.

I have a rough implementation in progress, which still needs testing and the addition of the parameter periodic_container_log. This parameter, when set to True, will also print the log from the trigger and keep it False by default.

https://github.com/apache/airflow/compare/main...astronomer:airflow:kpo_logging?expand=1

pankajkoti:
We verified that PR https://github.com/apache/airflow/pull/37279 is able to fetch the container logs periodically.

<img width="1728" alt="Screenshot 2024-02-12 at 12 49 29 PM" src="https://github.com/astronomer/astronomer-providers/assets/10206082/25af42f5-d869-4826-a38d-4d8c672456f1">

pankajastro:
OSS PR Fix: 
- https://github.com/apache/airflow/pull/37279
- https://github.com/apache/airflow/pull/37454
- https://github.com/apache/airflow/pull/37363
- https://github.com/apache/airflow/pull/37508
- https://github.com/apache/airflow/pull/37514

pankajkoti:
I have linked the above PRs in the description above.

PR #1465, opened 2024-01-30
deprecate KubernetesPodOperatorAsync

Lee-W:
Deperecate KubernetesPodOperatorAsync and proxy it to its OSS counterpart.

related: #1418


astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py
```diff
@@ -21,129 +15,20 @@ class PodNotFoundException(AirflowException):
...
-            pod=self.pod or self.pod_request_obj,
-            remote_pod=remote_pod,
-        )
-        ti = context["ti"]
-        ti.xcom_push(key="pod_name", value=self.pod.metadata.name)
-        ti.xcom_push(key="pod_namespace", value=self.pod.metadata.namespace)
-        if self.do_xcom_push:
-            return result
+        self.logging_interval = logging_interval
+        super().__init__(deferrable=True, **kwargs)
```
pankajastro, line 34: we also need to pass logging_interval

pankajkoti: If it's present in the kwargs, it will be passed is. The default previous value here for logging_interval was None, and the upstream has default None too, so no additional care needs to be taken.

astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py
```diff
@@ -21,129 +15,20 @@ class PodNotFoundException(AirflowException):
...
-    .. warning::
-        By default, logs will not be available in the Airflow Webserver until the task completes. However,
-        you can configure ``KubernetesPodOperatorAsync`` to periodically resume and fetch logs.  This behavior
-        is controlled by param ``logging_interval``.
-
-    :param poll_interval: interval in seconds to sleep between checking pod status
-    :param logging_interval: max time in seconds that task should be in deferred state before
-        resuming to fetch latest logs. If ``None``, then the task will remain in deferred state until pod
-        is done, and no logs will be visible until that time.
+    Please use :class: `~airflow.providers.cncf.kubernetes.operators.kubernetes_pod.KubernetesPodOperator`
```
pankajkoti, line 20:
```suggestion
    Please use :class: `~airflow.providers.cncf.kubernetes.operators.pod.KubernetesPodOperator`
```

astronomer/providers/cncf/kubernetes/operators/kubernetes_pod.py
```diff
@@ -21,129 +15,20 @@ class PodNotFoundException(AirflowException):
...
-        self.defer()
-
-    def execute_complete(self, context: Context, event: Dict[str, Any]) -> Any:  # type: ignore[override]
-        """Deprecated; replaced by trigger_reentry."""
+    def __init__(self, **kwargs: Any):
         warnings.warn(
-            "Method `execute_complete` is deprecated and replaced with method `trigger_reentry`.",
+            (
+                "This module is deprecated."
+                "Please use `airflow.providers.cncf.kubernetes.operators.kubernetes_pod.KubernetesPodOperator`"
```
pankajkoti, line 28:
```suggestion
                "Please use `airflow.providers.cncf.kubernetes.operators.pod.KubernetesPodOperator`"
```

pankajkoti (approved):
I will merge this so that we can test our regressions DAGs together on all RCs and create a separate PR later to update the versions in setup.cfg to released versions from the current RCs

PR #1505, opened 2024-04-10
Update test-rc job schedule to run only once at midnight

pankajkoti:
This is with an intent to save some costs and running once at midnight should be sufficient.


Lee-W (approved):

pankajastro:
I think earlier it was once in 24hr but we changed since for rc2 etc we get 24hr

pankajastro (approved):

vatsrahul1001 (approved):

PR #1507, opened 2024-04-12
Fix task name to get cluster IP using xcom

vatsrahul1001:
We were not using correct task for cluster IP retrieval. Updated task name correctly.

### DIFF
--- a/astronomer/providers/sftp/example_dags/example_sftp.py
+++ b/astronomer/providers/sftp/example_dags/example_sftp.py
@@ -167,7 +167,7 @@ def revoke_inbound_rules(task_instance: TaskInstance) -> None:
         FromPort=22,
         ToPort=22,
         GroupId=task_instance.xcom_pull(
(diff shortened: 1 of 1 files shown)


phanikumv (approved):

astronomer/providers/sftp/example_dags/example_sftp.py
```diff
@@ -167,7 +167,7 @@ def revoke_inbound_rules(task_instance: TaskInstance) -> None:
         FromPort=22,
         ToPort=22,
         GroupId=task_instance.xcom_pull(
-            key="cluster_response_master_security_group", task_ids=["describe_created_cluster"]
+            key="instance_response_master_security_group", task_ids=["create_ec2_instance"]
```
pankajastro, line 170: maybe we should use env instead of hardcoded value https://github.com/astronomer/astronomer-providers/pull/1509

PR #1514, opened 2024-04-19
Fix issue with API response not containing all tags for runtime version update

vatsrahul1001:
We noticed that the [action](https://github.com/astronomer/astronomer-providers/actions/runs/8746284419) job to update the runtime version in the integration test deployment created a [PR](https://github.com/astronomer/astronomer-providers/pull/1513) in which the version was not the latest. After debugging, we found out that the response from the API [call](https://github.com/astronomer/astronomer-providers/blob/afdbace9f41dab4fe7159926505604a0e18735b4/.github/scripts/get_latest_runtime_image_tag.py#L14) does not contain all the tags. This PR addresses the issue after adding a limit query parameter


Lee-W (approved):

.github/scripts/get_latest_runtime_image_tag.py
```diff
@@ -25,7 +26,10 @@ def get_latest_tag(repository: str) -> str:
         except ValueError:
             continue
     if valid_tags:
+        print(valid_tags)
```
pankajkoti, line 29: Can we remove this? 
```suggestion
```

.github/scripts/get_latest_runtime_image_tag.py
```diff
@@ -25,7 +26,10 @@ def get_latest_tag(repository: str) -> str:
...
         latest_tag = max(valid_tags)
+
+        # print(latest_tag)
```
pankajkoti, line 32:
```suggestion
```

.github/scripts/get_latest_runtime_image_tag.py
```diff
@@ -10,10 +10,11 @@
 
 def get_latest_tag(repository: str) -> str:
     """Get the latest semantic version tag from a Quay.io repository."""
-    url = f"https://quay.io/api/v1/repository/{repository}/tag/"
+    url = f"https://quay.io/api/v1/repository/{repository}/tag/?limit=1000"
     response = requests.get(url)
     response.raise_for_status()
     data = response.json()
+    print("Response data:", data)
```
pankajkoti, line 17:
```suggestion
```

.github/scripts/get_latest_runtime_image_tag.py
Lee-W: I thought this line was for easy debugging?

PR #1519, opened 2024-04-30
Pass `parameters` to parent class in SnowflakeSensorAsync

josh-fell:
Currently the SnowflakeSensorAsync does not propagate `parameters` to its parent class, SqlSensor. Because of this, `parameters` are not passed for execution and can cause task failures when attempting doing so.

### DIFF
--- a/astronomer/providers/snowflake/sensors/snowflake.py
+++ b/astronomer/providers/snowflake/sensors/snowflake.py
@@ -46,7 +46,7 @@ class SnowflakeSensorAsync(SqlSensor):
         *,
         snowflake_conn_id: str,
         sql: str,
-        parameters: str | None = None,
+        parameters: dict[str, Any] | None = None,
         success: str | None = None,
         failure: str | None = None,
         fail_on_empty: bool = False,
@@ -63,6 +63,7 @@ class SnowflakeSensorAsync(SqlSensor):
         super().__init__(
             conn_id=snowflake_conn_id,
             sql=sql,
+            parameters=parameters,
             success=success,
             failure=failure,
             fail_on_empty=fail_on_empty,
--- a/tests/snowflake/sensors/test_snowflake.py
+++ b/tests/snowflake/sensors/test_snowflake.py
@@ -103,3 +103,25 @@ class TestPytestSnowflakeSensorAsync:
         )
         with pytest.raises(AirflowSkipException):
             sensor.execute(context)
+
+    @mock.patch("airflow.providers.common.sql.sensors.sql.SqlSensor.__init__")
+    def test_call_with_parameters(self, mock_sql_sensor):
+        """Ensure SQL parameters are passed to the SqlSensor parent class."""
+        SnowflakeSensorAsync(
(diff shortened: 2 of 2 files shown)


Lee-W (approved):

astronomer/providers/snowflake/sensors/snowflake.py
```diff
@@ -63,6 +63,7 @@ def __init__(
         super().__init__(
             conn_id=snowflake_conn_id,
             sql=sql,
+            parameters=parameters,
```
pankajkoti, line 66: Would they have not been passed already as part of **kwargs that get passed to the parent class?

josh-fell: Unfortunately not. `parameters` are extracted as their own keyword arg (as they are in SqlSensor too) and wouldn't be part of arbitrary `**kwargs` when being passed to the parent.

Hopefully to illustrate, `ChildOne` inherits from `ParentOne` and `parameters` are not passed to `ParentOne`. Both `kwargs` and `parameters` are empty dicts.
```
>>>
>>> class ParentOne:
...     def __init__(self, parameters=None, **kwargs) -> None:
...         self.parameters = parameters or {}
...         self.kwargs = kwargs
...
>>>
>>> class ChildOne(ParentOne):
...     def __init__(self, parameters, **kwargs) -> None:
...         super().__init__(**kwargs)
...
>>>
>>> c1 = ChildOne(parameters={"param1": 123})
>>> print(c1.parameters)
{}
>>> print(c1.kwargs)
{}
```
Now a similar setup but `ChildTwo` passes `parameters` to `ParentTwo`. `ParentTwo.parameters` is what was initialized when originally calling `ChildTwo`.
```
>>>
>>>
>>> class ParentTwo:
...     def __init__(self, parameters=None, **kwargs) -> None:
...         self.parameters = parameters or {}
...         self.kwargs = kwargs
...
>>>
>>> class ChildTwo(ParentTwo):
...     def __init__(self, parameters, **kwargs) -> None:
...         super().__init__(parameters=parameters, **kwargs)
...
>>>
>>> c = ChildTwo(parameters={"param1": 123})
>>> print(c.parameters)
{'param1': 123}
>>> print(c.kwargs)
{}
>>>
```

pankajkoti: oops sorry I totally misread the the existing implementation. I thought we were not having an existing parameters param and were newly adding there. 

I thought we had a setup like this
```
class A(base):
    def __init__(self, parameters=None, **kwargs):
        self.parameters = parameters
        print("A parameters", parameters)
        super().__init__(**kwargs)

class B(A):
    def __init__(self, **kwargs):
        print("B")
        super().__init__(**kwargs)


b = B(parameters="test")
```

din't realise B already defined parameters and extracted from kwargs.

pankajkoti (approved):

pankajkoti:
@Lee-W perhaps we need to make a release for astronomer-providers after merging this?

Issue #1528, opened 2024-05-13
Unable to use `failure` callable in SnowflakeSensorAsync when deferred


josh-fell:
**Describe the bug**
Version used: `astronomer-providers==1.19.1`

When using a `failure` callable to determine whether a SnowflakeSensorAsync task should fail based on the SQL result, the task fails stating the `failure` callable is _not_ actually a callable. This failure only occurs when the SnowflakeSensorAsync task goes into a deferred state and the SQL check is handled by the Triggerer.

FWIW, the same setup with a `failure` callable works fine using the SqlSensor from the Common SQL provider though.

**To Reproduce**
1. Write a SnowflakeSensorAsync task with a configured `failure` callable and make sure it won't return after the first poke. Something like:
```py
SnowflakeSensorAsync(
    task_id=...,
    snowflake_conn_id=...
    sql=...,
    failure=lambda result: result == "failed",
)
```
2. Execute task and observe a similar exception message:
```
File "/usr/local/lib/python3.11/site-packages/astronomer/providers/utils/sensor_util.py", line 27, in raise_error_or_skip_exception
    raise AirflowException(error_message)
airflow.exceptions.AirflowException: self.failure is present, but not callable ->         lambda result: result == "failed"
```

**Expected behavior**
When configured with a `failure` callable that is a callable type, the task should not fail but execute the `failure` callable to assess whether the task should indeed fail.

**Additional context**
This is probably an issue with the `success` callable as well. My initial suspicion is that these callable parameters are getting serialized to the SnowflakeTrigger as strings and therefore not callable. Perhaps it would be better to run the `success`/`failure` callables as part of the sensor rather than the trigger? Meaning the SnowflakeTrigger pokes and awaits a result, and if there is a `success`/`failure` callable configured, that check is executed by the SnowflakeSensorAsync logic.

PR #1530, opened 2024-05-14
Fix SnowflakeSensorTrigger storing callable as string

pankajastro:
closes: https://github.com/astronomer/astronomer-providers/issues/1528

When saved in the trigger kwargs, both success and failure parameters are stored as strings rather than callable functions. This PR aims to move the validation logic from trigger to operator component  

Screenshot 2024-05-22 at 10 48 47 PM


astronomer/providers/snowflake/example_dags/example_snowflake_sensor.py
```diff
@@ -30,6 +30,7 @@
     "retry_delay": timedelta(seconds=int(os.getenv("DEFAULT_RETRY_DELAY_SECONDS", 60))),
 }
 
+
```
utkarsharma2, line 33:
```suggestion
```

astronomer/providers/snowflake/sensors/snowflake.py
```diff
@@ -74,30 +75,61 @@ def __init__(
...
-                    failure=self.failure,
-                    fail_on_empty=self.fail_on_empty,
-                    dag_id=context["dag"].dag_id,
-                    task_id=context["task"].task_id,
-                    run_id=context["dag_run"].run_id,
-                    snowflake_conn_id=self.snowflake_conn_id,
-                ),
-                method_name=self.execute_complete.__name__,
-            )
+            self._defer(context)
```
sunank200, line 78: why is private method needed here?

astronomer/providers/snowflake/sensors/snowflake.py
```diff
@@ -74,30 +75,61 @@ def __init__(
...
 
-    def execute_complete(self, context: Context, event: dict[str, str] | None = None) -> Any:
+    def _validate_result(self, result: list[tuple[Any]]) -> Any:
```
sunank200, line 80: wondering why is this private?

astronomer/providers/snowflake/sensors/snowflake.py
```diff
@@ -74,30 +75,61 @@ def __init__(
...
+            else:
+                raise AirflowException(f"self.failure is present, but not callable -> {self.failure}")
+        if self.success is not None:
+            if callable(self.success):
+                return self.success(first_cell)
+            else:
+                raise AirflowException(f"self.success is present, but not callable -> {self.success}")
+        return bool(first_cell)
+
+    def _defer(self, context: Context) -> None:
```
sunank200, line 102: why is this separate method required

astronomer/providers/snowflake/sensors/snowflake.py
```diff
@@ -74,30 +75,61 @@ def __init__(
...
+    def execute_complete(self, context: Context, event: dict[str, Any] | None = None) -> Any:
         """
         Callback for when the trigger fires - returns immediately.
         Relies on trigger to throw an exception, otherwise it assumes execution was
         successful.
         """
         if event:
+            if "status" in event and event["status"] == "validate":
+                marker = self._validate_result(event["result"])
+                if marker:
```
sunank200, line 129:
```suggestion
                if self._validate_result(event["result"]):
```

astronomer/providers/snowflake/sensors/snowflake.py
sunank200: I think it's a separate method used in execute_complete

astronomer/providers/snowflake/sensors/snowflake.py
pankajastro: I prefer private method because it easy to remove in future without deprecation

astronomer/providers/snowflake/sensors/snowflake.py
pankajastro: I prefer private method because it easy to remove in future without deprecation

astronomer/providers/snowflake/sensors/snowflake.py
pankajastro: it has been used in execute_complete and execute so to avoid duplicate code

vatsrahul1001 (commented):
should we add example intergration dag test for this?

astronomer/providers/snowflake/sensors/snowflake.py
```diff
@@ -74,30 +75,61 @@ def __init__(
...
+                    return
```
pankajkoti, line 130: Does this mean successful completion of the task. We should log such a message in that case before returning.

astronomer/providers/snowflake/sensors/snowflake.py
```diff
@@ -74,30 +75,61 @@ def __init__(
...
+                else:
+                    self._defer(context)
             if "status" in event and event["status"] == "error":
                 raise_error_or_skip_exception(self.soft_fail, event["message"])
             self.log.info(event["message"])
```
pankajkoti, line 135: Not your change, but the else block below looks wrong. When can event be not set? If not set, we should raise an exception?

pankajkoti (commented):
Were we able to to test changes with our example DAG?

astronomer/providers/snowflake/sensors/snowflake.py
pankajastro: yes, it will be considered as successful completion. I have added a log

astronomer/providers/snowflake/sensors/snowflake.py
pankajastro: I have fixed this PTAL

pankajastro:
> should we add example intergration dag test for this?

Added a task PTAL

pankajkoti (approved):
LGTM if we tested this with a DAG

PR #672, opened 2022-09-29
Test RC provider packages

pankajkoti:

PR #1531, opened 2024-05-15
Update cluster configuration to Azure

vatsrahul1001:
"In accordance with [Issue #672](https://github.com/astronomer/issues-airflow/issues/672), we need to migrate our Databricks account to Azure. Consequently, we should update this DAG with Azure cluster configuration.

### DIFF
--- a/astronomer/providers/databricks/example_dags/example_databricks.py
+++ b/astronomer/providers/databricks/example_dags/example_databricks.py
@@ -26,23 +26,22 @@ default_args = {
 
 new_cluster = {
     "num_workers": 1,
-    "spark_version": "10.4.x-scala2.12",
-    "spark_conf": {},
-    "aws_attributes": {
-        "first_on_demand": 2,
-        "availability": "SPOT_WITH_FALLBACK",
-        "spot_bid_price_percent": 100,
-        "ebs_volume_type": "GENERAL_PURPOSE_SSD",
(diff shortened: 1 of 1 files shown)


pankajastro (approved):

astronomer/providers/databricks/example_dags/example_databricks.py
```diff
@@ -13,7 +13,7 @@
 
 DATABRICKS_CONN_ID = os.getenv("ASTRO_DATABRICKS_CONN_ID", "databricks_default")
 # Notebook path as a Json object
-notebook_task = '{"notebook_path": "/Users/phani.kumar@astronomer.io/quick_start"}'
+notebook_task = '{"notebook_path": "/Users/rahul.vats@astronomer.io/quick_start"}'
```
pankajkoti, line 16: I think we should use a shared notebook here now that we're changing this.

pankajkoti (approved):
LGTM. Except that if we could a shared notebook would be nice.

vatsrahul1001:
verified [run](https://cll0nj92h00iu01j51htnafwh.astronomer-stage.run/dxpo33w7/dags/example_async_databricks/grid?search=example_async_databricks&dag_run_id=manual__2024-05-17T09%3A49%3A20.043908%2B00%3A00&task_id=submit_run) with update notebook name

Issue #1561, opened 2024-08-05
Automate databricks workflow system test run with rc testing


vatsrahul1001:
Automate databricks workflow system run with rc testing

pankajkoti:
Include the example DAG https://github.com/apache/airflow/blob/main/tests/system/providers/databricks/example_databricks_workflow.py to be run as part of our astronomer-providers deployment so that newer RCs ensure that there is no regression.

PR #1565, opened 2024-08-16
Add databricks workflow system to integration tests

vatsrahul1001:
We want to add the Databricks workflow system to our integration tests so that we can be notified in case of any failure during RC testing. This PR copies the Databricks [workflow system test](https://github.com/apache/airflow/blob/main/tests/system/providers/databricks/example_databricks_workflow.py) from the Airflow repo and integrates it with our integration tests.

closes: https://github.com/astronomer/astronomer-providers/issues/1561


pankajkoti (approved):
Thanks for the contribution and help here @vatsrahul1001 :)

PR #1578, opened 2024-08-28
add-condition-for-cloning-system-tests

vatsrahul1001:
[PR](https://github.com/astronomer/astronomer-providers/pull/1565) introduced a new step of cloning airflow and system test in a reusable wf [here](https://github.com/astronomer/astronomer-providers/pull/1565/files#diff-5681502dadc6c6fb56b8bd817f11c08de363c6ec13c3d020ee14858d2d317b0dR47). As it is a reusable workflow it's used in astronomer-providers-logging wf which started failing also we do not need that step in logging providers tests. This PR address this issue by running step only when repo is astronomer-providers

### DIFF
--- a/.github/workflows/reuse-wf-deploy-to-astro-cloud.yaml
+++ b/.github/workflows/reuse-wf-deploy-to-astro-cloud.yaml
@@ -45,6 +45,7 @@ jobs:
 
 
       - name: Clone airflow and copy system tests
+        if: ${{ github.repository == 'astronomer-providers' }}
         working-directory: .circleci/integration-tests
         run: |
             git clone https://github.com/apache/airflow.git


pankajkoti:
I am not following how the change here affects astronomer-providers-logging wf. Is there some reference we could look at?

pankajkoti (commented):
It feels a bit odd to add a condition here to check if this workflow, which is already in this repo, is run from the same repo. If the workflow were stored in a shared location and referenced by multiple repos, the condition would make sense. For now, I think it would be better to copy the workflow to the logging provider repo where it's actually being used, rather than adding this condition here. What do you think?

vatsrahul1001:
> It feels a bit odd to add a condition here to check if this workflow, which is already in this repo, is run from the same repo. If the workflow were stored in a shared location and referenced by multiple repos, the condition would make sense. For now, I think it would be better to copy the workflow to the logging provider repo where it's actually being used, rather than adding this condition here. What do you think?

I agree better would be to copy workflow to logging provider. I will close this PR and open a new one in logging provider

